PR #4, opened 2021-03-10
Extract grading view from Table component

### LINKED ISSUE #3: Extract grading view from Table component
Create a Table grading component and extract the grading html from the student html.

geoffreykwan:
Make sure Table grading and student views still work properly. Also make sure a Table with Data Explorer enabled also works properly for the student and grading.

Closes #3


hirokiterashima (approved):
Looks good.

PR #15, opened 2021-03-18
Updated webpack dependency

hirokiterashima:
Also updated path to unit tests, removed codemirror and oclazyload node dependencies, and got unit test working again (see below). The docker image for wise-client went down from 1.66GB in v2 to 1.06G in v3.

**Test**
1. Pull latest from develop branch on WISE-Docker-Dev project. The updated docker-compose file will point to the new client docker dev image (v3).
2. Run docker-compose up to start the wise dev containers.
3. Test that student, teacher, author tools work as before.

**Running unit tests**
When you have wise dev containers running using ```docker-compose up```, you can run the unit tests with these command:
- ```docker exec -it wise-client npm test``` (single run)
- ```docker exec -it wise-client npm test -- "--watch=true"``` (watches for changes)

**Note**
You can now run vi in the container. ```docker exec -it wise-client sh``` to log in, and type ```vi```


geoffreykwan (approved):
Looks good.

breity (approved):
Looks good. We tested against production.

PR #25, opened 2021-03-25
Cleaned up NodeGradingView

### LINKED ISSUE #24: Refactor NodeGradingView
### Description

- Use CompletionStatus interface to define workgroup completion status.
- Remove getLatestComponentStateByWorkgroupIdAndComponentId() function. Not being used.
- Clean up code to make it easier to read.

hirokiterashima:
I added a CompletionStatus interface to define the workgroup completion status and removed unused ```getLatestComponentStateByWorkgroupIdAndComponentId``` function and cleaned up code to make it easier to read.

Test that NodeGradingView (grade by step) works as before.

### DIFF
--- a/src/assets/wise5/classroomMonitor/classroomMonitorComponents/nodeGrading/nodeGradingView/nodeGradingView.ts
+++ b/src/assets/wise5/classroomMonitor/classroomMonitorComponents/nodeGrading/nodeGradingView/nodeGradingView.ts
@@ -39,7 +39,7 @@ export class NodeGradingViewController {
   teacherWorkgroupId: number;
   workgroupInViewById: any = {}; // whether the workgroup is in view or not
   workgroups: any;
-  workgroupsById: any = {}; // workgroup names, statuses, scores, notifications, etc.
+  workgroupsById: any = {};
   workVisibilityById: any = {}; // whether student work is visible for each workgroup
   annotationReceivedSubscription: Subscription;
   studentWorkReceivedSubscription: Subscription;
@@ -213,7 +213,7 @@ export class NodeGradingViewController {
   }
 
   getCompletionStatusByWorkgroupId(workgroupId: number): CompletionStatus {
-    const status: CompletionStatus = {
+    const completionStatus: CompletionStatus = {
       isCompleted: false,
       isVisible: false,
       latestWorkTime: null,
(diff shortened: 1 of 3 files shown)


src/assets/wise5/classroomMonitor/classroomMonitorComponents/nodeGrading/nodeGradingView/nodeGradingView.ts
```diff
@@ -35,10 +37,10 @@ export class NodeGradingViewController {
   };
   sort: any;
   teacherWorkgroupId: number;
-  workgroupInViewById: any;
+  workgroupInViewById: any = {}; // whether the workgroup is in view or not
   workgroups: any;
-  workgroupsById: any;
-  workVisibilityById: any;
+  workgroupsById: any = {}; // workgroup names, statuses, scores, notifications, etc.
```
geoffreykwan: This comment is a little confusing. The variable name is called workgroupsById so what do "workgroup names, statuses, scores, notifications, etc." have to do with it? Do you mean name, status, score, notifications are in a workgroup object?

src/assets/wise5/classroomMonitor/classroomMonitorComponents/nodeGrading/nodeGradingView/nodeGradingView.ts
```diff
@@ -196,113 +186,81 @@ export class NodeGradingViewController {
...
+    return false;
   }
 
-  getCompletionStatusByWorkgroupId(workgroupId) {
-    let isCompleted = false;
-    let isVisible = false;
-    let latestWorkTime = null;
-    let latestAnnotationTime = null;
+  getCompletionStatusByWorkgroupId(workgroupId: number): CompletionStatus {
+    const status: CompletionStatus = {
```
geoffreykwan: Should this status variable name be changed to completionStatus? In this function there is a studentStatus variable and a nodeStatus variable and if you have another variable just called status it's a little confusing.

geoffreykwan (changes requested):
Functionality works well. I only have some minor code styling comments.

src/assets/wise5/classroomMonitor/classroomMonitorComponents/nodeGrading/nodeGradingView/nodeGradingView.ts
hirokiterashima: The original comment was 
```
this.workgroupsById = {}; // object that will hold workgroup names, statuses, scores, notifications, etc.
```
Which isn't that much more useful either. Should we just remove the comment?

geoffreykwan: If the comment was something like

// workgroup objects contain name, status, score, notifications, etc.

it would be better. Alternatively we could just remove the comment or create an interface for the workgroup object although we may need to differentiate this workgroup object (that has had data like status, score, notification injected into it) compared to a plain workgroup object that only contains id, name, period.

hirokiterashima: ok, I removed the comment to avoid confusion.

src/assets/wise5/classroomMonitor/classroomMonitorComponents/nodeGrading/nodeGradingView/nodeGradingView.ts
hirokiterashima: good point, I changed it to completionStatus.

geoffreykwan (approved):
Looks good.

PR #26, opened 2021-03-29
Extract Embedded component grading view

### LINKED ISSUE #20: Extract Embedded component grading view
Create an Embedded grading component and extract the grading html from the student html. Also load the student work into the Embedded grading component. We previously have not displayed anything in the grading view for Embedded components.

geoffreykwan:
Make sure embedded model student work is shown in the Grading Tool if the component authoring for the embedded component has "Saves Student Work" checked and the model actually saves student work. The grading view for the model will show the last state the model was in.

Also make sure in the Grade By Step view for a step, that the Step Info button in the upper right of the Grading Tool properly numbers the embedded component steps if it has work. An embedded component that saves work should be displayed as "1. Embedded (Custom)". If it does not save work, there will be no number and no embedded label.

Closes #20


hirokiterashima (approved):
Looks good. Tested against production.

PR #28, opened 2021-03-30
Extract Animation component grading view

### LINKED ISSUE #27: Extract Animation component grading view
Create an Animation grading component and extract the grading html from the student html. We previously have not displayed anything in the grading view for Animation components. Perhaps we can show some basic data such as the number of times the play button was clicked.

geoffreykwan:
Make sure Animation still works properly in the VLE and Grading Tool. In the Grading Tool, make sure any new Animation student data shows the number of times the animation was played (this will not show up for old student data because we did not save that value before).

Closes #27


hirokiterashima (approved):
Looks good. Tested against production.

PR #37, opened 2021-04-02
Clean up remaining grading mode checks

### LINKED ISSUE #29: Clean up remaining grading mode checks
There are still some checks for mode == grading or gradingRevision in html and controller files that I previously missed removing. We should also be able to remove the "grading" and "gradingRevision" mode check in the componentController.ts file now that all components have a grading component now.

Also 
- remove the "component" block from workgroupComponentGrading.html and workgroupComponentRevisions.ts

geoffreykwan:
Make sure all the component types still work in the grading, student, and authoring views.

Closes #29


hirokiterashima (approved):
Looks good, tested on production.

PR #41, opened 2021-04-05
Add structure choose location now properly displays number and title

### LINKED ISSUE #40: Add structure choose location does not show step title.
### Description

How to reproduce:
1. In the authoring tool, choose to add structure. Choose any structure to add, and go to the next page to select location for the new structure. Step titles do not appear, see below

Screen Shot 2021-04-05 at 8 03 27 AM

hirokiterashima:
This was happening because don't have projectController anymore in the child pages.
Also removed step background color because it doesn't matter in group selection.

Test that you can see the step number and title when choosing a location to add the new structure.

Closes #40


geoffreykwan (approved):
Looks good.

PR #45, opened 2021-04-05
Show scrollbar in NotebookNotes if notes content overflows visible area

### LINKED ISSUE #43: Scroll bar doesn't appear in NotebookNotes
Scrollbar never appears when viewing student notes in the VLE. It should appear if there is overflow.

breity:
To test:
- Run a unit (in preview mode or as a student) that has Notebook notes enabled.
- Add enough notes so that the content starts to overflow the visible area on the screen.
- Verify that a scrollbar appears.

Resolves #43.


hirokiterashima (approved):
Looks good. tested on production.

PR #51, opened 2021-04-06
Changed add step into a wizard

### LINKED ISSUE #47: Add new step wizard sequence
## Description
Change the add new step flow from an inline view in the unit view to its own route, like the add-structure view.

## Current behavior
1. In the unit view, the author clicks on the button to add a new step to the unit.
2. In the unit view, the author is asked to type in a new step title in the unit view. 
3. After typing the new title and hitting the checkmark, they are asked to choose a location for the new step. 
4. After selecting a location, they see the unit view with the new step added in the unit.

image

## Desired behavior
1. In the unit view, the author clicks on the button to add a new step to the unit.
2. This opens a new step wizard view. The author is asked to type in a new step title.
3. After typing the new title and hitting the ok button, they are asked to choose a location for the new step.
4. After selecting a location, they see the step authoring view for the new step.

Below is what the first page of the add step wizard view will look like in the future. For this task, the user will only have the option to type in the New Step Title and Cancel/Next.
image

hirokiterashima:
Adding a step in the AT is now done with a wizard. It will prompt you for a title on the first page, and a location on the next page.

Also cleaned up code related to add step and add group toggle.

## Test
1. Test that you can add a new step to the unit as the first step in an activity, or after an existing step, and you will be taken to the new step authoring view.
2. Test that you can cancel out of the wizard and return to the unit view.
3. Test that adding a new lesson works as before also.

Closes #47


geoffreykwan (approved):
Looks good.

PR #58, opened 2021-04-07
Always show Student Work tab in milestone reports

### LINKED ISSUE #57: Show Student Work tab in milestone reports when opened from step grading view
We hide the Student Work tab in milestone reports when they are opened from the step grading view. Original rationale was that the student work is already visible in the step grading view, so would be redundant in the report.

Now that we show special initial, revised, and change in score columns for milestones that have multiple locations, we should always show the Student Work tab (since the score comparison displays don't appear in the regular step grading view).

breity:
To test:
- Load Teacher Tools for a run with a milestone report.
- Open step in Grade By Step mode.
- Open milestone report by selecting "Report" button at top of Step Grading view.
- Make sure Student Work tab is present and viewable.

Resolves #57.


hirokiterashima (approved):
Looks good. Tested on production.

PR #65, opened 2021-04-09
Fixed i18n extraction

breity:
Fixed i18n extraction and updated package.json script. Using non-Ivy extraction for now, as we're still seeing errors with Ivy message extraction. So this only extracts template messages for now, not in-code.

Also updated messages.xlf.

To Test:
- Run `docker exec -it wise-client npm run extract-i18n` while Docker container is running and verify that i18n extraction completes without errors.


hirokiterashima (approved):
Looks good.

PR #68, opened 2021-04-13
In-code/template extraction working with ivy

### LINKED ISSUE #67: Extract i18n strings in template and code using Angular only
### Description

Build-in Angular i18n extraction using ivy now supports in-code translations. Switch over to it from using our custom extraction library and script.

hirokiterashima:
## Notes
- The issue was .babelrc file hindering the extract-i18n build.
- Also removed old scripts and library to merge locale files manually.
- The pre-commit hook will also work, but only if the wise-client Docker service is running. We might be able to find a better extraction solution in a future task.

## Test
Test that you can trigger a script to extract i18n strings from both template and code by running ```docker exec -it wise-client npm run extract-i18n``` while the wise-client docker service is running. This should update src/messages.xlf if there are any changes in the template or code.

Closes #67


breity (approved):
Looks great.

PR #77, opened 2021-04-14
Make notifyConnectedParts use an observable for Multiple Choice

### LINKED ISSUE #75: Make notifyConnectedParts use an observable for Multiple Choice
Right now in nodeController.ts we use componentToScope to obtain the child scope to call scope.handleConnectedComponentStudentDataChanged() but we should use observables.

Implement notifyConnectedParts with an observable for Multiple Choice.

geoffreykwan:
Make sure Multiple Choice connected to a Multiple Choice works.

Test that the connected component works when the second Multiple Choice component
- is on the same step
- is on a different step
- is importing work
- is showing work

Closes #75

### DIFF
--- a/src/assets/wise5/components/component-student.component.ts
+++ b/src/assets/wise5/components/component-student.component.ts
@@ -87,6 +87,7 @@ export abstract class ComponentStudent {
   subscribeToSubscriptions(): void {
     this.subscribeToAnnotationSavedToServer();
     this.subscribeToNodeSubmitClicked();
+    this.subscribeToNotifyConnectedComponents();
     this.subscribeToStudentWorkSavedToServer();
     this.subscribeToRequestComponentState();
   }
@@ -113,6 +114,22 @@ export abstract class ComponentStudent {
         }
       })
     );
+  }
+
+  subscribeToNotifyConnectedComponents(): void {
+    this.subscriptions.add(
+      this.ComponentService.notifyConnectedComponentSource$.subscribe(
+        ({ nodeId, componentId, componentState }) => {
+          if (this.nodeId === nodeId && this.componentId === componentId) {
+            this.processConnectedComponentState(componentState);
+          }
+        }
+      )
+    );
+  }
+
+  processConnectedComponentState(componentState: any): void {
+    // overridden by children
   }
 
   isForThisComponent(object: any) {
--- a/src/assets/wise5/components/component-student.component.ts
+++ b/src/assets/wise5/components/component-student.component.ts
@@ -120,7 +120,7 @@ export abstract class ComponentStudent {
     this.subscriptions.add(
       this.ComponentService.notifyConnectedComponentSource$.subscribe(
         ({ nodeId, componentId, componentState }) => {
-          if (this.nodeId === nodeId && this.componentId === componentId) {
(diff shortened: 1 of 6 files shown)


src/assets/wise5/components/component-student.component.ts
```diff
@@ -115,6 +116,22 @@ export abstract class ComponentStudent {
     );
   }
 
+  subscribeToNotifyConnectedComponents(): void {
+    this.subscriptions.add(
+      this.ComponentService.notifyConnectedComponentSource$.subscribe(
+        ({ nodeId, componentId, componentState }) => {
+          if (this.nodeId === nodeId && this.componentId === componentId) {
```
hirokiterashima:
```suggestion
          if (nodeId === this.nodeId && componentId === this.componentId) {
```

src/assets/wise5/vle/node/nodeController.ts
```diff
@@ -879,6 +879,22 @@ class NodeController {
...
+            publishingComponent.componentId === componentState.componentId
+          ) {
+            this.ComponentService.notifyConnectedComponents(
+              this.nodeId,
+              subscribingComponent.id,
+              componentState
+            );
+          }
+        }
+      }
```
hirokiterashima: This code block is a bit confusing... It's just trying to find a component in this node to notify about a component state from its connected component, right? Maybe it's confusing because of the variable names subscribingComponent and publishingComponent. Not all components in this.getComponents() are subscribingComponents, so should it be replaced with component? And publishingComponent can be replaced with connectedComponent?

Also, we don't check for changedComponentId != null and componentState != null in this block like we do on like 898. Can we assume that those passed-in params will be non-null and remove the null checks?

hirokiterashima (changes requested):
Functionality works good. I have one code improvement suggestion and one question about a code block. See my comments inline.

src/assets/wise5/components/component-student.component.ts
geoffreykwan: I usually write it with this.nodeId on the left. Would you write
```this.nodeId === 'node1'```
or
```'node1' === this.nodeId```
?
For me, the first method of doing it reads more like regular English. Would you say "is my age equal to 10?" or "is 10 equal to my age?"?

hirokiterashima: Hmm. Doesn't the variable (what you're checking) usually go on the left side of the ```===``` and the constant (i.e., what you're checking against) go on the right? In this case, the variable is ```nodeId```, since it can be different each time, and the constant is ```this.nodeId```, since it remains the same for this component?

src/assets/wise5/vle/node/nodeController.ts
geoffreykwan: I've changed the variable names. I also changed the notifyConnectedComponents function name because that's what I was originally trying to make clearer. We're not notifying the connected components. We're actually notifying the component that subscribes to the connected component.

The code at 898 and below will be removed once all the components subscribe to the observable so we don't need to worry about that. The changedComponentId parameter probably won't be needed in the future either so we'll get rid of that later. We can get the nodeId and componentId from the componentState.

hirokiterashima (approved):

PR #78, opened 2021-04-14
Refactored code to use Subscription.add

### LINKED ISSUE #66: Refactor: use Subscription.add
### Description

Instead of creating a local variable for each subscription, use Subscription.add(obj.subscribe()). In the onDestroy(), call Subscription.unsubscribe() to unsubscribe from all subscriptions. This will remove the need to keep track of subscription variables and remembering to unsubscribe from each one.

hirokiterashima:
## Changes
- For components that had multiple observable subscriptions, I replaced them by adding the subscription to a shared Subscription via subscription.add(), and unsubcribed all at once in ngOnDestry().
- For components with single subscription, I left it the same.
- Fixed some subscriptions that were never unsubscribed.

## Additional changes
In those affected files, I also cleaned up the code:
- removed unused dependencies in controllers
- removed unnecessary variable declarations in controllers
- removed unsubscribeAll() functions and instead call unsubscribe() directly
- removed EditComponentController because it's no longer used

## Test
- Test all apps (student, teacher, author) and make sure that everything works as before. There is no new functionality, only refactored code.

Closes #66


geoffreykwan (approved):
Looks good.

PR #81, opened 2021-04-16
Make wiselink work again

### LINKED ISSUE #80: Make wiselink work again
wiselink in HTML components don't work anymore. This is probably because we converted HTML component to Angular and now use innerHTML to render the html. Make wiselink work again.

geoffreykwan:
Test that wiselink works.

Scenarios to test
- link to another step
- link to a component in another step
- link to a component in the current step
- button to another step
- button to a component in another step
- button to a component in the current step

Closes #80


src/assets/wise5/components/html/html-student/html-student.component.ts
```diff
@@ -34,11 +37,84 @@ export class HtmlStudent extends ComponentStudent {
...
+
+  goToNode(nodeId: string, componentId: string): void {
+    this.currentNodeChangedSubscription = this.StudentDataService.currentNodeChanged$.subscribe(
+      () => {
+        if (componentId != '') {
+          this.scrollToComponentAndHighlight(componentId);
+        }
+      }
+    );
+    this.StudentDataService.endCurrentNodeAndSetCurrentNodeByNodeId(nodeId);
```
hirokiterashima: Should the subscription and scroll/highlight occur outside of html-student component, like in a service? When the currentNode changes, this html-student component should no longer be alive, right? It seems strange that this subscription is alive and participating in events even after the user has navigated away from the component.

hirokiterashima (changes requested):
It looks like the highlight does not go away if the link is to the component that is on the same step.

Also, I wrote a question about code inline. Let me know what you think.

geoffreykwan:
The listener was being fired once for each HTML component on the step. Now the wiselinkclicked only fires once within that component. This fixed the problem of highlighting not going away.

We now listen for the currentNodeChanged event in the nodeService.ts.

I also fixed a bug where if you created a wiselink and then modified the html, the wiselink attribute would be lost and the wiselink would no longer work.

hirokiterashima (approved):
New changes look good.

PR #84, opened 2021-04-20
Disable Open Response textarea when all submits are used up

### LINKED ISSUE #83: Disable Open Response textarea when all submits are used up
Also add a note next to the max submits input in the Authoring Tool that informs the author the textarea will become disabled once all the submits are used.

geoffreykwan:
Test that the Open Response textarea and save/submit buttons becomes disabled once all the submits are used. Also leave the step and come back and make sure the Open Response is still disabled.

Closes #83


hirokiterashima (approved):
Looks good.

PR #86, opened 2021-04-20
Prevent copy pasting images into TinyMCE editor

### LINKED ISSUE #85: Prevent copy pasting images into TinyMCE editor
Prevent authors from copy-pasting images into the HTML component authoring. Copy-pasting was resulting in large base64 image strings being saved into the html content.

geoffreykwan:
- Try copy-pasting an image from a Word document. This should be prevented.
- Try drag-dropping an image from a webpage or local file. This should be prevented.
- Make sure regular copy-paste (for text and other basic html content) still works.
- You can still right click on a webpage image then choosing "Copy image" and then pasting the image into TinyMCE. This will create an image element with the src set to the absolute url to the image on the other website.

Here are the options for the [paste plugin](https://www.tiny.cloud/docs/plugins/opensource/paste) for reference.

Closes #85


hirokiterashima (approved):
Looks good.

PR #90, opened 2021-04-21
Make attachStudentAssets use an observable for Discussion

### LINKED ISSUE #76: Make attachStudentAssets use an observable for Discussion
In nodeController.ts we call showStudentAssets which uses componentToScope to obtain the child scope. The child scope componentController is then used in themeController.ts when it calls

this.StudentAssetService.showStudentAssets$.subscribe(({ componentController, $event })

Then in studentAssetController.ts we make calls to

componentController.attachStudentAsset(studentAsset)

We should change all this to use an observable instead of calling the attachStudentAsset function directly from the child component controller.

geoffreykwan:
1. Create a run that contains a Discussion component
2. Log in as a student and launch that run
3. Go to the Discussion component
4. Click the "Add picture" button to open the "My Files" popup
5. Add an image and choose it so it gets added to the post
6. Submit the post

Closes #76

### DIFF
--- a/src/assets/wise5/components/discussion/discussionController.ts
+++ b/src/assets/wise5/components/discussion/discussionController.ts
@@ -141,6 +141,7 @@ class DiscussionController extends ComponentController {
     }
     this.disableComponentIfNecessary();
     this.initializeScopeGetComponentState();
+    this.subscribeToAttachStudentAsset();
     this.registerStudentWorkReceivedListener();
     this.initializeWatchMdMedia();
     this.broadcastDoneRenderingComponent();
@@ -326,6 +327,16 @@ class DiscussionController extends ComponentController {
         }
       }
     }
+  }
+
+  subscribeToAttachStudentAsset() {
+    this.subscriptions.add(
+      this.StudentAssetService.attachStudentAsset$.subscribe((data) => {
+        if (this.isForThisComponent(data)) {
+          this.attachStudentAsset(data.asset);
+        }
+      })
+    );
   }
 
   registerStudentWorkReceivedListener() {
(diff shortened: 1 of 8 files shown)


src/assets/wise5/components/discussion/discussionController.ts
```diff
@@ -328,6 +329,16 @@ class DiscussionController extends ComponentController {
     }
   }
 
+  subscribeToAttachStudentAsset() {
+    this.subscriptions.add(
+      this.StudentAssetService.attachStudentAsset$.subscribe((data) => {
+        if (this.isForThisComponent(data)) {
+          this.attachStudentAsset(data.asset);
+        }
+      })
```
hirokiterashima: Instead of a generic "data" param, what do you think about creating a new StudentAssetRequest interface?

src/assets/wise5/services/studentAssetService.ts
```diff
@@ -13,6 +13,8 @@ export class StudentAssetService {
   audioFileExtensions = ['wav', 'mp3', 'ogg', 'm4a', 'm4p', 'raw', 'aiff', 'webm'];
   private showStudentAssetsSource: Subject<any> = new Subject<any>();
   public showStudentAssets$: Observable<any> = this.showStudentAssetsSource.asObservable();
+  private attachStudentAssetSource: Subject<any> = new Subject<any>();
+  public attachStudentAsset$: Observable<any> = this.attachStudentAssetSource.asObservable();
```
hirokiterashima: Instead of a generic "any" type, what do you think about creating a new StudentAssetRequest interface?

hirokiterashima (commented):
I noticed that when you upload a new image in the "My Files" dialog, it gets automatically added in the post compose box without having to choose it, and this bypasses the new code (discussionController's new subscription). The new code is run only when you choose an already-uploaded asset. Is this the intended behavior?

Also, I think it'd improve code readability if we created a new type for the data that is being passed through the attachStudentAsset$, like StudentAssetRequest. 

What do you think?

hirokiterashima (approved):
Looks good.

PR #95, opened 2021-04-22
Make wiselinks work with component annotations

### LINKED ISSUE #94: Make wiselinks work with component annotations
Make the new wiselinks work in the component annotations comment.

geoffreykwan:
1. Author a CRater step that contains feedback
2. Put a wiselink in the feedback
```<wiselink node-id='node1' link-text='First Step'></wiselink>```
3. Preview the project or sign in as a student to the run
4. Submit an answer so you receive the feedback
5. Make sure the wiselink works when the annotation is first created
6. Make sure the wiselink works when you revisit the step
7. Make sure wiselinks still work in the HTML component

Closes #94

### DIFF
--- a/src/assets/wise5/directives/componentAnnotations/component-annotations.component.ts
+++ b/src/assets/wise5/directives/componentAnnotations/component-annotations.component.ts
@@ -63,13 +63,11 @@ export class ComponentAnnotationsComponent {
   ngAfterViewInit() {
     this.processAnnotations();
     this.wiseLinkCommunicator = document.getElementById(this.wiseLinkCommunicatorId);
-    if (this.wiseLinkCommunicator != null) {
-      this.wiseLinkClickedHandler = this.WiseLinkService.createWiseLinkClickedHandler(this.nodeId);
-      this.WiseLinkService.addWiseLinkClickedListener(
-        this.wiseLinkCommunicator,
-        this.wiseLinkClickedHandler
-      );
-    }
+    this.wiseLinkClickedHandler = this.WiseLinkService.createWiseLinkClickedHandler(this.nodeId);
+    this.WiseLinkService.addWiseLinkClickedListener(
+      this.wiseLinkCommunicator,
+      this.wiseLinkClickedHandler
+    );
   }
 
   ngOnChanges(changes: SimpleChanges) {
@@ -80,9 +78,7 @@ export class ComponentAnnotationsComponent {
 
   ngOnDestroy() {
     this.studentWorkSavedToServerSubscription.unsubscribe();
-    if (this.wiseLinkClickedHandler != null) {
-      this.wiseLinkCommunicator.removeEventListener('wiselinkclicked', this.wiseLinkClickedHandler);
-    }
+    this.wiseLinkCommunicator.removeEventListener('wiselinkclicked', this.wiseLinkClickedHandler);
   }
 
   processAnnotations(): void {
--- a/src/assets/wise5/directives/componentAnnotations/component-annotations.component.ts
+++ b/src/assets/wise5/directives/componentAnnotations/component-annotations.component.ts
@@ -57,7 +57,7 @@ export class ComponentAnnotationsComponent {
         }
       }
     );
-    this.wiseLinkCommunicatorId = `wise-link-communicator-component-annotations-${this.componentId}`;
+    this.wiseLinkCommunicatorId = `wise-link-communicator-component-annotations-${this.nodeId}-${this.componentId}`;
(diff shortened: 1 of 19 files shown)

### CODE REVIEW

File: src/assets/wise5/directives/componentAnnotations/component-annotations.component.ts [hunk 1]
```diff
@@ -45,34 +57,88 @@ export class ComponentAnnotationsComponent {
...
       }
     );
+    this.wiseLinkCommunicatorId = `wise-link-communicator-component-annotations-${this.nodeId}-${this.componentId}`;
   }
 
-  ngOnChanges() {
+  ngAfterViewInit() {
     this.processAnnotations();
+    this.wiseLinkCommunicator = document.getElementById(this.wiseLinkCommunicatorId);
+    if (this.wiseLinkCommunicator != null) {
```
hirokiterashima: Is this null check necessary? It looks like wiseLinkCommunicator is always available in the DOM because it's in the template?

src/assets/wise5/directives/componentAnnotations/component-annotations.component.ts
```diff
@@ -45,34 +57,88 @@ export class ComponentAnnotationsComponent {
...
+
+  ngOnChanges(changes: SimpleChanges) {
+    if (!changes.annotations.isFirstChange()) {
+      this.processAnnotations();
+    }
   }
 
   ngOnDestroy() {
     this.studentWorkSavedToServerSubscription.unsubscribe();
+    if (this.wiseLinkClickedHandler != null) {
```
hirokiterashima: Is this null check necessary? It looks like the handler will always be created in ngAfterViewInit?

src/assets/wise5/directives/componentAnnotations/component-annotations.component.ts
```diff
@@ -45,34 +57,88 @@ export class ComponentAnnotationsComponent {
...
+  getNodeId(annotations: any): string {
+    return this.hasCommentAnnotation(annotations)
+      ? annotations.comment.nodeId
+      : annotations.score.nodeId;
+  }
+
+  getComponentId(annotations: any): string {
+    return this.hasCommentAnnotation(annotations)
+      ? annotations.comment.componentId
+      : annotations.score.nodeId;
```
hirokiterashima: Should this be ```annotations.score.componentId;``` ? 

I see that it is nodeId in the original code. It looks wrong, but doesn't seem to affect the behavior.

hirokiterashima (commented):
Functionality looks good. I had a few questions about code- see inline comments.

src/assets/wise5/directives/componentAnnotations/component-annotations.component.ts
geoffreykwan: I added nodeId and componentId as input parameters to component-annotations. Currently only multiple-choice-student passes those two parameters in. All the other components don't pass in nodeId and componentId in yet which is why we need the null check. We could either change all components to pass in nodeId and componentId or keep the null check until we eventually do pass in nodeId and componentId for all components after we've converted all the components to Angular.

src/assets/wise5/directives/componentAnnotations/component-annotations.component.ts
geoffreykwan: The null check is there for the same reason above.

src/assets/wise5/directives/componentAnnotations/component-annotations.component.ts
geoffreykwan: Ah, I missed that. I've fixed it now.

src/assets/wise5/directives/componentAnnotations/component-annotations.component.ts
hirokiterashima: I see. It looks like this.wiseLinkCommunicator will never be null even if nodeId and componentId are not passed in though. And it seems to work without any problems when I remove the null checks in ngAfterViewInit() and ngOnDestroy().

this.wiseLinkCommunicatorId becomes "wise-link-communicator-component-annotations-undefined-undefined" and this.wiseLinkCommunicator is a div element.

So we should either:
1. Use the undefined id (remove null-checks) <- might cause problem with multiple components that are not MC.
2. Check for nodeId/componentId null <- short-term solution, need to remember to remove null-checks later.
3. Pass in nodeId and componentId to all the component-annotations. <- Eventual solution. No need to remember to remove null-checks later.

hirokiterashima (approved):
Looks good.

PR #98, opened 2021-04-23
Created Node class

### LINKED ISSUE #97: Create Node class and move getNodeIcon function
### Description

Create a Node class.

Then move ```getNodeIcon()``` function from ProjectService to the new Node class so that we can replace call to
```
ProjectService.getNodeIconByNodeId(nodeId)
```
with
```
ProjectService.getNode(nodeId).getIcon()
```

hirokiterashima:
## Changes

- Created Node class. It does not have all the fields and functions yet, but we can add them gradually.
- Moved ```ProjectService.getNodeIconByNodeId()``` function to ```Node.getIcon()``` and updated references.
- Created ```ProjectService.getNode(nodeId)```, which returns a Node instance. Note that this is different from ```this.ProjectService.getNodeByNodeId(nodeId)```, which returns a JSON object. We can gradually replace all calls with the new function.
- Added unit tests.

## Test

- Update node icons in the AT by adding/editing the icons JSON block for a step or group.
Here's a sample icons JSON block:
```
"icons": {
        "default": {
            "fontName": "info",
            "color": "#2196F3",
            "fontSet": "material-icons",
            "type": "font"
        }
    }
```

- Verify that node icons appear correctly in CM, AT, VLE.
- Verify that there is no noticeable degradation of performance.

#97


src/assets/wise5/services/projectService.ts
```diff
@@ -648,6 +649,10 @@ export class ProjectService {
     return null;
   }
 
+  getNode(nodeId: string): Node {
+    return Object.assign(new Node(), this.getNodeById(nodeId));
```
geoffreykwan: Should we cache the Node object each time it's generated for a given nodeId?

src/assets/wise5/common/Node.ts
```diff
@@ -0,0 +1,27 @@
...
+    if (this.icons != null && this.icons.default != null) {
+      icon = $.extend(true, defaultIcon, this.icons.default);
+    } else {
+      icon = defaultIcon;
+    }
+    if (!icon.imgSrc) {
+      icon.type = 'font';
+    }
+    return icon;
+  }
```
geoffreykwan: Does the node icon ever change? If not, maybe we should save the icon object to an instance variable. If it does, we can leave it as is. getIcon() gets called for every node in the project whenever you go to another step.

geoffreykwan (changes requested):
Functionality works well. I posted a couple questions below.

hirokiterashima:
I cached the Node and NodeIcon values in private fields.

geoffreykwan (approved):
Looks good.

PR #103, opened 2021-04-27
Use Node in NodeAuthoringController.

### LINKED ISSUE #102: Refactor: use Node in NodeAuthoringComponent
### Description

Now that we have a Node class, use it in the NodeAuthoringComponent. It should be bound to the NodeAuthoringController. 

Define Node.isGroup() and replace calls like ```projectController.isGroupNode($ctrl.node.id)``` with ```$ctrl.node.isGroup()```

Clean up surrounding code.

hirokiterashima:
## Changes
- Set ```node: Node``` and ```projectId: number``` in resolve of the UI router paths so they are available in the controllers as bindings.
- Renamed old ```node``` to ```nodeJson``` in nodeAuthoringController until we have a good solution for updating node in cache. This controller now contains both ```node: Node``` and ```nodeJson: any```.
- Defined Node.isGroup() and updated references.
- Removed unused functions and dependencies.
- Updated path from '/wise5' to '/assets/wise5'

## Test
- Test that the AT works as before, in particular node authoring, node advanced authoring, and asset authoring.

Closes #102


src/assets/wise5/authoringTool/node/advanced/node-advanced-authoring.component.html
```diff
@@ -3,25 +3,25 @@
...
   </md-button>
-  <md-button ng-if='!$ctrl.isGroupNode($ctrl.node.id)'
+  <md-button ng-if='!$ctrl.node.isGroup()'
       class='topButton md-raised md-primary'
       ng-click='$ctrl.showEditConstraintsView()'>
     <md-icon>block</md-icon>
     <md-tooltip md-direction='top' class='projectButtonTooltip'>{{ ::'editConstraints' | translate }}</md-tooltip>
   </md-button>
-  <md-button ng-if='!$ctrl.isGroupNode($ctrl.node.id)'
+  <md-button ng-if='!$ctrl.node.isGroup()'
```
geoffreykwan: Should the isGroup value be stored in a variable in the controller and then be used in the template so the template doesn't have to call the isGroup() function so many times? This would cut down on the number of times the template watcher needs to evaluate the ng-if since it would be a variable and not a function call.

src/assets/wise5/authoringTool/node/nodeAuthoring.html
```diff
@@ -26,39 +26,39 @@
       </md-tooltip>
     </md-button>
     <md-button id='addComponentButton'
-        ng-if='$ctrl.showStepButtons && !projectController.isGroupNode($ctrl.node.id)'
+        ng-if='$ctrl.showStepButtons && !$ctrl.node.isGroup()'
```
geoffreykwan: Maybe change all isGroup() calls in this file to reference a boolean variable instead too.

geoffreykwan (changes requested):
Functionality works well. I made a suggestion about the isGroup() function call below.

src/assets/wise5/authoringTool/node/advanced/node-advanced-authoring.component.html
hirokiterashima: Good suggestion. I made the change.

src/assets/wise5/authoringTool/node/nodeAuthoring.html
hirokiterashima: Good suggestion. I made the change.

geoffreykwan (approved):
Looks good.

PR #108, opened 2021-04-29
Teacher score and comment does not override automated score and comment

### LINKED ISSUE #96: Teacher score and comment does not override automated score and comment
1. Set up a run that contains a CRater Open Response component that gives automated score and feedback.
2. Sign in as a student in one browser and and the teacher of that run in another browser (or incognito).
2. As the student, submit an answer. You should receive an automated score and comment.
3. As the teacher of the run, go the Grading Tool and find the student work you just created and change the score and comment.
5. Refresh the student view. You should see the teacher score and comment but you will still see the automated score and comment.

Here's a CRater item if you need one

Item Id: GREENROOF-II

5 points
The plants get the light energy from the sun and transform the energy to a new kind of energy called chemical energy. Also carbon dioxide, water, and the light energy come together and make the plants food, a sugar named glucose. They say that glucose is so energy rich because the light energy that was turned into chemical energy can be stored int the glucose for later and thats why these people have plants growing on there roof.

4 points
The energy comes from the sun and the energy transfers from place to place by bouncing off one another. The energy goes into the plant and is stored and is stored as chemical energy. Energy changes from light energy to chemical energy by the other energy's together..

3 points
energy comes from pre-existing energy that transforms, energy is stored in molecules. Energy explodes and the chemicals inside change.

2 points
it would lower the energy use because it goes t the plants before it gets to the building

1 point
hello world

geoffreykwan:
1. Set up a run that contains a CRater Open Response component that gives automated score and feedback.
2. Sign in as a student in one browser and and the teacher of that run in another browser (or incognito).
2. As the student, submit an answer. You should receive an automated score and comment.
3. As the teacher of the run, go the Grading Tool and find the student work you just created and change the score and comment.
5. Refresh the student view. You should see the teacher score and comment (previously you would still see the automated score and comment).

Closes #96

### DIFF
--- a/src/assets/wise5/services/annotationService.ts
+++ b/src/assets/wise5/services/annotationService.ts
@@ -280,7 +280,8 @@ export class AnnotationService {
   getTotalScoreForWorkgroup(annotations = [], workgroupId = -1) {
     let totalScore = 0;
     const scoresFound = []; // to prevent double counting if teacher scored component multiple times
-    for (const annotation of this.sortByNewestToOldest(annotations)) {
+    for (let i = annotations.length - 1; i >= 0; i--) {
+      const annotation = annotations[i];
       if (annotation.toWorkgroupId === workgroupId) {
         if (this.isScoreOrAutoScore(annotation)) {
           const nodeId = annotation.nodeId;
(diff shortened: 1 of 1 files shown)


hirokiterashima (approved):
Looks good.

PR #113, opened 2021-04-30
Use Node in NodeController

### LINKED ISSUE #111: Refactor: use Node in NodeController and node.html
### Description

Now that we have a Node class, use it in the NodeController and node.html. It should be bound to the NodeController. To keep the change set scope small, only change files that are used in the student VLE.

Clean up surrounding code.

hirokiterashima:
## Changes
- Created new functions in Node and updated references.
- Removed unused functions and unnecessary null checks.
- Moved bulk of code in the constructor into $onInit()
- I removed this if-block below because the node in NodeController should always be set, and should be an application node (i.e., not a group node). When you go to the 'unit view' as a student, NodeController is not used. Lmk if you see a problem with this change.
```
   if (
      this.StudentDataService.getCurrentNode() &&
      this.ProjectService.isApplicationNode(this.StudentDataService.getCurrentNodeId())
    ) {
```
- No new functionality was added.

## Test
- Test that student vle works as before, especially saving and reloading work in the components.

Closes #111

### DIFF
--- a/src/assets/wise5/common/Node.ts
+++ b/src/assets/wise5/common/Node.ts
@@ -1,7 +1,11 @@
 export class Node {
+  components: any[] = [];
   icons: any;
   icon: any;
   id: string;
+  rubric: any;
+  showSaveButton: boolean;
+  showSubmitButton: boolean;
   title: string;
   type: string;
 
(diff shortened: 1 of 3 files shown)


geoffreykwan (approved):
Looks good.

PR #114, opened 2021-04-30
Fix snip image from all component

### LINKED ISSUE #112: Fix snip image from all component
Enable Clipping in the Notebook Settings. Students should be able to click on an image and save it to their Notebook. Nothing happens when you click on an image now. This used to work but doesn't work anymore.

geoffreykwan:
1. Enable Clipping in the Notebook Settings in the Authoring Tool for a run.
2. Author an image into an HTML component. Also add an image into other components in their prompts as well as other possible places like choice in Multiple Choice and choice in Match.
2. Log in as a student.
3. Launch the run.
4. Click on images to add them as a Notebook note.

Closes #112

### DIFF
--- a/src/assets/wise5/services/notebookService.ts
+++ b/src/assets/wise5/services/notebookService.ts
@@ -96,10 +96,7 @@ export class NotebookService {
 
   createSnipImageHandler(): any {
     return (event: CustomEvent) => {
-      const imageObject = this.UtilService.getImageObjectFromImageElement(event.detail.target);
-      if (imageObject != null) {
-        this.addNote(imageObject);
-      }
+      this.addNote(this.UtilService.getImageObjectFromImageElement(event.detail.target));
     };
   }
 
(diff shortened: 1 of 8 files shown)


src/assets/wise5/services/notebookService.ts
```diff
@@ -76,17 +78,34 @@ export class NotebookService {
...
+    this.snipImageHandler = this.createSnipImageHandler();
+    window.addEventListener('snip-image', this.snipImageHandler);
+  }
+
+  createSnipImageHandler(): any {
+    return (event: CustomEvent) => {
+      const imageObject = this.UtilService.getImageObjectFromImageElement(event.detail.target);
+      if (imageObject != null) {
+        this.addNote(imageObject);
+      }
```
hirokiterashima: Can we ensure that calls to UtilService.getImageObjectFromImageElement() won't return a null object? It looks like the function returns null iff the param is null, so can we remove the null-check and assume that it is always called with a non-null param?

hirokiterashima (commented):
Functionality looks good. I had a question about code, see comments inline.

src/assets/wise5/services/notebookService.ts
geoffreykwan: Yea, I removed the null check.

hirokiterashima:
Looks good.

PR #116, opened 2021-05-04
Refactored handleStudentWorkSavedToServer()

### LINKED ISSUE #74: Call handleStudentWorkSavedToServer with the component state
Right now we call it like this

handleStudentWorkSavedToServer({ studentWork: componentState });

but we can change it to

handleStudentWorkSavedToServer(componentState);

since we only use the component state and no other fields.

hirokiterashima:
## Changes
- handleStudentWorkSavedToServer() now expects a componentState instead of ```{ studentWork: componentState }```.
- ```StudentService.studentWorkSavedToServer``` observable now also passes componentStates directly instead of wrapping it in an object.
- No new functionality was added.

## Test
Test that saving and loading student work in the student VLE works as before.

Closes #74

### DIFF
--- a/src/assets/wise5/components/component-student.component.ts
+++ b/src/assets/wise5/components/component-student.component.ts
@@ -144,14 +144,13 @@ export abstract class ComponentStudent {
 
   subscribeToStudentWorkSavedToServer(): void {
     this.subscriptions.add(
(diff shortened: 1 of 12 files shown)


geoffreykwan (approved):
Looks good.

PR #119, opened 2021-05-05
Fix node scores for workgroups in the Teacher Tools

### LINKED ISSUE #107: Teams that haven't submitted work for a component are given a 0 in the Teacher Tools
Teams that haven't submitted work for a component show up as having a score of 0 in the Teacher Tools grading views. Those 0 scores are also incorrectly included in the node's mean score display.

breity:
Show "-" for workgroup step score in the Teacher Tools if they workgroup has not received a score for any component in the step. (We used to show 0 as the score.) This also fixes the node average score that we display in the Grade by Step view for a node.

To test:
- Set up a run that has an autoscored (e.g. c-rater) item in a step and at least one other step that captures student work.
- Add some students to the run and complete the autoscored item and the other step's work for at least one but not all of the students.
- As the teacher, give scores to some but not all of the students in the second step.
- Verify that the student(s) who didn't receive scores are given a "-" instead of a 0 in the Score column (both in the Grade by Team and Grade by Step views).
- In the Grade by Step view, verify that the mean score for each step only takes into account students who have received a score on a component in that step.

Resolves #107.

### DIFF
--- a/src/app/services/annotationService.spec.ts
+++ b/src/app/services/annotationService.spec.ts
@@ -66,9 +66,9 @@ function getTotalScore() {
 }
 
 function getTotalScore_noAnnotationForWorkgroup_return0() {
-  it('should return 0 when no annotations for workgroup', () => {
-    expect(service.getTotalScoreForWorkgroup([], WORKGROUP_1)).toEqual(0);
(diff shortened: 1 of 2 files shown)


hirokiterashima (approved):
Looks good.

PR #120, opened 2021-05-06
Keep existing style attributes when inserting snip to notebook code

### LINKED ISSUE #118: Injecting snip to notebook code to images overwrites existing style attributes
### Description
**Observed behavior:** If snipping image to notebook is enabled in a unit, existing `style` attributes for images in the unit's content will be overwritten with `style="cursor: pointer;"`. 

**Expected Behavior:** Any images with `style` attributes set should retain them when the click action is injected.

### Additional Information
Image snipping should also be keyboard accessible (snip on Enter keypress when focused, add `tabindex="0"` to the image as well as an `aria-label` for screen readers).

breity:
Also added keyboard accessibility for clipping images to notebook.

To test:
- Make sure "Enable Clipping" is checked in notebook settings for a unit.
- Add an image to an HTML component.
- Add a custom `style` attribute to the image using either the TinyMCE code view or image settings (e.g. `style="border: 1px solid;"`).
- Preview the unit (or run as student) and verify that your custom styles are retained by the image.
- Make sure that clicking on the image still initiates the new note dialog.
- Also ensure that you can tab to focus on the image using the keyboard and that pressing the Enter key when focused initiates the new note dialog.

Resolves #118.

### DIFF
--- a/src/assets/wise5/services/projectService.ts
+++ b/src/assets/wise5/services/projectService.ts
@@ -607,7 +607,11 @@ export class ProjectService {
         return matchedString.replace(
           '<img',
           `<img onclick=\\"window.dispatchEvent(new CustomEvent('snip-image', ` +
-            `{ detail: { target: this } }))\\" snip`
+            `{ detail: { target: this } }))\\" ` +
+            `onkeypress=\\"javascript: if (event.key === 'Enter' || event.keyCode === 13 || ` + 
+            `event.which === 13 ) { window.dispatchEvent(new CustomEvent('snip-image', ` +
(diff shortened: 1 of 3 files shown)


geoffreykwan (approved):
Looks good.

hirokiterashima:
Hmm. This doesn't work for me. I see the keyboard accessibility improvement, but the border on my image does not get saved in the notebook.

hirokiterashima:
Clarified with @breity that this PR fixes the issue where styles on images were being overwritten. It does not keep the styles when the images is snipped to the notebook.

hirokiterashima (approved):
Looks good.

PR #122, opened 2021-05-07
Upgraded CM and AT StepTools to Angular

### LINKED ISSUE #115: Convert AT CM StepTools to Angular
Combine StepTools in CM and AT into one.
Upgrade it to Angular, then downgrade it to work with the hybrid AngularJS application.

hirokiterashima:
## Changes
- Combined StepTools in AT and CM into one.
- Fixed issue where going back from node editing view to project view using the back button did not reset title in the step tools.
- Code clean up: removed ununsed code, added types to function input parameters.

## Test
- You can navigate to steps using the prev/next arrow and drop-down in the CM and AT.
- First time opening the AT shows "Select a Step" in the drop-down.
- Going back to project view in the AT using the "Back to Unit Plan" arrow show "Select a Step".

## Other notes
- Please style as you see fit.
- There are some instances where the step title still does not reset to "Select a Step" (e.g. when you click on the "unit home" icon). This is an existing behavior and should be fixed in the future.

Closes #115.


breity (approved):
Looks good.

I removed jQuery usage for detecting RTL/LTR layout and replaced with Angular Material's [BidiModule](https://material.angular.io/cdk/bidi/overview). Also merged from develop to resolve conflicts.

PR #129, opened 2021-05-10
Extracted component save and submit button

### LINKED ISSUE #128: Refactor: extract component save and submit buttons
### Description

The save and submit buttons are duplicated across the components' templates. Extract this into a common component.

hirokiterashima:
## Changes
- Extracted the save and submit buttons to one component, ComponentSaveSubmitButtons
- Updated components that have been already converted to angular: Label, MC, and AudioOscillator
- No new functions were added

## Test
- Enable component save/submit buttons for Label, MC and AudioOscillator, and test that they work as before.

## Notes
- The existing functions saveButtonClicked and submitButtonClicked on ComponentStudent are called from ComponentSaveSubmitButtons. Ideally this would be defined in the new ComponentSaveSubmitButtons, but right now the body of those functions use other fields and functions in ComponentStudent, so it was not easy to do this. This should be another refactoring task for the future.

Closes #128


breity (approved):
Looks good.

I made minor style/layout changes and removed old wise5 classnames and unnecessary `aria-label` attributes.

PR #132, opened 2021-05-11
Extracted AddToNotebook button

### LINKED ISSUE #131: Refactor: extract component add to notebook button
### Description

The "Add to Notebook" button is duplicated across the components' templates. Extract this into a common component.

hirokiterashima:
## Changes
- Extracted AddToNotebook button
- Cleaned up surrounding code and removed unneeded comments
- No new feature was added

## Test
- Test that you can snip a label to the notebook like before

## Notes
- Please style as you see fit. 
- I kept ```margin-right:20px``` style at the label component level instead of moving it to the new component, thinking that this will allow more flexibility for future components (instead of all AddToNotebook button having 20px margin). Maybe the button layout can be better handled with a fxLayoutGap?

Closes #131


breity:
> * I kept `margin-right:20px` style at the label component level instead of moving it to the new component, thinking that this will allow more flexibility for future components (instead of all AddToNotebook button having 20px margin). Maybe the button layout can be better handled with a fxLayoutGap?

Yep, should update the label component to use fxLayoutGap instead of the margins. fxLayoutGap doesn't add additional css and handles LTR + RTL languages properly. I think this can be part of #146.

breity (approved):
Looks good.

I updated the button styles to match existing add to notebook buttons for now. Also merged from develop to resolve conflicts.

PR #142, opened 2021-05-13
Refactored Teacher[Data/Project]Service

### LINKED ISSUE #141: Refactor TeacherDataService and TeacherProjectService
### Description

TeacherProjectService
- Move getNodeIdComponentIds to Node.

TeacherDataService
- Change retrieveStudentDataForNode() to take in Node instead of nodeId and update references
- Change
```
const x = getValue();
if (x != null) {
  return x;
}
return [];
```
with
```
return getValue() || [];
```
- Use array.find and array.some where applicable

hirokiterashima:
## Changes
- Moved getNodeIdComponentIds() to Node class
- Used Node class as param to retrieveStudentDataByNodeId instead of nodeId
- Use array.some/find where applicable.
- Changed to ```return variable || defaultValue``` syntax

## Test
- Node grading view works as before: show work, grade work
- [Un]pause screens works as before

Closes #141

### DIFF
--- a/src/app/services/teacherProjectService.spec.ts
+++ b/src/app/services/teacherProjectService.spec.ts
@@ -49,7 +49,6 @@ describe('TeacherProjectService', () => {
   testDeleteTransition();
   testGetNodeIdAfter();
   testCreateNodeAfter();
-  shouldGetTheNodeIdAndComponentIdObjects();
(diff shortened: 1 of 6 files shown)


geoffreykwan (approved):
Looks good.

PR #153, opened 2021-05-17
Change node icon in StepTools when node changes

### LINKED ISSUE #130: Step Tools node icon does not change in AT/CM
The node icon on the toolbar doesn't change when you switch nodes (using the arrows or drop-down) in AT/CM. It's working in student.

breity:
## Changes
- Updated NodeIcon to get icon from ProjectService whenever inputs change and the `icon` input variable is not sent.

Fixes #130.

## Test
- That node icon in the step tools toolbar updates to reflect the current node being viewed (when using both the select drop-down and the navigation arrows).


hirokiterashima (approved):
Looks good.

PR #156, opened 2021-05-18
Support enabling teacher notebook when student notebook is disabled

### LINKED ISSUE #155: Teacher notebook can't be enabled if student notebook is disabled
## Description
If the student notebook is disabled for a run unit, the teacher notebook cannot be enabled.

## Steps to Reproduce
1. In the authoring tool for a run unit, disable the student notebook and enable the teacher notebook (`/teacher/edit/unit/[runid]/notebook`).
2. Switch to the Teacher Tools for the run and check if the teacher notebook report is shown.

### Expected Behavior
The notebook report should be visible to the teacher.

### Observed Behavior
The teacher notebook report is not visible.

breity:
## Changes
The teacher notebook can now be enabled for a run if the student notebook is disabled.

Closes #155.

## Test
- Disabled student notebook and enable teacher notebook for a run unit in the authoring tool.
- Switch to Teacher Tools and verify that teacher notebook report is visible and functions.
- Try out all the different combinations for turning on and off the teacher and student notebooks and make sure the settings work properly.

### DIFF
--- a/src/assets/wise5/classroomMonitor/classroomMonitorController.ts
+++ b/src/assets/wise5/classroomMonitor/classroomMonitorController.ts
@@ -77,8 +77,10 @@ class ClassroomMonitorController {
     this.projectId = this.ConfigService.getProjectId();
     this.runId = this.ConfigService.getRunId();
     this.runCode = this.ConfigService.getRunCode();
-    this.notebookConfig = this.NotebookService.getTeacherNotebookConfig();
-    this.reportEnabled = this.notebookConfig.enabled;
+    if (this.NotebookService.isNotebookEnabled('teacherNotebook')) {
+      this.notebookConfig = this.NotebookService.getTeacherNotebookConfig();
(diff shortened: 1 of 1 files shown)


hirokiterashima:
I wonder if the original code should have been
```
if (this.NotebookService.isNotebookEnabled('teacherNotebook')) {
```
The null check in that function may be a backwards-compatible check, for projects that did not have a teacherNotebook block in the json?

And ```this.reportEnabled``` should really be looking at ```teacherNotebook.itemTypes.report.enabled``` instead of ```this.notebookConfig.enabled```. 🤦

breity:
@hirokiterashima Yep, thanks for catching that. The null check in `this.NotebookService.isNotebookEnabled('teacherNotebook')` should be included. I added that back.

Looks like `this.reportEnabled` actually needs both `this.notebookConfig.enabled` and `this.notebookConfig.itemTypes.report.enabled` to be true. Because the unit could have the report enabled but the teacher notebook disabled.

I updated with both of these changes.

hirokiterashima (approved):
Looks good.

PR #170, opened 2021-06-01
Exclude workgroups no longer in run from student statuses

### LINKED ISSUE #166: Step completion % in CM is miscalculated when workgroups are removed from run
## Description
When a teacher moves all the students from a workgroup with an existing studentStatus to another workgroup, the percent completion values for nodes that the old workgroup visited are miscalculated. `StudentStatusService.getNodeCompletion` takes into account all `studentStatus` entries, even for workgroups that are no longer active.

## Steps to Reproduce
- Start a run and add student 1. Complete some work for that student.
- Add student 2 to the run (in a new workgroup) and visit some of the steps in the unit.
- As the teacher, move student 2 into student 1's workgroup.

### Observed Behavior
Completion percentage for the steps that student 2 visited will be incorrectly calculated. The percent completion will be calculated using both the inactive workgroup's completion status and the existing workgroup's completion status for the node.

For example:
image

image

Percent completion is showing as 50% when all the teams have completed the step.

### Expected Behavior
Percent completion should only take into account active workgroups in the run. It should be 100% for the example above.

breity:
## Changes
Excludes workgroups that are no longer active in a run from being included in the student statuses for the run. This resolves a bug that resulted in incorrect percent completion displays for nodes in the Classroom Monitor.

Closes #166.

## Test
- Start a run and add student 1. Complete some work for that student.
- Add student 2 to the run (in a new workgroup) and visit some of the steps in the unit.
- As the teacher, move student 2 into student 1's workgroup.
- Verify that percent completion for the nodes in the run is accurate (e.g. shows 100% for nodes that student 1 has completed).

### DIFF
--- a/src/assets/wise5/services/studentStatusService.ts
+++ b/src/assets/wise5/services/studentStatusService.ts
@@ -24,10 +24,15 @@ export class StudentStatusService {
       .get(`/api/teacher/run/${this.ConfigService.getRunId()}/student-status`)
       .toPromise()
       .then((studentStatuses: any) => {
+        const workgroups = this.ConfigService.getClassmateUserInfos();
         for (const studentStatus of studentStatuses) {
           const parsedStatus = JSON.parse(studentStatus.status);
-          parsedStatus.postTimestamp = studentStatus.timestamp;
-          this.studentStatuses.push(parsedStatus);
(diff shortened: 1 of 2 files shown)


hirokiterashima (approved):
looks good.

PR #178, opened 2021-06-07
Only show steps with work in Teacher Tools node select dropdown

### LINKED ISSUE #177: Node select dropdown in Teacher Tools should only show steps with work
## Description
In Teacher Tools (CM), the node select dropdown in `StepTools` should only show steps that capture student work.

## Step to Reproduce
1. Open the Teacher Tools for a run.
2. In Grade by Step, open the node select dropdown from the toolbar.

### Observed Behavior
All of the nodes in the unit are shown in the select dropdown.

### Expected Behavior
Only group (lesson) nodes and steps that capture student work should be shown, since there's no need to open the grading view for a node that doesn't capture any work.

breity:
## Changes
Added option to filter out steps that don't capture work in the node select dropdown in the StepTools component. Teacher Tools (CM) now only shows steps with work.

Resolves #177.

## Test
- Open the Teacher Tools for a run.
- In Grade by Step, open the node select dropdown from the StepTools in the toolbar. Verify that only steps that capture work are shown.
- Open the Authoring Tool for the run.
- Verify that all steps are shown in the node select dropdown in the toolbar.

### DIFF
--- a/src/assets/wise5/authoringTool/components/shared/toolbar/toolbar.ts
+++ b/src/assets/wise5/authoringTool/components/shared/toolbar/toolbar.ts
@@ -52,7 +52,7 @@ const Toolbar = {
           <md-icon> menu </md-icon>
         </md-button>
         <span class="toolbar__title" ng-if="!$ctrl.showStepTools">{{ $ctrl.viewName }}</span>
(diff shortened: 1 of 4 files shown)


hirokiterashima (approved):
LGTM

PR #183, opened 2021-06-07
Changed retrieveStudentStatuses() to use subscribe().

### LINKED ISSUE #182: Modify and test StudentStatusService.retrieveStudentStatuses()
### Description

- StudentStatusService.retrieveStudentStatuses() should use subscribe() instead of .toPromise().then().
- StudentStatusService.retrieveStudentStatuses() is not tested properly. Test that it is called and correctly parses the returned studentStatuses array.

hirokiterashima:
## Changes
- Changed retrieveStudentStatuses() to use subscribe().
- Updated test to ensure that old workgroups are properly filtered out, to test work done in #170.
- Also updated test to ensure that studentStatuses field was properly set after http call completes.

## Test
- Code and test are easy to read and understand
- CM loads properly

Closes #182

### DIFF
--- a/src/app/services/studentStatusService.spec.ts
+++ b/src/app/services/studentStatusService.spec.ts
@@ -25,9 +25,9 @@ describe('StudentStatusService', () => {
         UtilService
       ]
     });
-    http = TestBed.get(HttpTestingController);
-    service = TestBed.get(StudentStatusService);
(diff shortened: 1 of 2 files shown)


breity (approved):
Working well, code looks good.

PR #193, opened 2021-06-10
Replace all nodeId and componentId references when importing a step

### LINKED ISSUE #192: Replace all nodeId and componentId references when importing a step
When importing a step, give the nodes and components new node ids and component ids and also replace all references to the old node ids and component ids.

geoffreykwan:
Test importing a step that imports work.
1. Open a unit in the Authoring Tool. I'll refer to this unit as unit 1.
2. Create a Draw step that contains two Draw components. Make sure to turn on the Submit button for the first Draw component.
3. Make the second Draw component import work from the first Draw component.
4. Open a new unit in the Authoring Tool.
5. Import the Draw step from unit 1.

Test adding a lesson structure (like the Jigsaw since it has branching and is more complex) to a unit.

Closes #192


hirokiterashima (approved):
LGTM. We should refactor teacherProjectService[spec].ts in a separate issue. It's starting to get big.

PR #194, opened 2021-06-10
Authoring Tool can't click Home button and then go back to step previously opened

### LINKED ISSUE #188: Authoring Tool can't click Home button and then go back to step previously opened
2. Click on a step to go in to it (for example Step 1.1)
3. Click the "Unit Home" button
4. Click on the step you were previously in (for example Step 1.1). You will not be able to go in to that step. You can go to other steps though. Once you go to a different step, you can then go back in to Step 1.1.

geoffreykwan:
2. Create a new step and insert it into the unit
3. Click the step drop down and make sure you can see the new step within the drop down. The step used to not show up here.
4. Click the "Unit Home" button on the left side bar
5. Click on the step you just created. You should be taken into the step authoring view. Previously nothing would happen and you would just stay in the unit home view.

Closes #188

### DIFF
--- a/src/assets/wise5/authoringTool/node/nodeAuthoringComponent.ts
+++ b/src/assets/wise5/authoringTool/node/nodeAuthoringComponent.ts
@@ -135,6 +135,7 @@ class NodeAuthoringController {
   }
 
   $onDestroy() {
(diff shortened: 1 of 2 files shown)


hirokiterashima (approved):
Looks good.

PR #196, opened 2021-06-10
Extracted and upgraded maxScore grading to Angular

### LINKED ISSUE #195: Upgrade component maxScore grading to Angular
### Description

Upgrade component maxScore grading component to Angular component

hirokiterashima:
## Changes
- Extracted maxScore grading component from ComponentGrading.
- Removed unnecessary maxScore bindings from parent components.
- Cleaned up ComponentGrading

## Test
- Updating max score in grading views (grade by step, grade by team) works
- Max score appears correctly in component revisions popup

## Note
- There is a very slight height difference between the score and max score input boxes. I think this because the score is md-input-container (angularjs) and the max score is mat-form-field (angular). I could not figure out how to align them. @breity can you take a look? If it's takes too much time, perhaps it's best to leave it as is and spend the time upgrading the score into angular component instead.

Closes #195


src/assets/wise5/classroomMonitor/classroomMonitorComponents/grading-edit-component-max-score/grading-edit-component-max-score.component.ts
```diff
@@ -0,0 +1,52 @@
...
+    );
+  }
+
+  ngOnDestroy() {
+    this.subscriptions.unsubscribe();
+  }
+
+  saveMaxScore() {
+    const maxScore = this.UtilService.convertStringToNumber(this.maxScore);
+    if (typeof maxScore === 'number' && maxScore >= 0) {
```
geoffreykwan: Is the typeof maxScore === 'number' check required?

geoffreykwan (changes requested):
Overall it's working well. There are 2 things I noticed.

- If you delete the max score value, nothing happens. If the teacher deletes the max score value should we then set it to null or 0? In the Authoring Tool when this happens we set it to null.

- If you change the max score in the Grade by Team view, the whole step flashes when it updates. This does not happen in the Grade by Step view.

src/assets/wise5/classroomMonitor/classroomMonitorComponents/grading-edit-component-max-score/grading-edit-component-max-score.component.ts
hirokiterashima: If you delete the max score in the input, this.maxScore becomes null, and maxScore >= 0 evaluates to true. UtilService.setMaxScoreForComponent does a null-check and won't save it, so perhaps it is redundant?

hirokiterashima:
> * If you delete the max score value, nothing happens. If the teacher deletes the max score value should we then set it to null or 0? In the Authoring Tool when this happens we set it to null.

Set to null to keep it consistent with the AT?

> * If you change the max score in the Grade by Team view, the whole step flashes when it updates. This does not happen in the Grade by Step view.

This is a behavior that's existed from before. I think it's outside the scope of this PR and would be better handled in a separate issue?

geoffreykwan:
>Set to null to keep it consistent with the AT?

Yea, I think we should set the max score to null when the teacher deletes the max score value.

> This is a behavior that's existed from before. I think it's outside the scope of this PR and would be better handled in a separate issue?

Yea, we can handle the flashing behavior in a separate issue.

src/assets/wise5/classroomMonitor/classroomMonitorComponents/grading-edit-component-max-score/grading-edit-component-max-score.component.ts
hirokiterashima: Removed unnecessary checks.

geoffreykwan (approved):
Looks good.

PR #200, opened 2021-06-11
Extracted CopyNodesService

### LINKED ISSUE #199: Extract ProjectService.copyNodes into CopyNodesService
### Description

Extract ProjectService.copyNodes() into its own service class CopyNodesService, and have it return an array of nodes only. The subscribe should be in choose-import-step-location, where it takes the copied nodes, replaces the id's, and then insert them into the specified location.

This will make the code easier to read and test.

hirokiterashima:
## Changes
- extracted copyNodes() in TeacherProjectService to its own CopyNodesService class. CopyNodesService.copyNodes() simply calls http post to copy the nodes and returns an observable
- created InsertNodesService to insert nodes into project
- refactored ChooseImportStepLocation.importSelectedNodes() to use the new classes
- no new feature was added

## Test
- import step works as before
- import component works as before
- ChooseImportStepLocation.importSelectedNodes() is easy to understand

Closes #199


src/assets/wise5/services/teacherProjectService.ts
```diff
@@ -448,86 +450,11 @@ export class TeacherProjectService extends ProjectService {
...
-           *
-           * This means after we add nodeA, we must update
-           * nodeIdToInsertInsideOrAfter to be nodeA so that when we add
-           * nodeB, it will be placed after nodeA.
-           */
-          nodeIdToInsertInsideOrAfter = newNode.id;
-        }
-        return newNodes;
-      });
+  getNewNodeIds(nodes: any[]): any[] {
```
geoffreykwan: The name of this function gives the impression that it returns node ids. Maybe change it to something like getNodesWithNewNodeIds?

geoffreykwan (changes requested):
Functionality works well. I just have one minor function name change suggestion.

src/assets/wise5/services/teacherProjectService.ts
hirokiterashima: I agree. I renamed the function.

geoffreykwan (approved):
Looks good.

PR #206, opened 2021-06-14
Student data that contains an html image element includes additional attributes that shouldn't be saved

### LINKED ISSUE #201: Student data that contains an html image element includes additional attributes that shouldn't be saved
For components that contain \ elements in the saved student data, remove the attributes that we inject into them when we save the student data. This can occur in Multiple Choice, Match, and possibly other types of components.

Here is an example of Match student data that is saved.
```
{
	"id": "1dxic353e7",
	"type": "bucket",
	"value": "<img onclick=\"window.dispatchEvent(new CustomEvent('snip-image', { detail: { target: this } }))\" onkeypress=\"javascript: if (event.key === 'Enter' || event.keyCode === 13 || event.which === 13 ) { window.dispatchEvent(new CustomEvent('snip-image', { detail: { target: this } } )) }\" aria-label=\"Select image to add to notebook\" title=\"Add to notebook\" tabindex=\"0\" snip src=\"/curriculum/69/assets/project_thumb.png\"/>",
	"items": [{
		"feedback": "",
		"isIncorrectPosition": null,
		"id": "hjemcz6n2y",
		"type": "choice",
		"value": "<img onclick=\"window.dispatchEvent(new CustomEvent('snip-image', { detail: { target: this } }))\" onkeypress=\"javascript: if (event.key === 'Enter' || event.keyCode === 13 || event.which === 13 ) { window.dispatchEvent(new CustomEvent('snip-image', { detail: { target: this } } )) }\" aria-label=\"Select image to add to notebook\" title=\"Add to notebook\" tabindex=\"0\" snip src=\"/curriculum/69/assets/project_thumb.png\"/>",
		"isCorrect": null
	}]
}
```

geoffreykwan:
Test Match
1. Create a match component that has a choice value and a bucket value that contain an image.
2. Create student work on that component.
3. Make sure the student data saved to the database does not contain the img attributes we inject (see bad and good below).

Test Multiple Choice
1. Create a Multiple Choice component that has a choice that contains an image.
2. Create student work on that component.
3. Make sure the student data saved to the database does not contain the img attributes we inject (see bad and good below).

Bad

Good

Closes #201

### DIFF
--- a/src/assets/wise5/components/match/match-student/match-student.component.ts
+++ b/src/assets/wise5/components/match/match-student/match-student.component.ts
@@ -616,6 +616,11 @@ export class MatchStudent extends ComponentStudent {
     return componentState;
   }
 
+  /**
+   * @param originalComponentContent The component content that has not had any additional content
+   * injected into it such as onclick attributes and absolute asset paths.
+   * @param buckets
+   */
   cleanBuckets(originalComponentContent: any, buckets: any): any {
     for (const bucket of buckets) {
       bucket.value = this.getCleanedValue(originalComponentContent, bucket);
@@ -627,6 +632,11 @@ export class MatchStudent extends ComponentStudent {
     return buckets;
   }
 
+  /**
+   * @param originalComponentContent The component content that has not had any additional content
+   * injected into it such as onclick attributes and absolute asset paths.
+   * @param matchObj
+   */
   getCleanedValue(originalComponentContent: any, matchObj: any): string {
     return this.getValueById(originalComponentContent, matchObj.id) ?? matchObj.value;
   }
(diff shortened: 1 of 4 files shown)


src/assets/wise5/components/match/match-student/match-student.component.ts
```diff
@@ -611,15 +618,40 @@ export class MatchStudent extends ComponentStudent {
...
     return buckets;
   }
 
+  getCleanedValue(matchObj: any): string {
+    const value = this.getValueById(this.originalComponentContent, matchObj.id);
+    if (value == null) {
+      return matchObj.value;
+    } else {
+      return value;
+    }
```
hirokiterashima: Use nullish coalescing for brevity?
```suggestion
    return this.getValueById(this.originalComponentContent, matchObj.id) ?? matchObj.value;
```

hirokiterashima (changes requested):
Functionality works. I have few comments:

- I added an inline comment about nullish coalescing.
- Also, I wonder if it's better to get the originalComponentContent right before we actually use it (in cleanBuckets() and pass in as second parameter to getCleanedValue()?), instead of getting and setting it in ngOnInit()? It would remove the possibility of being accessed/modified by another piece of code, and reduce the time to run ngOnInit().

What do you think?

hirokiterashima (approved):
LGTM

PR #212, opened 2021-06-16
Extracted EditComponentScore and upgraded to Angular

### LINKED ISSUE #211: Extract and upgrade component score grading to Angular
### Description

Extract and upgrade component maxScore grading component to an Angular component.

hirokiterashima:
## Changes
- Extracted EditComponentScore and upgraded to Angular
- Also removed @Input from GradingEditComponentMaxScore since it's not an input

## Test
- Items without score should show 0.
- Verify auto-score from c-rater shows with the label "auto score".
- You can edit the auto-score. Doing so will change the label to "item score".
- See past scores in the revisions dialog. All the inputs should be disabled, and the labels should say "teacher score" or "auto score".
- You can remove a score, which will set it to null (and show "0" in the UI).

Closes #211


geoffreykwan (approved):
Looks good.

PR #214, opened 2021-06-16
Default feedback annotation

### LINKED ISSUE #198: Default feedback annotation
### Description

Allow authors to specify default feedback on any component that has work in the component's advanced authoring view.
They can specify one or more strings (which can contain html). Each string will map to the submit count. ["feedback 1 for submit 1", "feedback 2 for submit 2", ...].

When the student submits, the default feedback will be chosen, and a new annotation will be created. The annotation will have data.autoGrader = "defaultFeedback". 

A new annotation will only created if there is a default feedback associated with the submit count. For example, if there are 2 default feedbacks, a third submit will not create a new default feedback annotation

geoffreykwan:
- Author default feedback for all component types that can submit work.
- Make sure the default feedback shows up when the student submits work.
- The default feedback should be overridden by any other feedback such as CRater.

Closes #198


hirokiterashima (changes requested):
Functionality looks good.

edit-component-default-feedback.component.spec.ts will introduce duplications. Can you look into refactoring it a bit? Maybe use a helper function to compare if the expected array is equal to the defaultFeedback array?

```
function expectDefaultFeedbackEquals(expectedArray: string[]) {
  expectArrayEquals(component.authoringComponentContent.defaultFeedback, expectedArray);
}

function expectArrayEquals(a: string[], b: string[]) {
  a.forEach((val, index) => {
    expect(val).toEqual(b[index]);
  });
}
```

hirokiterashima (approved):
LGTM

PR #215, opened 2021-06-18
Extracted and upgraded edit component comment to Angular

### LINKED ISSUE #213: Extract and upgrade component comment grading to Angular #211
### Description

Extract and upgrade component comment grading component to an Angular component.

hirokiterashima:
## Changes
- Renamed editComment() to toggleEditComment, and removed timeout.
- Cleaned up componentGrading.ts.

## Test
- You can save comments
- You can see past comments in the revisions dialog. The comments should be disabled (can't be edited)
- For auto-scored comments, you can toggle the focus to edit comment textarea using the "Edit" link

## Notes
- This will create some duplications with the EditComponentScoreComponent. We can extract the two components to a parent EditComponentAnnotationComponent in a future task.

Closes #213


geoffreykwan (changes requested):
Main functionality works well. I have some additional thoughts.

- I've always felt a debounce of 5 seconds is too long. If the teacher types a comment and closes the student work or goes to another step before 5 seconds, their comment is not saved. Maybe it should be something shorter like 1 or 2 seconds and/or maybe save on focusout/blur?

- When there is an auto comment and the teacher clicks "Edit", should the auto comment be populated into the teacher comment textarea? That way they have the choice of editing it or deleting everything from the auto comment.

- This is more of a corner case that is less likely to occur. If you have an auto comment and click "Edit" and then type something in the "Teacher Comment" textarea but then change your mind and don't want to save your "Teacher Comment" so you click the "Edit" again to toggle off the "Teacher Comment", your "Teacher Comment" will still save after the debounce.

geoffreykwan (approved):
Looks good.

PR #234, opened 2021-06-29
Timeout CRater requests after 40 seconds

### LINKED ISSUE #233: Timeout CRater requests after 40 seconds
Implement a timeout for C-Rater requests. If the request does not return within 40 seconds, show a message to the student to try again, and move on if it continues to fail.

hirokiterashima:
## Changes
- Added a timeout for CRater requests after 40 seconds
- Changed CRaterRequests to return an observable instead of a promise

## Test
- CRater times out after 40 seconds of no-response from the backend. The student sees a message, and the data is still saved. The data should not be a submit, and the submit counter should remain the same. The student should be able to submit again, or move to the next activity.
- CRater works properly when it responds before 40 seconds.

Closes #233


geoffreykwan (approved):
Looks good.

PR #248, opened 2021-07-08
Extracted delete/move node and copy project functions

### LINKED ISSUE #247: Extract delete/move node and copy project functions
### Description

These are functions in the TeacherProjectService that can be extracted into their own service classes.

hirokiterashima:
## Changes
- Extracted delete/move node and copy project functions from teacherProjectService to new services
- Added function param types and return values
- Moved CopyProject message dialog to appear before making POST request
- Added CopyProject error handling

## Test
- Delete node(s) works as before
- Move node(s) works as before
- Copy project works as before. If the request fails (debug WISE-API -> make the AuthorAPIController.copyProject function throw error, or stop wise-api docker service before making request), you should see an alert message and the "Copying..." dialog should close.

Closes #247


geoffreykwan (approved):
Looks good.

PR #250, opened 2021-07-08
Manage students display workgroups

### LINKED ISSUE #245: Manage Students: display workgroups
### Description

In the Manage Student re-write v1:
- display the workgroups in the period
- switching between periods should show workgroups for the period
- show period name, workgroup and student counts should appear at the top

image

Takumu678:
## Changes

- Created manage students page with team info component and user info component
- Renamed old manage students page to manage students legacy

## Test

- Get the latest from develop branch on wise-api
- New manage student page displays the workers and the members in them
- Period number of students and number of teams appear at the top of the page
- Switching periods in the drop bar updates the view

Closes #245


geoffreykwan (changes requested):
Selecting individual periods works properly but "All Periods" doesn't seem to work.

Screenshot 2021-07-08 184537

hirokiterashima:
@geoffreykwan good catch. Having a view for All Periods wouldn't make sense for Manage Students page, so we can either:

1. Show a message like "Managing students in all periods is not currently supported. Please choose a specific period."
2. Not show the All Periods drop down option in this case and choose the first period in the drop down if it was selected before coming to this view

I think option 1 is simpler to implement and consistent with the UI in the rest of the views. What do you think?

hirokiterashima:
@geoffreykwan We opted for showing the message when All Periods is selected. We also created a PeriodInfo component and wrote unit tests. Can you re-review?

geoffreykwan (approved):
Looks good.

PR #254, opened 2021-07-12
Extract importComponents into its own service

### LINKED ISSUE #253: Extract TeacherProjectService.importComponents into its own service
### Description

Move this function into a new service and update references.

hirokiterashima:
## Changes
- Extract importComponents into its own service
- Refactored toPromise code to use InsertComponentService.insertComponents()
- Refactored choose-component-location by removing extra importComponents() function definition

## Test
- Import component (via node editing view) works as before
  - import as first component
  - import after existing component

Closes #253


geoffreykwan (approved):
Looks good.

PR #272, opened 2021-07-26
Fixed manage students period select dropdown updating issues

### LINKED ISSUE #271: Manage students: period select dropdown updating issues
### Description
The period select drop down has these issues related to the new manage students tool

**Period select does not update number of teams**
1. Move a team to another period.
2. The period select does not update the number of teams in each period.

**Can't view team that is moved to empty period**
1. Move a team to another period that has no teams.
2. You will not be able to view the period that the team was moved to because the period drop down still thinks there are 0 teams in that period.
3. The team and students in the team will no longer appear anywhere even if you refresh the page.

hirokiterashima:
## Changes
- Period select correctly updates number of teams
- Can view team that is moved to an empty period
- Moved empty teams to bottom of manage periods view
- Cleaned up code

## Test
- Move a team to another period. The period select should update the number of teams in each period.
- Move a team to another period that has no teams. You should be able to go to that period and see the team.

Closes #271

### DIFF
--- a/src/app/services/sampleData/sample_classmateUserInfos.json
+++ b/src/app/services/sampleData/sample_classmateUserInfos.json
@@ -1,0 +1,12 @@
+[
+  {
+    "workgroupId": 7,
+    "periodId": 1,
+    "users": [
+      { "username": "student0103" },
+      { "username": "student0104" }
+    ]
+  },
+  { "workgroupId": 6, "periodId": 2, "users": [{ "username": "student0102" }] },
+  { "workgroupId": 5, "periodId": 1, "users": [{ "username": "student0101" }] }
+]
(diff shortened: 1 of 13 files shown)


geoffreykwan (changes requested):
I found a couple bugs.

- An empty team shows up in all the periods.

- Can't move student into an empty team.

- For an empty team you can click on the "Change Period" link but the period select in the popup does not work. Maybe disable the "Change Period" link for empty teams since it doesn't really make sense?

- Can't go to period with 0 teams. Should be able to go to Period with 0 teams to see empty teams.

hirokiterashima:
@geoffreykwan good catch. I addressed all those issues.

geoffreykwan (approved):
Looks good.

PR #275, opened 2021-07-28
Authoring Tool previous and next buttons are disabled after adding a step

### LINKED ISSUE #274: Authoring Tool previous and next buttons are disabled after adding a step
Note: this problem only occurs for steps added after the first step during your authoring session

1. Add a step
2. Add another step
3. The previous and next buttons are disabled when they should not be

geoffreykwan:
1. Add a step
2. Add another step
3. The previous and next buttons should not be disabled

Test adding a step
- as the first step in the project
- as the last step in the project
- as a middle step in the project

Closes #274


hirokiterashima (approved):
LGTM

PR #280, opened 2021-07-29
Import Draw into Concept Map sometimes does not work

### LINKED ISSUE #279: Import Draw into Concept Map sometimes does not work
1. Create a Draw component
2. Create a Concept Map component on another step
3. Make the Concept Map component import the draw as a background
4. When running the project, the Draw will sometimes not be imported into the Concept Map. This will happen particularly if you do work on the Draw step and then visit the Concept Map more than once without doing any work on it.

geoffreykwan:
1. Create a Draw component
2. Create a Concept Map component on another step
3. Make the Concept Map component import the draw as a background
4. Preview the project and make sure the Draw gets imported as the background to the Concept Map
5. Also test it with a student in a run

Also make sure Concept Map works
- with no background
- with a predefined background

Closes #279

### DIFF
--- a/src/assets/wise5/components/conceptMap/concept-map-student/concept-map-student.component.html
+++ b/src/assets/wise5/components/conceptMap/concept-map-student/concept-map-student.component.html
@@ -56,8 +56,8 @@
       </div>
       <svg id="{{svgId}}"
           class="concept-map-svg"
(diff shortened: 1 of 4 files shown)


Takumu678 (approved):
looks good

hirokiterashima (approved):
Looks good.

PR #296, opened 2021-08-18
Import work from previous step sometimes does not import

### LINKED ISSUE #285: Import work from previous step sometimes does not import
1. Create a step with a component type that generates work (like Open Response)
2. Create another step with the same component type and have it import work from the previous step
3. Preview the first step
4. Type work into the first step but do not click the save button
5. Click the next button to go to the next step (make sure to do this before it auto saves)
6. The second step will not import the work from the first step because the first step did not save the work in time

geoffreykwan:
1. Create a step with a component type that generates work (like Graph)
2. Create another step with the same component type and have it import work from the previous step
3. Preview the first step
4. Generate work into the first step but do not click the save button
5. Click the next button to go to the next step (make sure to do this before it auto saves)
6. The second step should import the work (previously it would not import the work because the first step did not save the work in time)

Closes #285


hirokiterashima (approved):
Looks good. I removed the redundant function from discussion.

PR #339, opened 2021-09-17
UI changes to Dialog Guidance

### LINKED ISSUE #338: Dialog Guidance UI changes
Based on feedback from research testing:
- Limit width so that it looks more like a chatbox
- Once max number has been reached, hide response box
- Change arrow icon button to “Send” or add "Send" to button

breity:
## Changes
Made UI updates requested in #338.

Closes #338.

## Test
Make sure Dialog Guidance components still work as before.

### DIFF
--- a/src/assets/wise5/components/dialogGuidance/dialog-guidance-student/dialog-guidance-student.component.html
+++ b/src/assets/wise5/components/dialogGuidance/dialog-guidance-student/dialog-guidance-student.component.html
@@ -19,7 +19,8 @@
     <button mat-flat-button
         color="primary"
         (click)="submitStudentResponse()"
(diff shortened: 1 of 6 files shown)


hirokiterashima (changes requested):
UI changes look good.

Instead of calling hasMaxSubmitCountAndUsedAllSubmits() in the template (which calls it on each change detection cycle and can impact performance), can you calculate it and set it to a variable in the component?

[Here's an article stating the problem and offering some possible solutions](https://medium.com/showpad-engineering/why-you-should-never-use-function-calls-in-angular-template-expressions-e1a50f9c0496).

hirokiterashima (approved):

geoffreykwan (approved):
Looks good.

PR #343, opened 2021-09-21
Extract DialogGuidanceFeedbackRuleEvaluator

### LINKED ISSUE #342: Dialog Guidance extract feedback rule evaluator to new class
### Description

Extract the code that evaluates the feedback rule to its own class and add tests for various responses.

hirokiterashima:
## Changes
- Extract DialogGuidanceFeedbackRuleEvaluator
- "idea" changed to "rule" in component content feedback block

## Test
- change every "idea" to "rule" in feedbackRules block in the component content
- make sure that Guidance Dialog feedback matching works as before

Closes #342


geoffreykwan (approved):
Looks good.

PR #348, opened 2021-09-25
feat(Dialog Guidance): Add author feedback rules

### LINKED ISSUE #347: Dialog Guidance author feedback rules
Allow authors to 
- add, edit, delete, and order rules
- specify rule expression and feedback text
- specify max submission count

hirokiterashima:
## Changes
- Allow authors to add/edit/delete/order DG Feedback Rules
- Specify max number of submits

## Test
- If you haven't already, change FeedbackRule.rules array to to FeedbackRule.expression string in component content
- Add/edit/delete/order DG feedback rules
- Specify max number of submits

@breity please style as you see fit. It would be great if each feedback rule didn't take up so much space, so we can fit more on the screen. Also, have some kind of indication that the items can be reordered by drag&drop.

Closes #347


breity (approved):
Looks good. I updated styles, added up/down arrows to move rules, added support for scrolling while dragging.

@hirokiterashima Can you review these changes before merging? Thanks!

PR #383, opened 2021-10-26
feat(Data Export): Implement Embedded table export

### LINKED ISSUE #381: Implement export for Embedded table
Implement the component export for the Embedded table in this step [5.8: Impacts from Different Devices](https://wise.berkeley.edu/preview/unit/34569/node265).

Place the logic in the dataExportController.ts. Add a tag to the component in the project content and then look for the tag to determine when this Embedded table export can be used for an Embedded component.

geoffreykwan:
Make sure the "Export Component Data" works properly when exporting data saved in the Embedded table on this step.

In order for this device table to be exportable, the component must have the tag devices-kwh-co2-table
[3.3 A: Investigation #2: Device Usage Plan](https://wise.berkeley.edu/preview/unit/32968/node162)
You can use the updated model (that properly saves every time) that I've placed in the shared curriculum folder [appliances-new.html](https://wise.berkeley.edu/curriculum/shared/appliances-kwh-co2-table/v1/appliances-new.html). If you want to download it locally, you will need the html, js, and styles.css files.

In order for this transportation table to be exportable, the component must have the tag transportation-co2-table
[3.3 B: Investigation #3: Transportation Plan](https://wise.berkeley.edu/preview/unit/32968/node163)
You can use the updated model (that properly saves every time) that I've placed in the shared curriculum folder [transportation-new.html](https://wise.berkeley.edu/curriculum/shared/transportation-co2-table/v1/transportation-new.html). If you want to download it locally, you will need the html, js, and styles.css files.

1. Launch the Classroom Monitor for the run
2. Go to the Data Export view
4. Find the Embedded component
5. Click the "All" button and make sure it exports all the work
6. Click the "Latest" button and make sure it exports the latest work

Make sure the "Export Component Data" works for other components that can be exported such as Match or Discussion.

### DIFF
--- a/src/assets/wise5/classroomMonitor/dataExport/dataExportController.ts
+++ b/src/assets/wise5/classroomMonitor/dataExport/dataExportController.ts
@@ -11,8 +11,8 @@ import { TeacherProjectService } from '../../services/teacherProjectService';
 
 class DataExportController {
   $translate: any;
-  availableComponentAllRevisionsDataExports = ['DialogGuidance', 'Discussion', 'Match'];
-  availableComponentLatestRevisionsDataExports = ['DialogGuidance', 'Match'];
+  allowedComponentTypesForAllRevisions = ['DialogGuidance', 'Discussion', 'Match'];
+  allowedComponentTypesForLatestRevisions = ['DialogGuidance', 'Match'];
   componentExportTooltips = {};
   componentExportDefaultColumnNames = [
     '#',
@@ -51,6 +51,12 @@ class DataExportController {
   dialogGuidanceRevisionLabel: string = 'Revision';
   dialogGuidanceScoreLabel: string = 'Score';
   dialogGuidanceStudentResponseLabel: string = 'Student Response';
+  embeddedTableKeyToValue = {
+    co2saved: 'CO2 Saved',
+    current: 'Current',
+    future: 'Future',
+    kwhsaved: 'kWh Saved'
+  };
   exportStepSelectionType: string = 'exportAllSteps';
   exportType: string = null; // type of export: [latestWork, allWork, events]
   includeAnnotations: boolean;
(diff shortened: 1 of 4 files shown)


src/assets/wise5/classroomMonitor/dataExport/dataExportController.ts
```diff
@@ -2594,27 +2600,31 @@ class DataExportController {
...
-        return true;
-      }
+  isDataSpring20201EmbeddedComponentAndCanExport(component: any) {
+    if (
+      component.type === 'Embedded' &&
+      component.tags != null &&
+      component.tags.includes('DataSpring20201')
+    ) {
+      return true;
     }
```
hirokiterashima:
```suggestion
  isDataSpring20201EmbeddedComponentAndCanExport(component: any): boolean {
    return component.type === 'Embedded' && component.tags != null && 
        component.tags.includes('DataSpring20201');
```
This function should always return a boolean value. Right now it can return true or void.

hirokiterashima (changes requested):
Can you put the embeddedComponentExport functions into another class instead of adding it to dataExportController? 

Also see the inline code change suggestion to clarify the ```isDataSpring20201EmbeddedComponentAndCanExport``` function.

src/assets/wise5/classroomMonitor/dataExport/dataExportController.ts
geoffreykwan: Ok, fixed the function.

I think we should move all the component specific exports into their own classes in another issue. All the different export types should also be in their own routes and classes (work export, events export, notebook export, etc) and done in another issue as well.

geoffreykwan:
I've added another Embedded table model (transportation) to test for exporting in the description.

hirokiterashima (approved):

PR #390, opened 2021-10-29
fix(Connected Components): Send work on save and submit

### LINKED ISSUE #389: Notify connected components on save and submit
Right now new student work is only sent to the connected components when studentDataChanged() is called. We also want to send the new student work to the connected components when the student work is saved or submitted.

geoffreykwan:
Test that submitting the Embedded table renders the graph.
https://wise.berkeley.edu/preview/unit/32968/node162

Make sure to use the updated appliances table that saves on submit.
https://wise.berkeley.edu/curriculum/shared/appliances-kwh-co2-table/v1/appliances-new.html

Previously, submitting the updated appliances table would not trigger the graph rendering but now it should.

Also make sure other instances of connected components on the same step still work properly. For example [Thermodynamics Challenge - 2.8: Run Your Experiments](https://wise.berkeley.edu/preview/unit/32121/node81).

Closes #389

### DIFF
--- a/src/assets/wise5/vle/node/nodeController.ts
+++ b/src/assets/wise5/vle/node/nodeController.ts
@@ -536,6 +536,9 @@ class NodeController {
             componentEvents,
             componentAnnotations
           } = this.getDataArraysToSaveFromComponentStates(componentStatesFromComponents);
+          componentStates.forEach((componentState: any) => {
+            this.notifyConnectedParts(componentId, componentState);
+          });
           return this.StudentDataService.saveToServer(
             componentStates,
             componentEvents,


breity (approved):
Looks good.

For the Embedded table and graph activities (like this one: https://wise.berkeley.edu/preview/unit/32968/node162), the grading view shows every previous trial on the graph while the VLE only shows the most recent (which matches the table). We should implement individual submits/saves for connected graphs in the future.

As a temporary solution, we can check "Enable Trials" and "Hide Existing Trials on New Trial" in the authoring tool for the connected graph.

PR #405, opened 2021-11-12
fix(Events Export): Fix duplicate events bug

### LINKED ISSUE #404: Bug: export events duplicate rows
### Description

Events are duplicated for each workgroup in the run

### Steps to Reproduce

1. Export events for a run with at least 2 workgroups

**Expected behavior:**
Export should contain events for each workgroup

**Actual behavior:**
Export contains duplicate events for every workgroup in the run

geoffreykwan:
1. Go to a run that has at least 2 workgroups with events
2. Download the Events export
3. Make sure all the event rows are unique. If you copy an Event ID in the export (Column R) and search for it, you should only find it in one row.

The fix was a one liner but I also refactored some functions. This line with the workgroupId check was the only thing that was needed to fix the bug.
https://github.com/WISE-Community/WISE-Client/blob/3c0abe9c1c7b5afe7a9446ece25161f810de0983/src/assets/wise5/classroomMonitor/dataExport/dataExportController.ts#L987-L988

Closes #404

### DIFF
--- a/src/assets/wise5/classroomMonitor/dataExport/dataExportController.ts
+++ b/src/assets/wise5/classroomMonitor/dataExport/dataExportController.ts
@@ -963,32 +963,71 @@ class DataExportController {
     columnNameToNumber: any,
     events: any[]
   ): number {
-    const workgroups = this.ConfigService.getClassmateUserInfosSortedByWorkgroupId();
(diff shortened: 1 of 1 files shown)


breity (approved):
Looks good. 👍

PR #408, opened 2021-11-19
feat(Peer Group): Implement teacher manual assignment view

### LINKED ISSUE #407: Create view for manually assigning peer groups
The view should be opened in a popup from a button in the Peer Chat step grading view.

It should
- show all the groups
- show all unassigned workgroups
- allow creating a new group
- allow dragging and dropping workgroups into a different group

Here is a mockup of what it should roughly look like.
https://balsamiq.cloud/s8brvqj/p9j1not/r2370

geoffreykwan:
1. Log in as a teacher and open the Classroom Monitor for a run that contains a Peer Chat component
2. Go to the Grade By Step view for the step that contains the Peer Chat component
3. Click the "Peer Group" button near the top of the step to open the Peer Group view
4. The Groups in this view are mocked using the existing workgroups in the run. Nothing has been hooked up with the backend so nothing will save.
5. Test out moving the workgroups around
6. Test out creating a new group
7. Test out changing periods

Some questions
Should we allow the teacher to create groups across periods in the the All Periods view?
What are we going to show in the [Peer Chat activity details](https://balsamiq.cloud/s8brvqj/p9j1not/r2370) popup in the upper right?

Closes #407

### DIFF
--- a/src/assets/wise5/classroomMonitor/classroomMonitorComponents/nodeGrading/nodeGradingView/nodeGradingView.ts
+++ b/src/assets/wise5/classroomMonitor/classroomMonitorComponents/nodeGrading/nodeGradingView/nodeGradingView.ts
@@ -5,6 +5,7 @@ import { ConfigService } from '../../../../services/configService';
 import { NodeService } from '../../../../services/nodeService';
 import { MilestoneService } from '../../../../services/milestoneService';
 import { NotificationService } from '../../../../services/notificationService';
+import { PeerGroupService } from '../../../../services/peerGroupService';
 import { StudentStatusService } from '../../../../services/studentStatusService';
 import { TeacherDataService } from '../../../../services/teacherDataService';
 import * as angular from 'angular';
@@ -29,6 +30,7 @@ export class NodeGradingViewController {
   nodeHasWork: boolean;
   nodeId: string;
   numRubrics: number;
+  peerGroupComponentIds: string[];
   sortOrder: object = {
     team: ['-isVisible', 'workgroupId'],
     '-team': ['-isVisible', '-workgroupId'],
@@ -52,6 +54,7 @@ export class NodeGradingViewController {
     'MilestoneService',
     'NodeService',
     'NotificationService',
+    'PeerGroupService',
     'ProjectService',
     'StudentStatusService',
     'TeacherDataService'
@@ -64,6 +67,7 @@ export class NodeGradingViewController {
     protected MilestoneService: MilestoneService,
     protected NodeService: NodeService,
     protected NotificationService: NotificationService,
+    protected PeerGroupService: PeerGroupService,
     protected ProjectService: TeacherProjectService,
     protected StudentStatusService: StudentStatusService,
     protected TeacherDataService: TeacherDataService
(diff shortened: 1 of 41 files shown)


src/assets/wise5/classroomMonitor/classroomMonitorComponents/nodeGrading/nodeGradingView/nodeGradingView.ts
```diff
@@ -359,6 +364,14 @@ export class NodeGradingViewController {
...
     this.MilestoneService.showMilestoneDetails(this.milestoneReport, $event);
   }
+
+  showPeerGroupGroupings(nodeId: string, componentId: string): void {
+    this.NodeService.showPeerGroupDetails(
+      this.TeacherDataService.getCurrentPeriod().periodId,
+      nodeId,
+      componentId
+    );
+  }
```
hirokiterashima, line 374: What's the reason for putting showPeerGroupDetails() in NodeService when all it does is open a dialog? Can that be done in this function instead?

src/assets/wise5/classroomMonitor/classroomMonitorComponents/peer-group/peer-group-dialog/peer-group-dialog.component.ts
```diff
@@ -0,0 +1,209 @@
...
+      return 0;
+    }
+  }
+
+  createGrouping(id: number = this.nextAvailableGroupId, workgroups: any[]): any {
+    return {
+      id: id,
+      workgroups: workgroups
+    };
+  }
```
hirokiterashima: Instead of returning any, should we define a Grouping class and return that?

hirokiterashima (changes requested):
Looking good.

See inline comments and questions. My main suggestion is to create a new Groupings class and use that instead of "any". I think that would make the code more clear.

src/assets/wise5/classroomMonitor/classroomMonitorComponents/nodeGrading/nodeGradingView/nodeGradingView.ts
geoffreykwan: nodeGradingView.ts is old Angular. PeerGroupDialogComponent.ts is a new Angular component. Right now it's easier to open the dialog from NodeService.ts since NodeService is a new Angular service. Maybe we could downgrade PeerGroupDialogComponent and try to open it from the old Angular world if we really wanted to. Is it worth it?

src/assets/wise5/classroomMonitor/classroomMonitorComponents/peer-group/peer-group-dialog/peer-group-dialog.component.ts
geoffreykwan: This is code used for mocking and will be deleted later. We should create a PeerGroup class when we integrate with the real backend calls.

hirokiterashima (approved):

breity:
Looking good.
> Should we allow the teacher to create groups across periods in the the All Periods view?

I updated so that each period is distinct in the All Periods view and teacher can only create groups within period.
> What are we going to show in the [Peer Chat activity details](https://balsamiq.cloud/s8brvqj/p9j1not/r2370) popup in the upper right?

Yeah, let's discuss this. We can create a new issue for it.

One other thought - do we need the confirmation dialog when moving teams? I wonder if it's necessary for peer groups, seeing as it will only affect work for a specific component in this case. Or maybe could just clarify in the warning text that the work will still be visible to the previous group but not to the student(s) that move out of the group.

Also should probably only show the warning if the team has added any component states tied to the group. But we can create a new issue for this.

What do you think @geoffreykwan @hirokiterashima? Once we decide, I'll either remove the confirmation or update the warning text.

breity:
After chatting with @geoffreykwan, we decided to keep confirmation dialog for now. I updated the text to better clarify what happens when removing students from a group. I'll create a separate task to discuss whether we should only show if team has submitted work to the group.

breity (approved):

hirokiterashima (approved):

PR #439, opened 2021-12-22
feat(Error Handling): Filter errors from external requests

### LINKED ISSUE #437: Filter errors from external requests
## Description
HttpErrorInterceptor shows a generic toast message when intercepting Http Errors with status of 0 or > 500. Message tells user an error has occurred and instructs to reload the page.

For requests to resources external to WISE (https://wise-discuss.berkeley.edu Discourse instance, for example), don't show the toast message and handle error in the specific request callback, if desired. Rationale is that external failed requests don't generally break critical app functionality.

breity:
## Changes
Don't show toast message when external request results in an HttpError.

Closes #437.

## Test
Verify that generic toast message is shown for HttpErrors from internal requests but not for external requests like `https://wise-discuss.berkeley.edu/...`.

### DIFF
--- a/src/app/http-error.interceptor.ts
+++ b/src/app/http-error.interceptor.ts
@@ -9,10 +9,11 @@ import {
 import { Observable, throwError } from 'rxjs';
 import { catchError } from 'rxjs/operators';
 import { MatSnackBar } from '@angular/material/snack-bar';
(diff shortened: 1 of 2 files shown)


hirokiterashima (approved):
Looks good.

PR #444, opened 2022-01-05
chore(VLE): Convert component.ts to Angular

### LINKED ISSUE #443: Convert component.ts to Angular component
### Description

This component is used in the student vle to show the component (in node.html) and also when generating image using a dialog to be imported as a background into another component (in generate-image-dialog.html)

hirokiterashima:
## Changes
Removed unused attributes like mode and instead hard-code it to 'student' in the component itself since only students use this component.

## Test
- Launch preview and student vle
- Make sure all data loads: prompt, saved work (if any), imported work, teacher annotations
- Importing work as background (e.g. draw -> concept map)

Closes #443


geoffreykwan (approved):
Looks good.

PR #450, opened 2022-01-07
chore(Match): Remove AngularJSModule import

### LINKED ISSUE #449: Remove AngularJSModule dependency from MatchModule
### Description

Import only the modules that are used in the Match component instead.

hirokiterashima:
Now, only the necessary dependencies are imported in the MatchCommonModule.

Test that Match component works as before in the student and grading views.

Closes #449


breity (approved):
I added a few more Material dependencies. Looks good now.

PR #451, opened 2022-01-07
chore(VLE): Convert Generate Image Dialog

### LINKED ISSUE #448: Convert Generate Image Dialog to Angular

geoffreykwan:
Make sure importing a component as a background to another component works. For example, have a Draw component in one step and then on the next step have a Label component that imports the Draw as a background. Use connected component functionality to author this.

Closes #448


hirokiterashima (changes requested):
Importing with connected component works like before. 👍 

It looks like we can now remove the UpgradeModule dependency from component-student.component.ts (and all of its children classes and tests)?

hirokiterashima (approved):

PR #460, opened 2022-01-13
chore(VLE): Convert student account menu

### LINKED ISSUE #459: Convert accountMenu.html to an Angular component

geoffreykwan:
When signed in as a student, click the upper right avatar icon to show the account menu.

Make sure
- The username and workgroup id shows up. If there are multiple students in the workgroup, the names should be comma separated.
- The unit completion is displayed.
- If the unit has any max scores authored, make sure the total score is displayed.
- The "Go Home" button works.
- The "Sign Out" button works.

Closes #459


hirokiterashima (changes requested):
I can't log out from an inactive run. I can from production, so this looks like a new behavior.

Looks like this line in StudentDataService.saveToServer() is the culprit, but I haven't found a fix:
```
...
    } else if (!this.ConfigService.isRunActive()) {
      return this.upgrade.$injector.get('$q').defer().promise;
    } else {
...
```

hirokiterashima:
I can now logout from inactive runs. 👍 

It looks like unit completion changed from a bar to a circle. I think the circle is harder to see the progress because you can't see the entire "out-of" circle, especially when in the single digits. It takes a while for me to see that it's trying to draw a circle. How difficult would it be to add a lighter background circle outline to make it clearer?

Before PR:
Screen Shot 2022-01-20 at 8 52 24 AM
Screen Shot 2022-01-20 at 8 47 10 AM

PR:
Screen Shot 2022-01-20 at 8 47 18 AM
Screen Shot 2022-01-20 at 8 51 35 AM

breity:
@hirokiterashima I updated the progress indicators to both use Material progress bars.

I also fixed the progress indicator in the top bar, which was broken. This was caused when the following lines were removed from the ThemeController in an earlier commit, so I put them back.
```
this.rootNode = this.ProjectService.rootNode;
this.rootNodeStatus = this.nodeStatuses[this.rootNode.id];
```

Should be ready for review.

hirokiterashima (approved):
Looks good. Nice ui improvements. Easier to read and more fun.

PR #463, opened 2022-01-21
refactor(OpenResponse): Remove mode=student checks

### LINKED ISSUE #462: Remove mode=student from open-response-student
Having mode="student" check is redundant in open-response-student component. Remove it and make sure that everything works as before.

hirokiterashima:
## Changes
- Removed mode=student input and tests in open-response-student component because they're redundant. I think this is remnants from the time when a the teacher and student views shared the same component code. Now that we have separated out into -student, -grading, -authoring components, we shouldn't need to do these checks anymore.

## Test 
- OpenResponse component works as before
   - show work
   - import work
   - authoring
   - authoring preview
   - grading

Closes #462


geoffreykwan (commented):
Looks good

PR #480, opened 2022-02-02
feat(Authoring): Implement Peer Group Activity Tag authoring

### LINKED ISSUE #479: Implement Peer Group Activity Tag authoring
Implement it for all components that save student work.

geoffreykwan:
Make sure you can author the Peer Group Activity Tag in the advanced authoring view for components that save work.

Closes #479

### DIFF
--- a/src/app/authoring-tool/edit-component-peer-group-activity-tag/edit-component-peer-group-activity-tag.component.ts
+++ b/src/app/authoring-tool/edit-component-peer-group-activity-tag/edit-component-peer-group-activity-tag.component.ts
@@ -1,39 +1,52 @@
-import { Component } from '@angular/core';
+import { Component, ElementRef, Input, ViewChild } from '@angular/core';
 import { FormControl } from '@angular/forms';
 import { Observable } from 'rxjs';
 import { map, startWith } from 'rxjs/operators';
 import { TeacherProjectService } from '../../../assets/wise5/services/teacherProjectService';
-import { EditComponentFieldComponent } from '../edit-component-field.component';
 
 @Component({
   selector: 'edit-component-peer-group-activity-tag',
   templateUrl: './edit-component-peer-group-activity-tag.component.html',
   styleUrls: ['./edit-component-peer-group-activity-tag.component.scss']
 })
-export class EditComponentPeerGroupActivityTagComponent extends EditComponentFieldComponent {
+export class EditComponentPeerGroupActivityTagComponent {
+  @Input() authoringComponentContent: any;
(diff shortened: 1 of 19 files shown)


src/app/authoring-tool/edit-component-peer-group-activity-tag/edit-component-peer-group-activity-tag.component.ts
```diff
@@ -0,0 +1,48 @@
...
+    const tagsFound = new Map();
+    for (const component of this.projectService.getComponents()) {
+      const tag = component.peerGroupActivityTag;
+      if (tag != null && tag !== '' && !tagsFound.has(tag)) {
+        tags.push(tag);
+        tagsFound.set(tag, true);
+      }
+    }
+    return tags.sort();
+  }
```
hirokiterashima: Would it be cleaner if tagsFound is a Set<string> in this function? 
```suggestion
  getPeerGroupActivityTags(): string[] {
    const tagsFound = new Set();
    for (const component of this.projectService.getComponents()) {
      const tag = component.peerGroupActivityTag;
      if (tag != null && tag !== '') {
        tagsFound.add(tag);
      }
    }
    return Array.from(tagsFound).sort();
  }
```

src/assets/wise5/components/peerChat/edit-peer-chat-advanced-component/edit-peer-chat-advanced-component.component.ts
```diff
@@ -20,5 +20,7 @@ export class EditPeerChatAdvancedComponentComponent extends EditAdvancedComponen
     super(NodeService, NotebookService, ProjectService);
   }
 
-  ngOnInit(): void {}
+  ngOnInit(): void {
+    super.ngOnInit();
+  }
```
hirokiterashima: Not necessary?

```suggestion
```

hirokiterashima (changes requested):
Each time I type a letter in the peerGroupActivityTag input box, it creates a new row in the peer_group_activities table because it's saving changes per keystroke. Can you change this to save when input loses focus, or after a debounce?

Also, see code comments inline.

geoffreykwan:
I wonder if losing focus or debounce won't be good enough. Both of those can still lead to unwanted rows in the database.

Should we have a confirm dialog when adding a new tag?
Should we check for duplicate tags before saving a tag?
Should authors be able to delete tags?

hirokiterashima (approved):
Looks good.

PR #484, opened 2022-02-04
fix(Show Group Work): Show groups with no work

### LINKED ISSUE #483: ShowGroupWork: show groups with no work
### Description

In the Show Group Work view, show "No Response" for group members that has not done any work on the component. Right now, they do not show up at all in the view, only workgroups that have done work is displayed.

hirokiterashima:
## Changes
- Groups with no work will appear in the display and say "(No Response)"
- Also cleaned up the code so that setLayout() and setWorkgroupInfos() are not called as part of setStudentWorkFromGroupMembers().

## Test
- Groups with no work will appear in the display and say "(No Response)"

#483


breity (approved):
Looks good. I fixed width calculation based on number of workgroups in peer group and standardized heights for response containers.

PR #499, opened 2022-02-11
feat(Dialog Guidance): Implement computer avatars

### LINKED ISSUE #491: Implement Dialog Guidance avatar
- Allow author to enable computer-avatar in Dialog Guidance
- In student view, show computer-avatar chooser in Dialog Guidance html using ngIf
- Show avatar names next to each text response
- Add computer starter dialog to Dialog Guidance

[Design Doc](https://docs.google.com/presentation/d/1H2PqhmPgho2WqgQA5K08jkvH6IgHkOAPhuAbbYm01P8/edit#slide=id.g1101c85984c_1_0)

geoffreykwan:
## Changes
Add computer avatar to Dialog Guidance authoring
Add computer avatar selection to Dialog Guidance in student VLE

## Test
Authoring
- Enabling computer avatar in Dialog Guidance
- Authoring computer avatar fields (label, prompt, initial response) in Dialog Guidance
- Selecting which computer avatars to make available to students in Dialog Guidance

Student
- Make sure the student can select a computer avatar
- Make sure the computer avatar fields like label, prompt, and initial response show up properly
- Make sure choosing a computer avatar, leaving the step, and coming back, will re-populate the chosen computer avatar
- Make sure Dialog Guidance still works when computer avatar is not enabled

Grading
- Make sure computer avatars show up in Dialog Guidance grading view


breity (approved):
Looks great.

I updated the styles and UI and also changed the student view so that the avatar is automatically selected if there's only 1 available (as specified in the authoring).

hirokiterashima:
The functionality looked good. I refactored some code so that it would be simpler and easier to read (in my opinion). I tried to pass in the ComputerAvatar object (instead of computerAvatarId) where possible, introduced ComputerAvatarSettings class, and extracted EditDialogGuidanceComputerAvatarComponent from DialogGuidanceAuthoringComponent. The component content should now look like this:
```
...
    "isComputerAvatarEnabled": true,
    "computerAvatarSettings": {
         "label": "Best friend",
        "prompt": "Discuss your answer with with your best friend!",
        "initialResponse": "Aloha! What do you think about...",
        "ids": {
            "robot": true,
            "girl": true,
            "tiger": true
        }
   }
...
```

Let me know if you have any thoughts about these changes.

@geoffreykwan 

1. When no initial avatar is selected in the authoring tool, we still save it. Should we not allow this, make sure the author chooses initial avatars?

2. Is there a reason why the component.computerAvatarIds (now stored in component.computerAvatarSettings.ids after refactoring) is an object (e.g. {"robot":true, "girl":true}) instead of just an array of ids, like ["robot", "girl"]? It doesn't look like we ever set the "robot" field to false. Wouldn't it be simpler to use an array of enabled id strings?

@breity while refactoring, I touched a lot of files. I tried to be careful copying over your styles, but I might have messed up inadvertently. Can you please check that my work kept your styles in tact? Thanks!

geoffreykwan:
> 1. When no initial avatar is selected in the authoring tool, we still save it. Should we not allow this, make sure the author chooses initial avatars?

I made a change to handle this.

> 2. Is there a reason why the component.computerAvatarIds (now stored in component.computerAvatarSettings.ids after refactoring) is an object (e.g. {"robot":true, "girl":true}) instead of just an array of ids, like ["robot", "girl"]? It doesn't look like we ever set the "robot" field to false. Wouldn't it be simpler to use an array of enabled id strings?

Having the ids as keys in a map make it easier to check/uncheck avatars in the author view and also makes it easier to find which avatars to display to students in the student view.

PR #500, opened 2022-02-14
Use show work views in ShowWorkStudent

### LINKED ISSUE #498: Use ShowWork views in ShowWorkStudent
Modify ShowWorkStudent to display corresponding [ComponentType]ShowWork for each WISE component type. We currently show the disabled student views.

breity:
## Changes
- Display corresponding `<[component type]-show-work>` for each component type in ShowWorkStudent.
- Discussion and PeerChat still use student views because their show work views require teacher service and/or teacher permissions. (Will create separate issues.)
- Removed HTML, OutsideURL, and Summary component types, since we don't need show work views for those.

Closes #498.

## Test
- Make sure that ShowMyWork and ShowGroupWork components work for each component type (except Discussion and Peer Chat).
- Make sure that grading still works for each component type.


hirokiterashima (approved):
Looks good.

PR #502, opened 2022-02-15
fix(Show Group Work): Properly hide my work when authored to

### LINKED ISSUE #501: Show Group Work does not hide my work
When Show Group Work is authored to not "Show My Work", the student still sees a blank card with their name on it. The card with their name should not show up in this scenario.

geoffreykwan:
## Changes
Show Group Work, unchecking "Show My Work" now properly works.

## Test
Author a Show Group Work component and make sure the "Show My Work" checkbox properly shows and hides the student work in the student view.

Closes #501

### DIFF
--- a/src/assets/wise5/components/showGroupWork/show-group-work-student/show-group-work-student.component.html
+++ b/src/assets/wise5/components/showGroupWork/show-group-work-student/show-group-work-student.component.html
@@ -5,8 +5,7 @@
       [fxFlex.gt-sm]="widthMd"
       [fxFlex.gt-md]="widthLg"
       fxLayout="column">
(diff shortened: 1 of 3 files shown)


hirokiterashima:
Using ngIf to remove the cards without work resulted in empty spaces on the page (due to flex styling). We actually want to show the team's avatar and the message "Waiting for response" in this case.

If showMyWork is set to false, it should remove my workgroup from the ngFor loop consideration altogether.

@geoffreykwan can you test my changes? We should make sure that the following works:
- if isShowWork = false, it should remove my card, and don't leave any empty space on the page where my card would have been shown
- if my peer has not done work yet, it should show a card and say "Waiting for response"

PR #505, opened 2022-02-16
chore(VLE): Convert Edit Notebook Item dialog

### LINKED ISSUE #481: Convert EditNotebookItemController to Angular

geoffreykwan:
## Changes
- Convert Edit Notebook Item dialog in the student VLE
- Create Wise Link component and module
- Create Drag and Drop module

## Test
- Student can create a new Note
- Student can change the text of a Note
- Student can upload an image to a Note using file chooser or drag and drop
- Student can click on the Wise Link in the Note to bring them to the step the Note was created on
- Student can snip a Note from a graphical component like Concept Map, Draw, Graph, Label, etc.
- Student can delete an image from a Note
- Student can delete a Note
- Student can still upload an image to their Student Assets like when they are allowed to upload the background image to a Draw component

Closes #481


hirokiterashima (changes requested):
Functionality looks good.

It looks like these are no longer being used. Can we remove:
- EditNotebookItemController.js
- editNotebookItem.html
- EditNotebookItemDialogComponent.getItemNodePosition()

Also, I think making a WISELink class and passing it into WISELinkComponent (instead of passing in linkClass, linkText, nodeId, type) would make the code cleaner and easier to read. What do you think?

The styles changed in the add/edit note dialog since moving to Angular Material. @breity can you take a quick pass?

hirokiterashima:
Making a WIseLink class merits further discussion, so we'll keep it as is for now.

@breity can you look into any styling issues with this new component?

hirokiterashima (approved):

breity (changes requested):
I updated and moved styles to component and removed unused global styles. I added error text when note text is required but missing.

I updated wiselink links to use `<a>` instead of `<div>`. I also fixed the "Share with Class" checkbox. Sharing sometimes doesn't seem to work properly, but I think this an existing bug.

One issue: Changes to the note content (text and/or attachments) seem to persist even when clicking "Cancel".

breity (approved):
Looks good. 👍

PR #510, opened 2022-02-18
feat(Peer Chat): Send a notification when a Peer Chat message is sent

### LINKED ISSUE #509: Send notification when a Peer Chat message is made
- When a student sends a Peer Chat message, also send a notification to the Peer Chat members
- Make sure only one notification shows up at a time for the Peer Chat component even if multiple notifications are sent for that specific Peer Chat component

geoffreykwan:
## Changes
- When a Peer Chat message is sent, a notification is also sent to Peer Chat members

## Test
- Make sure a notification shows up when you are not on the Peer Chat step
- Make sure a notification does not show up when you are on the Peer Chat step
- Make sure only one notification shows up even if multiple messages are sent
- This scenario
  - Student is on a Peer Chat step
  - Student receives a notification for the Peer Chat step but we don't show it to them because they are on the same step as the notification
  - Refresh the VLE
  - The notification should still not show up

Closes #509


hirokiterashima:
In this scenario:
- There are three students (A,B,C) in a chat together
- A and B are on the chat step, C is on a different step
- A sends a message to chat
- B sends a message to chat

C only receives one message "A sent a chat message". This is not wrong, but it would be nice if the notification reflects the fact that B also sent a chat message.

In this case, should C see 
- one notification "A and B sent a chat message"? 
- two notifications "A sent a chat message", "B sent a chat message"
- keep as is?

hirokiterashima (changes requested):
When a teacher is a participant of a chat, they will receive a new notification. Can you change it so that clicking on it will take the teacher to the step where the chat is on? Right now it just dismisses the notification.

hirokiterashima:
I noticed these things on the teacher’s end:
- they always receive a notification from a chatroom they were in even if they’re on the chat step
- when they submit a new message to the chat, it does not create a notification for the students

These can be handled in a separate issue.

hirokiterashima (approved):

PR #513, opened 2022-02-24
refactor(DG Avatar): Use array to store avatar ids

hirokiterashima:
## Changes
- Use array instead of Map to store selectable avatars

## Rationale for change
- We were not making use of the Map's value (true/false). Avatars stored in componentContent contained unused values {"girl": true, "robot": true}, and all fields were set to true
- Array vs Map performance is negligible with N being small (6 items)

## Why not Set?
- Set does not serialize with JSON.stringify without extra work. 
   - e.g. JSON.stringify(new Set()) returns {}.
- Set does not parse with JSON.parse without extra work, and it's hard to know which fields should be instantiated as Set.

## Test
- Change componentContent.computerAvatarSettings.ids from object {"girl":true,"robot":true} to array ["girl", "robot"]
- Test that DG avatar authoring and DG avatar student selection work as before

### DIFF
--- a/src/assets/wise5/components/dialogGuidance/ComputerAvatarSettings.ts
+++ b/src/assets/wise5/components/dialogGuidance/ComputerAvatarSettings.ts
@@ -1,5 +1,5 @@
 export class ComputerAvatarSettings {
-  ids: {};
+  ids: string[];
   label: string;
   prompt: string;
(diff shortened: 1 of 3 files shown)


geoffreykwan (approved):
Looks good.

PR #514, opened 2022-02-24
refactor(DG Avatar): Use Set instead of Map to store avatars

hirokiterashima:


hirokiterashima:
We decided to use Array instead. See PR #513

PR #515, opened 2022-02-24
refactor(Dialog Guidance Avatar)

hirokiterashima:
## Changes
- Removed hightAvatarFunction: Selecting and highlighting avatar is managed by Angular Material, so
avatar.isSelected field is also not necessary
- In AT
   - initialize new component with all available avatars: eliminate need for null check
   - Remove numSelectedComputerAvatars: this number is never set to 0 anymore
- cleaned up and consolidated functions related to numSelectedComputerAvatars

## Test
- authoring available avatars and selecting avatar as student work as before
- In particular, try creating a new DG avatar component and make sure that all avatars are selected


geoffreykwan (approved):
Looks good.

PR #523, opened 2022-03-07
fix(Show Group Work): Handle when the student is not in a group yet

### LINKED ISSUE #522: Show Group Work handle scenario when not in group yet
Show a message to the student when they are not in a group yet.

geoffreykwan:
## Changes

Check if the student is not in a group and display a message.

## Test

- When the student is not in a group, it should show a message informing them they are not in a group.
- When the student is in a group, make sure it properly shows the group work.

Closes #522


hirokiterashima (approved):
Looks good.

PR #538, opened 2022-03-12
feat(Export): Implement Open Response custom export

### LINKED ISSUE #535: Open Response custom export with ideas
- Enable 'Export Component Data' for Open Response
- If an Open Response was scored and analyzed for ideas, show the ideas that were found
- Each idea id should have its own column. If the student had an idea in their response, we will display a 1 in that column. If they did not have that idea in their response, we will display a 0.

geoffreykwan:
## Changes

Added Open Response custom export

## Test

Generate student work for 3 Open Response components.
- One should be a regular Open Response.
- One should be an Open Response with a CRater item that returns a single score (like ```GREENROOF-II```).
- One should be an Open Response with a CRater item that returns ideas and multiple scores (like ```CarOnAColdDay_score-ki_idea-ki_nonscorable```)

Download the export for the 3 Open Responses and make sure they contain the correct data such as student response, ideas, and scores when applicable.
1. Open Classroom Monitor
2. Go to Data Export view
4. Find the Open Response and click "All" and "Latest"

Closes #535

### DIFF
--- a/src/assets/wise5/classroomMonitor/dataExport/dataExportController.ts
+++ b/src/assets/wise5/classroomMonitor/dataExport/dataExportController.ts
@@ -11,8 +11,9 @@ import { TeacherProjectService } from '../../services/teacherProjectService';
 
 class DataExportController {
   $translate: any;
-  allowedComponentTypesForAllRevisions = ['DialogGuidance', 'Discussion', 'Match'];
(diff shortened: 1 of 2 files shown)


hirokiterashima (approved):
Looks good. We should refactor dataExportController soon. It's getting big.

PR #543, opened 2022-03-15
perf(Teacher Peer Chat): Retrieve messages only when needed

### LINKED ISSUE #542: Teacher: only retrieve PeerChat messages when needed
### Description

Right now, the teacher will fetch the PeerGroup's chat messages when any new student work comes in for any step from any student. This results in needless computation and could lead to performance issues.

Change this so that the teacher only fetches new chat messages only when a member of the PeerGroup has submitted new work for the chat component that they are viewing.

hirokiterashima:
## Changes
- Ensure that newly-arrived StudentWork is for the PeerGroup that teacher is viewing
   - Behavior before this PR: the teacher will fetch the PeerGroup's chat messages when any new student work comes in for any step from any student. This results in needless computation and could lead to performance issues

- Change type of workgroupId from any to number

## Test
- Teacher PeerChat works as before: new messages are displayed as they come in
- Teacher only makes request for new messages when the new message is for the PeerGroup that they are viewing

## Note
Right now, a call to this.retrievePeerChatComponentStates() is needed, instead of just calling this.peerChatMessages.push(studentWork) in the code below:

```
if (this.isForThisPeerGroup(studentWork)) {
    this.retrievePeerChatComponentStates();
}
```

If we use the push method, new messages from another student in the PeerGroup (but not the same student that you're viewing in the grading tool) will not appear.

I think this could be looked at again in the future for better optimization, since retrievePeerChatComponentStates() makes a GET request. But for now, this PR would be a significant improvement over the current behavior.

Closes #542

### DIFF
--- a/src/assets/wise5/components/component-show-work.directive.ts
+++ b/src/assets/wise5/components/component-show-work.directive.ts
@@ -27,4 +27,8 @@ export abstract class ComponentShowWorkDirective {
   }
 
   ngOnDestroy(): void {}
+
+  isForThisComponent(object: any): boolean {
+    return object.nodeId === this.nodeId && object.componentId === this.componentId;
+  }
 }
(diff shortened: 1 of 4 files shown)


geoffreykwan (approved):
Looks good.

PR #551, opened 2022-03-17
chore(VLE): Convert Node Controller to Angular

### LINKED ISSUE #531: Convert nodeController.ts to Angular

geoffreykwan:
## Changes

Converted Node Controller to an Angular component.

## Test

- Make sure steps load components properly in the VLE
- Author rubrics for a step and a component and make sure they work properly when previewing the project

Closes #531


hirokiterashima (changes requested):
Basic functionality looks good. I had some questions/suggestions:

- Is NodeComponent.scrollAndHighlightComponent() still being used? Going to a component via WISELink doesn't seem to invoke the function. If not, we should remove it
- Rename ```helpIconClicked()``` to ```stopPulsing()```?
- Add private method signatures in NodeComponent
- Run ```gulp compile-sass``` to generate new .css and .css.map files that omit hopscotch

hirokiterashima:
It looks like saving dirty components breaks with this change?
1. Go to OR component and type in something
2. Without hitting save and before auto-save happens, go to the next step
3. Come back to the OR component

What you typed is no longer there. This works on production. I didn't test all component types, but changes don't save for these other components that I tested: MC, Draw, ConceptMap, Graph

hirokiterashima (approved):

PR #554, opened 2022-03-21
feat(notifications): Add clear notifications buttons

### LINKED ISSUE #548: Add button to clear all notifications
Add a button to allow both teachers and students to dismiss all notifications. This should dismiss all notifications except the ones where you need dismiss code to dismiss.

breity:
## Changes
- Adds clear all button to student notifications dialog
- Re-enables clear all button in teacher notifications dialog

Closes #548.

## Test
- Ensure clear all notifications buttons work for both student and teacher.
- Student clear all button should clear all notifications that don't require a dismiss code.
- Make sure clearing single notifications still works.

### DIFF
--- a/src/app/student/top-bar/top-bar.component.ts
+++ b/src/app/student/top-bar/top-bar.component.ts
@@ -126,7 +126,8 @@ export class TopBarComponent {
   viewAlerts($event: any): void {
     $event.stopPropagation();
     this.dialog.open(NotificationsDialogComponent, {
-      panelClass: 'dialog-sm'
+      panelClass: 'dialog-sm',
+      autoFocus: false
     });
   }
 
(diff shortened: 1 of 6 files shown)


hirokiterashima (changes requested):
Functionality looks good. Some questions/suggestions:
- Do we need to pass in event to dismissAll()?
- Can you add private method signatures to private functions?

breity:
> Functionality looks good. Some questions/suggestions:
> 
> * Do we need to pass in event to dismissAll()?

Looks like dismissNotification() uses the event, which gets added to arguments for broadcastViewCurrentAmbientNotification in NotificationService. But we don't actually use the event anywhere. (We only use `args.notification` when calling displayAmbientNotification.) I'll work on taking those events out.
> * Can you add private method signatures to private functions?

Sure, will update.

hirokiterashima (approved):

PR #565, opened 2022-04-14
chore(VLE): Convert VLE Controller to Angular

### LINKED ISSUE #559: Convert VLE Controller to Angular

geoffreykwan:
## Changes

- Converted VLEController to an Angular component
- Created Dialog components
  - Dialog With Close
  - Dialog With Confirm
  - Dialog Without Close

## Test

- Make sure the VLE works
 - Make sure steps load properly
 - Make sure navigating to other steps works
 - Make sure the project view works
 - Make sure the Notebook works
 - Make sure the Report works
- Make sure the popups work
  - Pause Screen
  - CRater "We are scoring your work"
  - Constraint when trying to go to a step that is locked
 - Make sure the VLE looks like it did before

Closes #559


geoffreykwan:
@breity Can you fix the styling in the VLE? To see the problems, it may help to create a step with a lot of components so that the page requires scrolling.

- Step content displays too far up when it should be displayed below the top bar and navigation
- Project view background color doesn't go all the way to the bottom when it should
- Open Notebook button shows at the very bottom of the screen (you need to scroll all the way down) when it should always hover at the bottom right of the view
- Report shows at the very bottom of the screen (you need to scroll all the way down) when it should always hover at the bottom right of the view
- Notebook notes view spans the whole height of the page when it should only span the height of the view

hirokiterashima:
Functionality that I tested so far seems to work.

I have some comments about the code:

### General
- Add private function signature where appropriate
- Remove empty scss files and references to them? No need to have a bunch of empty files, I think.

### vle.component.ts
- Move most of the subscriptions and initialization code from constructor to ngOnInit()? Even thought it can be in either, a quick search says it is common practice to put injection/field settings in the constructor and component initialization code in ngOnInit()
- Put each subscribe() call into private function calls and move all subscription code together. Right now they're spread out 
- Clean up RunStatus null checks. Are they all necessary? Put the code into its own function.
- Extract component for isEndedAndLocked message?
- Extract connection lost/reconnected into component?
- Remove empty lines (e.g. line 282, 285, 288, ...)

### Dialog
- Are the constructors in DialogWithConfirmComponent and DialogWithCloseComponent necessary?
- "DialogWithCloseComponent extends DialogWithoutCloseComponent" sounds a bit strange. Should DialogWithoutCloseComponent be named DialogComponent or SimpleDialogComponent?
- Type the dialog data, instead of ```any```, create new ```DialogContent``` interface that has title and content fields?
```
from
@Inject(MAT_DIALOG_DATA) public data: any,
to
@Inject(MAT_DIALOG_DATA) public dialogContent: DialogContent,
```

hirokiterashima (commented):
- Should DialogData be an interface instead of class, since we're only using it for type checking? [resouce 1](https://stackoverflow.com/questions/51716808/when-use-a-interface-or-class-in-typescript), [resource 2](https://medium.com/front-end-weekly/typescript-class-vs-interface-99c0ae1c2136)
- It looks like the WISEAPI block was removed. Should we also remove the block that fetches and evaluates the script file?
```
let script = this.projectService.getProjectScript();
if (script != null) {
  this.projectService.retrieveScript(script).then((script: string) => {
    new Function(script).call(this);
  });
}
```
- Are this.upgrade.$injector != null checks necessary, or safe to remove?
```
if (this.upgrade.$injector != null && this.upgrade.$injector.get('$state') != null) {
...
if (this.upgrade.$injector != null) {
```
- Remove comments in subscribeToNodeClickLocked()
- Can we rewrite shouldScreenBePaused() using some():
```
(untested)
private shouldScreenBePaused(): boolean {
  const runStatus = this.studentDataService.getRunStatus();
  const periodId = this.configService.getPeriodId();
  return runStatus.periods.some((period) => period.periodId === periodId && period.paused);
}
```

breity (approved):
Looks good.

I updated styles and fixed layout issues. I also fixed some bugs and updated simple dialog styles to match the site dialog style conventions (and added a 'scroll' boolean field to DialogData).

hirokiterashima (approved):

PR #568, opened 2022-04-27
build(Fabric): Use v3.6.3

### LINKED ISSUE #567: Consolidate Fabric version
### Description

Label uses Fabric 3.6.6 and and Draw uses Fabric 3.6.3. Consolidate.

hirokiterashima:
## Changes
Label and Draw were using different versions of Fabric. Since Draw is an external library that uses 3.6.3, we opt to fix Fabric version to 3.6.3.

## Before Test
- run ```npm install``` to get latest deps

## Test
- Label and Draw components work as before. 
- Open dev tools and make sure that there is only one Fabric that is loaded, and it is v3.6.3.

Closes #567

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -44,7 +44,7 @@
         "dom-autoscroller": "^2.3.4",
         "drawing-tool": "^2.1.2",
         "eventemitter2": "^5.0.1",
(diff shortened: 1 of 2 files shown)


geoffreykwan (approved):
Looks good.

PR #577, opened 2022-05-05
fix(VLE): disable close for open response c-rater dialog

### LINKED ISSUE #576: Disable close for Open Response c-rater scoring dialog
## Description
Students shouldn't be able to dismiss (with esc key or mouse click on backdrop) the dialog informing them that WISE is scoring their work. The dialog closes on its own once the scoring has completed or times out.

breity:
## Test
Make sure user can't close (with esc key or clicking outside the dialog) the dialog that appears when c-rater is scoring a response to an Open Response item.

Closes #576.


hirokiterashima (approved):
LGTM

PR #581, opened 2022-05-06
feat(Dialog Guidance): Show different feedback for same FeedbackRule

### LINKED ISSUE #578: feat(Dialog Guidance): Show different feedback for same FeedbackRule
### Description

Request from researchers: When the student matches the same FeedbackRule, cycle through an array of feedbacks that should be shown to the student, instead of always showing the same feedback text.

Changes required:
- AT: allow authors to specify one or more feedback text for a FeedbackRule
- Student VLE: remember which FeedbackRule.feedback was show to the student before. Show the next one in the array. Use mod (%) to calculate which feedback to show. 
   - ```FeedbackRule.feedbacks[numTimesStudentsSawThisFeedback % FeedbackRule.feedbacks.length]```
- Make this backwards compatible; existing DG items should work as is

hirokiterashima:
## Changes
Add check for version in content. Version 2 assumes that each feedback rule has an id, and uses the id to determine if student hit the same FeedbackRule before, and which feedback within that rule to show next (use mod).

## Before Test
- create new DG component with V2 content structure
   - add "version": 2
   - set FeedbackRule.feedback to array

## Test
- Existing DG component works
- V2 DG component works

Closes #578

### DIFF
--- a/src/app/services/dialogGuidanceFeedbackService.spec.ts
+++ b/src/app/services/dialogGuidanceFeedbackService.spec.ts
@@ -1,0 +1,76 @@
+import { TestBed } from '@angular/core/testing';
+import { DialogGuidanceStudentComponent } from '../../assets/wise5/components/dialogGuidance/dialog-guidance-student/dialog-guidance-student.component';
+import { FeedbackRule } from '../../assets/wise5/components/dialogGuidance/FeedbackRule';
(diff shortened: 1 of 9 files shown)


breity (approved):
Looks good. 👍

geoffreykwan (approved):
Looks good.

PR #586, opened 2022-05-11
refactor(Component Service): Remove StudentDataService dependency

### LINKED ISSUE #585: refactor(Component Service): Remove StudentDataService dependency
Remove the StudentDataService dependency from the parent ComponentService. It is not being used by all components. Allow each child component service to add it as dependency as needed.

hirokiterashima:
## Changes
- Removed StudentDataService from parent ComponentService because it was not being used by all components. Some components use it, so they inject it themselves

## Test
- VLE, AT, and CM work as before

Closes #585


geoffreykwan (approved):
Looks good.

PR #595, opened 2022-05-16
refactor(ORService): Remove StudentDataService dependency

### LINKED ISSUE #587: refactor(ORService): Remove StudentDataService dependency
### Description

Refactor OpenResponseService to remove the dependency to StudentDataService. It's only being used to calculate completion (uses StudentDataService.studentData). Send the studentData as param to the refactored class/function instead.

hirokiterashima:
## Changes
- Extracted code related to calculating CompletionCriteria (which was only being used in OR) into OpenResponseCompletionCriteriaService (ex: isInOrderCompletionCriteriaSatisfied(), getComponentStateSubmittedAfter(), etc.), and added new function ORService.isCompletedV2() that calls OpenResponseCompletionCriteriaService.isSatisfied()
- Changed StudentDataService.isComponentCompleted() to call ORService.isCompletedV2(). isCompletedV2 is temporary, and will be refactored/replaced in the future when other component services no longer require StudentDataService.

## Test
- OR completion constraints and calculations work as before

Closes #587


geoffreykwan:
If this is no longer used should we just delete it?

https://github.com/WISE-Community/WISE-Client/blob/c9ef6fbd6d7828f54d5f40388c615ae7177a37b5/src/assets/wise5/components/openResponse/openResponseService.ts#L29-L31

geoffreykwan (approved):
Looks good.

PR #602, opened 2022-05-19
refactor(ConceptMapService): Remove NodeService dependency

### LINKED ISSUE #601: Remove NodeService dependency from ConceptMapService
### Description

This will cause the following circular dependency when we create ComponentServiceLookupService.
```CMService -> NodeService -> StudentDataService -> ComponentServiceLookupService -> CMService.```

We only use NodeService to create a component state in the cm-student-component (component -> CMService -> NodeService). This should be done this way instead: cm-student-component -> NodeService.

hirokiterashima:
## Changes
- Before: use NodeService to create a component state in the cm-student-component (component -> CMService -> NodeService). 
- After: call NodeService directly from cm-student-component (cm-student-component -> NodeService)

This will remove the circular dependency when we create ComponentServiceLookupService. Right now with the NodeService dependency, it will look like this: 
CMService -> NodeService -> StudentDataService -> ComponentServiceLookupService -> CMService.

## Test
- ConceptMap connected component works as before (the code that calls ```createMergedComponentState()```). For example, import Draw -> ConceptMap. 

Closes #601

### DIFF
--- a/src/app/services/conceptMapService.spec.ts
+++ b/src/app/services/conceptMapService.spec.ts
@@ -9,7 +9,6 @@ import { UtilService } from '../../assets/wise5/services/utilService';
 import { ConceptMapService } from '../../assets/wise5/components/conceptMap/conceptMapService';
 import { ConfigService } from '../../assets/wise5/services/configService';
 import { SessionService } from '../../assets/wise5/services/sessionService';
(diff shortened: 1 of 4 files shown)


geoffreykwan (approved):
Looks good.

Issue #592, opened 2022-05-13
Make Summary retrieve classmate work from Summary API endpoint


geoffreykwan:
### Description

When student data is retrieved from classmates in the summary-display component, use the Summary specific API endpoint instead of the general student data API endpoint.

hirokiterashima:
closed by #603

PR #603, opened 2022-05-20
refactor(Summary): Use summary endpoint

geoffreykwan:
You should test this with https://github.com/WISE-Community/WISE-API/pull/131

## Changes

- Retrieve summary data from summary endpoint instead of generic student data endpoint
- Fixed properly displaying data from period or all periods in classroom monitor based upon which period is currently selected
- Fixed the graph title based upon self, period, or class

# Test

- Summary works for VLE preview
- Summary works for Authoring preview
- In student VLE, Summary properly shows data depending on whether it is authored to show data from self, period, or all periods
- Summary works for Classroom Monitor (when you click Node Info on a step that contains a Multiple Choice component or a component with any scores)
- In Classroom Monitor, Summary properly shows data depending on which period is selected or all periods are selected
- The summary graph title properly displays the text "Your", "Period", or "Class"

#592

### DIFF
--- a/src/assets/wise5/directives/summaryDisplay/summary-display.component.ts
+++ b/src/assets/wise5/directives/summaryDisplay/summary-display.component.ts
@@ -1,12 +1,15 @@
 import * as Highcharts from 'highcharts';
 import { Component, Input, SimpleChanges } from '@angular/core';
 import { UpgradeModule } from '@angular/upgrade/static';
-import { Subscription } from 'rxjs';
+import { Observable, Subscription } from 'rxjs';
 import { AnnotationService } from '../../services/annotationService';
 import { ConfigService } from '../../services/configService';
 import { ProjectService } from '../../services/projectService';
 import { StudentDataService } from '../../services/studentDataService';
 import { UtilService } from '../../services/utilService';
+import { SummaryService } from '../../components/summary/summaryService';
+import { from, of } from 'rxjs';
+import { tap } from 'rxjs/operators';
 
 @Component({
   selector: 'summary-display',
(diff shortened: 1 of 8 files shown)


src/assets/wise5/directives/summaryDisplay/summary-display.component.ts
```diff
@@ -718,23 +727,53 @@ export class SummaryDisplay {
...
+      return this.getGraphTitleForPeriod();
     } else {
       return this.getGraphTitleForClass();
     }
   }
 
   getGraphTitleForSelf() {
-    return 'Your Response';
+    if (this.isStudentDataTypeResponses()) {
+      return 'Your Response';
```
hirokiterashima:
```suggestion
      return $localize`Your Response`;
```

src/assets/wise5/directives/summaryDisplay/summary-display.component.ts
```diff
@@ -718,23 +727,53 @@ export class SummaryDisplay {
...
+    } else if (this.isStudentDataTypeScores()) {
+      return 'Your Score';
```
hirokiterashima:
```suggestion
      return $localize`Your Score`;
```

hirokiterashima (changes requested):
I noticed few things while testing.

Functionality:
- In the CM, for the summary component, I always see summary graph for all periods even if I change the period selected
Screen Shot 2022-05-23 at 2 15 51 PM


Code:
- Remove StudentDataService.getClassmateStudentWork() and corresponding tests since they're no longer used?
- See inline code change suggestions
- Generate messages

hirokiterashima (approved):

PR #605, opened 2022-05-20
fix(VLE): Make components full width on small screens

### LINKED ISSUE #604: Components should be full width on small screens
## Description
Sometimes components are displayed side-by-side on smaller screens. This includes embedded components and components with narrow content. Components should always be full width on small screens.

breity:
## Changes
Ensures that components take up the full page width on smaller screens.

Closes #604.

## Test
- Make sure all component types take up the full width of the available page space when browser window is less than 900px wide.
- Make sure to test with embedded components.


geoffreykwan (approved):
Looks good.

PR #607, opened 2022-05-23
refactor(ComponentServiceLookupService): Replace $injector.get(componentService)

### LINKED ISSUE #606: refactor(ComponentServiceLookupService): Use instead of $injector.get(componentService)
### Description

Move away from using $injector.get(componentType + 'Service') by creating a new class ComponentServiceLookupService. This will help us move away from AngularJS.

hirokiterashima:
## Changes
- Create ComponentServiceLookupService
- Replace $injector.get(componentType + 'Service') with ComponentServiceLookupService.getService(componentType)
- Remove dependencies to $injector where possible
- Note: not all instances of $injector.get(componentService) were replaceable at this time (e.g. StudentDataService, UtilService) because it would have created circular dependencies. Those classes will first need to be refactored.

## Test
- MC Summary in CM via the "Step Tools" dialog: should call NodeInfo.componentHasCorrectAnswer() which calls ComponentServiceLookupService.getService()
- Node Authoring: deleting a component should call hideAllComponentSaveButtons() which calls ComponentServiceLookupService.getService()
- Data export should work as before and use ComponentServiceLookupService to look up component services 

Closes #606


geoffreykwan (approved):
Looks good.

PR #620, opened 2022-05-28
refactor(Component Type Label): move function into new ComponentTypeService

### LINKED ISSUE #619: Move ComponentTypeLabel lookup into new service class
### Description

Right now the component type and label lookup are in UtilService. They should be in a new class to reduce UtilService's complexity and increase its cohesion. Also makes will make it easier to upgrade to Angular by reducing circular dependency.

hirokiterashima:
## Changes
- Create ComponentTypeService and moved getComponentTypes() and getComponentTypeLabel() from UtilService
- Use ComponentServiceLookupService to look up service instead of upgrade.$injector.get(ComponentService)
- Update existing calls 
   - UtilService.getComponentTypes() -> ComponentTypeService.getComponentTypes()
   - UtilService.getComponentTypeLabel() -> ComponentTypeService.getComponentTypeLabel()

## Test

Component labels appear in the following pages:
- Authoring
   - Add new step
   - Add new component
   - Choose component location
- CM
   - Milestone
   - Step Info

Closes #619


geoffreykwan (approved):
Looks good.

PR #656, opened 2022-07-05
feat(Session): Show login page immediate on session expire

### LINKED ISSUE #655: Show log in page immediately when session times out
### Description

Instead of showing a toast message asking user to refresh the page so they can sign in again, show the sign in page with a message about what happened.

hirokiterashima:
## Changes
- When session times out while using VLE, AT, CM, show message indicating an error and prompt user to login again. Upon logging in, they will be redirected to where they where before

## Test
- While using VLE, AT, CM, invalidate the session via Redis-cli ```flushall``` command. Then, try saving work by making changes or going to the next step.
   - They should be redirected to the log in page and see an error message, prompting them to log back in and check recent work
   - When they log in, they should be redirected to where they were before in the app 

Closes #655

### DIFF
--- a/src/app/http-error.interceptor.ts
+++ b/src/app/http-error.interceptor.ts
@@ -31,6 +31,7 @@ export class HttpErrorInterceptor implements HttpInterceptor {
 
         if (err.status >= 500 || err.status === 0 || err.status === 403) {
           this.processError(err);
+          return;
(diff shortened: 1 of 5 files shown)


geoffreykwan (approved):
Looks good.

PR #664, opened 2022-07-13
fix(Teacher): Do not fetch units in homepage

### LINKED ISSUE #663: Don't fetch units in teacher run listing page
### Description

Right now we fetch personal, shared, library and community units when teacher logs in and sees the homepage (```/teacher/home/schedule``` route) and don't do anything with them. The units are only needed in the ```/teacher/home/library/*``` routes.

Remove these unnecessary API hits

hirokiterashima:
## Changes
- The units (personal, shared, community, library) are only fetched when the teacher accesses the ```/teacher/home/library/*``` routes. Before, we requested it in the run listing page (```/teacher/home/schedule``` route) even though we didn't use it
   - I used [mat-tab-nav-panel](https://material.angular.io/components/tabs/overview#tabs-and-navigation) to make this work 
- Cleaned up code
   - Moved component initialization in TeacherProjectLibraryComponent from constructor to ngOnInit
   - Removed unused code

## Test
- ```/api/project/[personal,shared,community/library]``` routes are *not* requested when teacher loads the home page (run listing page)
- The above routes *are* requested when teacher goes to the "Unit Library" tab
- Both run listing and unit library pages work as before

Closes #663

### DIFF
--- a/src/app/modules/library/community-library/community-library.component.ts
+++ b/src/app/modules/library/community-library/community-library.component.ts
@@ -13,15 +13,19 @@ export class CommunityLibraryComponent extends LibraryComponent {
   projects: LibraryProject[] = [];
   filteredProjects: LibraryProject[] = [];
 
-  constructor(libraryService: LibraryService, public dialog: MatDialog) {
+  constructor(protected libraryService: LibraryService, public dialog: MatDialog) {
     super(libraryService);
-    libraryService.communityLibraryProjectsSource$.subscribe((communityProjects) => {
-      this.projects = communityProjects;
-      this.filterUpdated();
-    });
   }
 
-  ngOnInit() {}
+  ngOnInit() {
+    super.ngOnInit();
+    this.subscriptions.add(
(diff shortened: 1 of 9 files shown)


breity (changes requested):
Looks good, as units are now only fetched when the library tab is loaded.

I found a bug when switching between tabs. See if you can reproduce:
- Load the teacher home and switch to the Unit Library tab.
- Switch back to the Class Schedule tab.
- Then go to the Unit Library tab again.
- Try to view the Community Built and My Units sub-tabs. The units don't show up in the DOM.

hirokiterashima:
Good catch. It should work now.

The [Personal,Community,Official]Library components are destroyed when teacher moves to /schedule route, so we need to fetch the units again. 

I also refactored the [Personal,Community,Official]LibraryComponent code by moving the subscriptions from the constructor to ngOnInit().

breity (approved):
Looks good. 👍

PR #668, opened 2022-07-14
refactor(TeacherRunListItemComponent): Reduce unnecessary computation

### LINKED ISSUE #667: Refactor: TeacherRunListItemComponent
### Description

We compute gradeAndManageLink for each RunListItem in ngOnInit() even though it may never be used. Move this to a function that is invoked only when necessary.

Also, the period strings tooltip ("Class Periods: ...") is calculated in every digest loop. This should only calculated once at the beginning, and when there are changes to the RunListItem's period by adding/deleting periods.

hirokiterashima:
## Changes
- Compute gradeAndManageLink only when CM is requested (before: always on component load)
- Compute period strings tooltip ("Class Periods: ...") only on component load and when the run's periods change (before: this was computed in every digest loop)
- Use TeacherService.runs$ observable to send new and updated Runs
- Clean up code

## Test
- You can launch CM via "Teacher Tools" button
- "Class Periods: ..." tooltip displays correctly for each run item
- Class periods tooltip and period filter update correctly when periods are added/deleted from a run
- Search and filter runs work as before

Closes #667

### DIFF
--- a/src/app/teacher/create-run-dialog/create-run-dialog.component.spec.ts
+++ b/src/app/teacher/create-run-dialog/create-run-dialog.component.spec.ts
@@ -29,7 +29,7 @@ export class MockTeacherService {
     });
   }
 
-  addNewRun() {}
+  broadcastRunChanges() {}
 
   setTabIndex() {}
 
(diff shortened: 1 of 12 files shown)


breity (approved):
Looks good. 👍

PR #672, opened 2022-07-15
fix(UtilService): Handle undefined param in makeCopyOfJSONObject()

### LINKED ISSUE #671: Bug: can't save table data
### Description

Type in a value in any cell and try to save. You'll see this js error:
```
core.mjs:6494 ERROR SyntaxError: Unexpected token u in JSON at position 0
    at JSON.parse (<anonymous>)
    at T.makeCopyOfJSONObject (utilService.ts:66:17)
    at rt.createComponentState (table-student.component.ts:408:55)
    at rt.createComponentStateAndBroadcast (component-student.component.ts:560:10)
    at rt.studentDataChanged (table-student.component.ts:1003:10)
    at table-student.component.html:46:32
```

Reported by Cristian

hirokiterashima:
## Changes
- UtilService.makeCopyOfJSONObject(undefined) returns undefined

## Test
- Type in a value in any table cell and save. This should work now. Before it was throwing a js error because UtilService.makeCopyOfJSONObject(undefined) was not properly handled.

Closes #671

### DIFF
--- a/src/app/services/utilService.spec.ts
+++ b/src/app/services/utilService.spec.ts
@@ -85,6 +85,10 @@ function makeCopyOfJSONObjectTests() {
     it('should return null for null input', () => {
       expect(service.makeCopyOfJSONObject(null)).toEqual(null);
     });
+
+    it('should return undefined for undefined input', () => {
+      expect(service.makeCopyOfJSONObject(undefined)).toEqual(undefined);
+    });
   });
 }
 
--- a/src/assets/wise5/services/utilService.ts
+++ b/src/assets/wise5/services/utilService.ts
@@ -62,8 +62,12 @@ export class UtilService {
     return str;
   }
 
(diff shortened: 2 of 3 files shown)


breity (changes requested):
Great. Saving student input on editable cells works now for cells that don't have authored starting values.

I found that editable cells that also have an authored value don't seem to repopulate student input values when the component is re-loaded. To reproduce:
1. Author a table component with a cell that is marked as editable.
2. Add a value in the authoring view to the editable cell.
3. In preview or in a run, edit the cell value and save.
4. Move to another step and come back to the table step. You should see the original authored value in the cell and not the edited student value.

breity (approved):
On further inspection, I can only reproduce with one specific component. Others work fine, so must be something specific to that component.

Looks good.

PR #677, opened 2022-07-19
feat(Data Explorer): Only allow an option to be selected once

### LINKED ISSUE #386: Data Explorer: Only allow table column to be selected once
### Description

Only allow students to select each column once for graph display in Data Explorer activities. As of now, they can select the same column for display multiple times on the graph, which is not productive and can be confusing.

**Expected behavior:**

When a student choose a table column for display as either the x-axis or a y-axis graph series, that column should not be selectable again.

**Actual behavior:** [What actually happens]

Students can choose a column from the table for multiple graph series.

geoffreykwan:
## Changes

Students can only select an option once.

## Test

Make sure once a student selects an option that the option does not show up in the other drop downs.

Closes #386

### DIFF
--- a/src/assets/wise5/components/table/table-student/table-student.component.ts
+++ b/src/assets/wise5/components/table/table-student/table-student.component.ts
@@ -1024,7 +1024,7 @@ export class TableStudent extends ComponentStudent {
     return (
       columnIndex === this.dataExplorerXColumn ||
       this.dataExplorerSeries.some((series) => {
-        return columnIndex === series.yColumn;
+        return series.yColumn === columnIndex;
       })
     );
   }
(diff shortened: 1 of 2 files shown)


src/assets/wise5/components/table/table-student/table-student.component.ts
```diff
@@ -1010,6 +1013,22 @@ export class TableStudent extends ComponentStudent {
...
+      this.columnIndexToIsUsed.set(c, this.isColumnUsed(c));
+    }
+  }
+
+  isColumnUsed(columnIndex: number): boolean {
+    return (
+      columnIndex === this.dataExplorerXColumn ||
+      this.dataExplorerSeries.some((series) => {
+        return columnIndex === series.yColumn;
+      })
```
hirokiterashima: In this block, should the operands be switched, since series.yColumn changes and columnIndex is fixed?
```suggestion
      this.dataExplorerSeries.some((series) => {
        return series.yColumn === columnIndex;
      })
```

hirokiterashima (commented):
Functionality works. I had a question about code, see inline comments.

hirokiterashima (approved):

PR #685, opened 2022-07-22
feat(Table): Allow authors to import table

### LINKED ISSUE #684: feat(Table): Allow authors to import table
### Description

Allow authors to import a csv file to populate the table in a Table component.

geoffreykwan:
## Changes

Added the ability for authors to import a table in the Table component authoring.

## Test

2. Go to a Table step
3. Go to the Advanced view for the Table component authoring
4. Click the "Import Table" button
5. Choose a csv file from your computer (that has 500 or less cells)
6. Make sure the Table component now contains the data from your csv file
7. Try importing a table with more than 500 cells, it should not allow you to do so

Closes #684

### DIFF
--- a/src/assets/wise5/components/table/edit-table-advanced/edit-table-advanced.component.ts
+++ b/src/assets/wise5/components/table/edit-table-advanced/edit-table-advanced.component.ts
@@ -11,11 +11,12 @@ import { TeacherProjectService } from '../../../services/teacherProjectService';
 })
 export class EditTableAdvancedComponent extends EditAdvancedComponentComponent {
   allowedConnectedComponentTypes = ['Embedded', 'Graph', 'Table'];
+  columnNames: string[] = [];
   isDataExplorerScatterPlotEnabled: boolean;
   isDataExplorerLineGraphEnabled: boolean;
   isDataExplorerBarGraphEnabled: boolean;
   numColumns: number;
-  columnNames: string[] = [];
+  importTableMessage: string;
 
   constructor(
     protected NodeService: NodeService,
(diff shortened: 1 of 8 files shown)


src/assets/wise5/components/table/edit-table-advanced/edit-table-advanced.component.ts
```diff
@@ -177,4 +178,130 @@ export class EditTableAdvancedComponent extends EditAdvancedComponentComponent {
...
+        this.importTableMessage = $localize`Successfully imported table`;
+      };
+      reader.readAsText(files[0]);
+    } else {
+      event.target.value = null;
+    }
+  }
+
+  importTable(fileContent: string): void {
+    const twoDimensionalStringArray = this.csvToArray(fileContent);
```
hirokiterashima: Can we use ```UtilService.CSVToArray(strData, strDelimiter)``` here instead?

hirokiterashima (changes requested):
Functionality works for small file size I tried (2KB, 12 rows, 20 cols). It increased size of project.json from 9KB to 20KB though, which may become an issue if the unit uses a lot of large table data sets. When I tried uploading 96KB file (108 rows, 24 cols), it froze the AT and the VLE preview didn't load, and the browser eventually gave me a "do you want to kill the tabs?" alert message.

Some questions:
1. Should we limit file size and/or number of rows/columns?
2. Should we store the data in a separate file (outside of project.json) and add a link to it instead? This will reduce project.json size and only fetch the file when users need it.
3. Should we show a "importing..." message?

See inline comment.

src/assets/wise5/components/table/edit-table-advanced/edit-table-advanced.component.ts
```diff
@@ -177,4 +183,91 @@ export class EditTableAdvancedComponent extends EditAdvancedComponentComponent {
       }
     }
   }
+
+  importTableFile(event: any): void {
+    if (confirm($localize`Are you sure you want to overwrite the existing table?`)) {
+      this.showImportingTableDisplay();
+      this.setImportTableMessage($localize`Importing table...`);
+      const files = event.target.files;
```
hirokiterashima: Move this files variable declaration to right above where we need it, or don't declare at all.
```
const files = event.target.files;
reader.readAsText(files[0]);
```
or
```
reader.readAsText(event.target.files[0]);
```

src/assets/wise5/components/table/edit-table-advanced/edit-table-advanced.component.ts
```diff
@@ -177,4 +183,91 @@ export class EditTableAdvancedComponent extends EditAdvancedComponentComponent {
...
+      const reader: any = new FileReader();
+      reader.onload = () => {
+        const fileContent = reader.result;
```
hirokiterashima:
```suggestion
      const reader: FileReader = new FileReader();
      reader.onload = () => {
        const fileContent = reader.result as string;
```

src/assets/wise5/components/table/edit-table-advanced/edit-table-advanced.component.ts
```diff
@@ -177,4 +183,91 @@ export class EditTableAdvancedComponent extends EditAdvancedComponentComponent {
...
+        } else {
+          this.importTable(tableContent);
+          this.setImportTableMessage($localize`Successfully imported table`);
+        }
+        event.target.value = null;
+        this.hideImportingTableDisplay();
+      };
+      reader.readAsText(files[0]);
+    } else {
+      event.target.value = null;
```
hirokiterashima: This line is in both the if-true and if-else cases: ```event.target.value = null;``` Can we move it outside and remove the else case?

src/assets/wise5/components/table/edit-table-advanced/edit-table-advanced.component.ts
```diff
@@ -177,4 +183,91 @@ export class EditTableAdvancedComponent extends EditAdvancedComponentComponent {
...
+    }
+  }
+
+  getNumCells(tableContent: string[][]): number {
+    let numCells = 0;
+    for (const row of tableContent) {
+      for (const cell of row) {
+        numCells++;
+      }
```
hirokiterashima:
```suggestion
      numCells += row.length;
```

src/assets/wise5/components/table/edit-table-advanced/edit-table-advanced.component.ts
```diff
@@ -177,4 +183,91 @@ export class EditTableAdvancedComponent extends EditAdvancedComponentComponent {
...
+        tableRow.push({
+          text: cell,
+          editable: true,
+          size: null
+        });
+      }
+      table.push(tableRow);
+    }
+    return table;
+  }
```
hirokiterashima:
```suggestion
  convertToTableData(stringArray: string[][]): any[][] {
    return stringArray.map((row) =>
      row.map((cell) => {
        return { text: cell, editable: true, size: null };
      })
    );
  }
```

hirokiterashima (changes requested):
New functionality looks good. See additional inline comments.

src/assets/wise5/components/table/edit-table-advanced/edit-table-advanced.component.ts
geoffreykwan: I left this as a double for loop since it will be more efficient when there are more cells. The double for loop is also less complex according to CodeMetrics.

hirokiterashima (approved):
Re-writing the map this way
```
    return stringArray.map((row) =>
      row.map((cell) => ({ text: cell, editable: true, size: null }))
    );
```
or this way (requires defining a new Cell class):
```
    return stringArray.map((row) =>
      row.map((cell) => new Cell(cell, true, null))
    );
```
seems to result in the same complexity as the for-loop (5 in the first way, 4 in the second if you also set function return type to Cell). We should define a Cell class in any case, in a future refactoring task.

```for...of``` loop does seem to be faster for large data sets than ```map``` ([reference](https://leanylabs.com/blog/js-forEach-map-reduce-vs-for-for_of/)), but since we are limiting the number of cells to 500 in this function, I wonder if the gains would be that much noticeable. I personally find it easier to read code that is more declarative and less verbose (5 lines with map vs 11 with for...of). But this is a personal opinion and we should discuss and agree on a convention as a team.

Approving for now, but let's continue to discuss and weigh the pros and cons of each.

PR #687, opened 2022-07-26
test(BranchService): Test getBranches()

### LINKED ISSUE #686: Write tests for BranchService

hirokiterashima:
## Note
- I had to set ```paths``` each time before a test inside beforeEach() instead of setting it once at the top of the file because ```paths``` gets mutated during calls to getBranches() and results in unexpected results.

## Test
- Run the tests and make sure that they pass

## Questions
- Are there other tests that we should write for getBranches()?
- Are all the null checks in the private functions (e.g. ```if (paths != null) {```, ```if (path1 != null && path2 != null)```) necessary? Will we ever call getBranches() where those values will be null?

Closes #686

### DIFF
--- a/src/app/services/branchService.spec.ts
+++ b/src/app/services/branchService.spec.ts
@@ -9,8 +9,24 @@ let service: BranchService;
 let configService: ConfigService;
 
 let paths: string[][];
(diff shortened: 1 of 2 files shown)


geoffreykwan (approved):
Looks good.

PR #689, opened 2022-07-27
refactor(ProjectService): Move functions to [VLE/Teacher]ProjectService

### LINKED ISSUE #688: Move functions from ProjectService to VLE/TeacherProjectService
### Description

In ProjectService, functions like ```retrieveProjectById()```, ```getProjectRubric()```,```getApplicationNodes()```, etc are only used by the CM or the AT, and should be moved to TeacherProjectService so that they are not loaded for the students.

```getConnectedComponentsByNodeIdAndComponentId()``` should be moved to VLEProjectService since it's only used by the students.

Also add ```private``` and parameter types as necessary.

hirokiterashima:
## Changes
- Move [student/teacher] specific functions from ProjectService to [VLE/Teacher]ProjectService
- Add private and parameter types to functions
- Remove unused functions

## Test
- VLE, CM and AT work as before
- In particular
   - constraints: authoring in AT and viewing in VLE

Closes #688

### DIFF
--- a/src/app/services/projectService.spec.ts
+++ b/src/app/services/projectService.spec.ts
@@ -84,7 +84,6 @@ describe('ProjectService', () => {
   // TODO: add test for service.removeDuplicatePaths()
   // TODO: add test for service.pathsEqual()
   // TODO: add test for service.getNodeContentByNodeId()
-  // TODO: add test for service.replaceComponent()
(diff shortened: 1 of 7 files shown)


geoffreykwan (approved):
Looks good.

PR #704, opened 2022-08-03
feat(Graph): Replace empty table values with NA string

### LINKED ISSUE #703: feat(Graph): Replace empty table values with 'NA' string
### Description

When table data is sent to a graph, replace all empty values with the string 'NA'.

geoffreykwan:
## Changes

- When a table is sent to a graph, have the empty values replaced with the 'NA' string
- Fixed a bug that would occur if the x value of a table was equal to the length of the table

## Test

- Make sure sending a table to a graph still works if there are empty cells in the x column
- Make sure sending a table to a graph still works if there are empty cells in the y column
- Make sure sending a table to a graph still works if there are empty cells in the x and y column for a given row
- Create a Data Explorer and set one of the x values equal to the number of data rows for example like this below. There are 3 data rows and a row has 3 for the x column (Cats). Previously this would throw an error and the graph would not render. It should now work without any errors.

Screen Shot 2022-08-02 at 5 07 41 PM

Closes #703

### DIFF
--- a/src/assets/wise5/components/graph/graph-student/graph-student.component.ts
+++ b/src/assets/wise5/components/graph/graph-student/graph-student.component.ts
@@ -1710,11 +1710,11 @@ export class GraphStudent extends ComponentStudent {
   addPointFromTableIntoData(xCell: any, yCell: any, data: any[]) {
     let xText = xCell.text;
     if (xText == null || xText === '') {
-      xText = 'NA';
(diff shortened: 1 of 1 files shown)


breity (approved):
Looks good.

I changed the value for empty cells from 'NA' to 'N/A', which makes it a little clearer that the value means "Not available".

PR #708, opened 2022-08-04
refactor(ProjectService): Extract injectClickToSnipImage to new service

### LINKED ISSUE #707: Extract ProjectService.injectClickToSnipImage to new class
This class has nothing to do with the rest of the ProjectService class.

hirokiterashima:
## Changes
- Extract injectClickToSnipImage to new InjectClickToSnipImageService and update references
- Move ProjectService field initialization from constructor to field declaration

## Test
- VLE, AT, CM parses project and loads as before
- For units with notebook & snip to notebook options enabled, an image in the HTML component can be snipped

Closes #707

### DIFF
--- a/src/app/services/clickToSnipImageService.spec.ts
+++ b/src/app/services/clickToSnipImageService.spec.ts
@@ -1,0 +1,33 @@
+import { TestBed } from '@angular/core/testing';
+import { ClickToSnipImageService } from '../../assets/wise5/services/clickToSnipImageService';
+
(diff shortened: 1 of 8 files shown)


geoffreykwan (approved):
Looks good.

I noticed that when a student clicks on an image to snip it, it opens the create note popup twice. That means you need to close it twice to get back to the VLE. This is an existing issue on production so we can fix this in another issue.

PR #710, opened 2022-08-04
fix(Preview): Getting logged out after inactivity

### LINKED ISSUE #709: Getting logged out while previewing unit
### Description

Without logging in, open unit in preview mode. After a while of inactivity, you get logged out and sent back to the homepage. You should never be logged out when you're in preview.

hirokiterashima:
## Changes
- Don't initialize session checks if the VLE is launched in preview mode
- Code cleanup
   - Lowercase ConfigService param in SessionService
   - Add private and function param and return types in SessionService

## Test
- You DO NOT get logged out while you're previewing unit
- You DO get logged out while you're working in the unit as a student and the session times out (test this by logging out of the session in another browser tab, or by running ```flushall``` in redis)

Closes #709


breity (approved):
Looks good.

PR #712, opened 2022-08-04
chore(NodeInfo): Upgrade to Angular

### LINKED ISSUE #711: Upgrade NodeInfo to Angular
Upgrade NodeInfo from AngularJS to Angular. It is used in CM to show step info in a dialog.

hirokiterashima:
## Notes
- @breity the rubric style is a bit off. Can you take a look please?

Screen Shot 2022-08-04 at 3 14 37 PM

## Changes
- Upgrade NodeInfo to Angular
- Code clean up

## Test
- These are displayed in the step info popup in the CM
   - Component preview 
   - Component summary, if applicable (e.g. scores, MC choices)
   - Step and component rubric, if authored
   - Step border color outline, if authored
 
Closes #711


breity (commented):
Looks good.

I fixed and updated the styling. I also removed the Step border color outline because i don't think it's necessary here.

Looks like Table components also have some display problems in the dialog. Header height isn't properly calculated, for example. So you can't fully see the column header text. Looks like this happens on production, so it's an existing problem. Can create a new issue for it.

PR #717, opened 2022-08-06
fix(Data Explorer): X axis label now updates when changing x column

### LINKED ISSUE #716: fix(Data Explorer): X axis label does not update when selecting a different column
### Description

- As a student, go to a Data Explorer step
- Select a column for the X Data
- Select a column for the Y Data
- The graph should get rendered and the X Data name should be displayed as the X axis label
- Change the X Data column to a different option
- The graph should get re-rendered and the new X Data name should be displayed as the X axis label. Previously the X axis label would not change.

geoffreykwan:
## Changes

The X axis label is now always updated when the student changes the X Data column.

## Test

- As a student, go to a Data Explorer step
- Select a column for the X Data
- Select a column for the Y Data
- The graph should get rendered and the X Data name should be displayed as the X axis label
- Change the X Data column to a different option
- The graph should get re-rendered and the new X Data name should be displayed as the X axis label. Previously the X axis label would not change.

Closes #716


breity (approved):
Looks good. 👍

PR #720, opened 2022-08-08
refactor(Authoring Tool): Remove notSignedIn handler on project save

### LINKED ISSUE #719: AT: Remove notSignedIn handler on save project
When project saving is requested and the user is logged out, we no longer handle it in TeacherProjectService.handleSaveProjectResponse() "notSignedIn" conditional. Instead, we handle it in http-error.interceptor.ts, so the handler code is never called. Remove this handler code and other code that it uses.

hirokiterashima:
## Changes
- Remove ```if (response.messageCode === 'notSignedIn') {``` error response handler from authoring tool when project was saved while use is logged out. We handle this case now in http-error.interceptor.ts, so the error response handler code is never called.

## Test
- In tab 1, open unit in AT
- In another tab, log out of WISE
- In tab 1, make change in AT. You should see the log in page with the message "There was an error saving data..."

Closes #719

### DIFF
--- a/src/app/services/teacherProjectService.spec.ts
+++ b/src/app/services/teacherProjectService.spec.ts
@@ -613,12 +613,6 @@ function shouldHandleSaveProjectResponse() {
   it('should broadcast project saved', () => {
     shouldHandleSaveProjectResponseSuccessHelper('broadcastProjectSaved');
   });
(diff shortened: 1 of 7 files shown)


geoffreykwan (approved):
Looks good.

PR #722, opened 2022-08-08
feat(Table): Set default max width for Tabulator columns

### LINKED ISSUE #721: Table: Set default column maxWidth
### Description
Currently, Tabulator sets the column width to match the width of the header label and/or cell content by default. Long header labels can result in very wide columns. We should set a default maxWidth that can be overridden by authoring the global or individual column width.

breity:
## Changes
Sets the default max width for each Tabulator column to 200px. 
This can be overridden by authoring the `globalCellSize` or individual column widths.

Closes #721.

## Test
- Author a Table component with a long column header (1st row) text.
- View table in preview mode or as a student.
- Verify that Tabulator restricts the column width to 200px and wraps the header text.
- Back in authoring, enter a value for 'Global Cell Size", e.g. 20.
- Verify in VLE that the Tabulator columns are wider than 200px (should be 320px for a cell size of 20).
- In authoring, clear Global Cell Size and enter a value for 'Column Cell Size' for the column with the long header.
- Verify in VLE that the column is wider than 200px (should be 320px for a cell size of 20).

### DIFF
--- a/src/assets/wise5/components/table/TabulatorData.ts
+++ b/src/assets/wise5/components/table/TabulatorData.ts
@@ -10,7 +10,10 @@ export class TabulatorData {
     const defaultOptions = {
       layout: 'fitDataTable',
       maxHeight: '500px',
-      reactiveData: true
+      reactiveData: true,
+      columnDefaults: {
+        maxWidth: 200
+      }
     };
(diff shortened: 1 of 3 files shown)


hirokiterashima (approved):
Looks good

PR #730, opened 2022-08-11
fix(Table): Make sure correct cells are editable by students

### LINKED ISSUE #727: First and last rows are not editable in Table component preview
(report from Cristian R) 

This behavior likely started happening after our switch to Tabulator.

How to reproduce:
1. Create a new table component. All the cells should be "editable" by default
2. Preview the component

The first row (header row) and the last row are not editable. This is what it looks like: 
Screen Shot 2022-08-10 at 4 21 04 PM

breity:
## Changes
Correct cells are now editable in student Table components. Previously, checking a cell as editable in the Table authoring would make the corresponding cell in the row above editable and not the checked cell.

PR also fixes a bug that would cause cells in the wrong row to be updated when table had a custom sorting set.

Closes #727.
[Note: Cells in the first row of a Table component (header row) are never editable now that we use Tabulator. We should probably remove the editable checkbox from first row cells in the Table authoring. Could explore adding an option to allow student to edit headers in the future.]

## Test
- Make sure The same cells that are checked as editable in Table authoring are actually editable on the student side.
- Make sure that the correct cell values are saved when you edit a table that has custom sorting applied.

### DIFF
--- a/src/app/services/tabulatorDataService.spec.ts
+++ b/src/app/services/tabulatorDataService.spec.ts
@@ -32,7 +32,7 @@ function convertTableDataToTabulator() {
       expect(tabulatorData.data.length).toBe(3);
       expect(tabulatorData.data[1]['0']).toBe('10');
       expect(tabulatorData.options.maxHeight).toBe('500px');
-      expect(tabulatorData.editableCells[0]).toEqual(['1','2']);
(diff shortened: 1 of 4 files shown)


geoffreykwan (approved):
Looks good.

PR #738, opened 2022-08-16
feat(Dialog Guidance): Add hasKIScore(num) token expression rule

### LINKED ISSUE #737: Add hasKIScore(num) token to Dialog Guidance feedback rule expression
The rule author can specify a number within the parenthesis to indicate which KI score should match this token. The number should be between 1 and 5, inclusive.

hirokiterashima:
## Changes
- Add hasKIScore(num) token to Dialog Guidance feedback expression rule, where 1 <= num <= 5
- CRaterResponse.getKIScore() now works for single-score items
- Remove unused code

## Test
- Add "hasKIScore(X)" feedback rules for both multiple-score item ("MusicalInstruments_Speed_score-ki_idea-ki_nonscor") and single-score item ("GREENROOF-II") and make sure that they are matched and the corresponding feedback is shown
- Try using operators to combine the token like ```hasKIScore(2) || hasKIScore(3)``` and make sure that it is matched and the corresponding feedback is shown

Closes #737

### DIFF
--- a/src/assets/wise5/components/dialogGuidance/DialogGuidanceFeedbackRuleEvaluator.ts
+++ b/src/assets/wise5/components/dialogGuidance/DialogGuidanceFeedbackRuleEvaluator.ts
@@ -114,11 +114,24 @@ export class DialogGuidanceFeedbackRuleEvaluator {
     return !this.evaluateTerm(termStack.pop(), response);
   }
 
-  private evaluateOrExpression(term1: string, term2: string, response: CRaterResponse): boolean {
-    return this.evaluateTerm(term1, response) || this.evaluateTerm(term2, response);
+  private evaluateTerm(term: string, response: CRaterResponse): boolean {
+    if (this.isHasKIScoreTerm(term)) {
+      return this.evaluateHasKIScoreTerm(term, response);
+    } else {
+      return this.evaluateIdeaTerm(term, response);
+    }
   }
 
-  private evaluateTerm(term: string, response: CRaterResponse): boolean {
+  private isHasKIScoreTerm(term: string): boolean {
+    return /hasKIScore\([1-5]\)/.test(term);
+  }
+
+  private evaluateHasKIScoreTerm(term: string, response: CRaterResponse): boolean {
+    const expectedKIScore = parseInt(term.match(/hasKIScore\((.*)\)/)[1]);
+    return response.getKIScore() === expectedKIScore;
+  }
+
+  private evaluateIdeaTerm(term: string, response: CRaterResponse): boolean {
     return term === 'true' || response.getDetectedIdeaNames().includes(term);
   }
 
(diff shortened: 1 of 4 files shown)


src/assets/wise5/components/dialogGuidance/DialogGuidanceFeedbackRuleEvaluator.ts
```diff
@@ -114,11 +114,24 @@ export class DialogGuidanceFeedbackRuleEvaluator {
...
+      return this.evaluateHasKIScoreTerm(term, response);
+    } else {
+      return this.evaluateIdeaTerm(term, response);
+    }
   }
 
-  private evaluateTerm(term: string, response: CRaterResponse): boolean {
+  private isHasKIScoreTerm(term: string): boolean {
+    return /hasKIScore\([1-5]\)/.test(term);
+  }
```
geoffreykwan, line 127: Should this be /hasKIScore\\(\d+\\)/ in case we ever have item ids that return scores more than 5?

geoffreykwan (commented):
Functionality works well. I just have one question that I commented on.

src/assets/wise5/components/dialogGuidance/DialogGuidanceFeedbackRuleEvaluator.ts
```diff
@@ -114,11 +114,24 @@ export class DialogGuidanceFeedbackRuleEvaluator {
...
+    if (this.isHasKIScoreTerm(term)) {
+      return this.evaluateHasKIScoreTerm(term, response);
+    } else {
+      return this.evaluateIdeaTerm(term, response);
+    }
   }
 
-  private evaluateTerm(term: string, response: CRaterResponse): boolean {
+  private isHasKIScoreTerm(term: string): boolean {
+    return /hasKIScore\([\d]\)/.test(term);
```
geoffreykwan: Should the regex use \d+ in case we have scores that are multiple digits in the future?

I think it should be ```/hasKIScore\(\d+\)/``` (brackets are not needed).

geoffreykwan (changes requested):

src/assets/wise5/components/dialogGuidance/DialogGuidanceFeedbackRuleEvaluator.ts
hirokiterashima: Hmm I disagree here. I don't think we should code to handle a case in the future that may or may not happen.

On further thought, I think we should actually revert it to checking for [1-5] because those are the KI scores that we work with now, and we don't foresee any score values outside of those ranges in the near future. The code is more precise this way, so the reader can see that we only allow values 1-5 when checking for KI score. If the score range should change in the future, we can update the code then.

hirokiterashima:
I reset the commit to the one that was checking for KI score in the [1-5] range, as per our discussion.

geoffreykwan (changes requested):
Should there be tests to test the boundaries of the hasKIScore() range of acceptable values?

geoffreykwan (approved):
Looks good.

PR #742, opened 2022-08-17
fix(Graph): Categories data point authoring

### LINKED ISSUE #741: Graph categories data point authoring bug
### Description

2. Create a Graph component in a step
3. Set the X Axis type to Categories
4. Go down to the Series and add two Data Points
5. Type a number with two digits like 11 into the first data point input "Category One"
6. When you are typing, the focus will switch to the second data point input "Category Two"
7. You will end up with the first input containing a 1 and the second input containing 11. This should not happen.
8. Try adding some more digits into the second data point input, it will only let you enter one at a time because the input will lose focus after you enter a digit. This should not happen and you should be able to type multiple digits without being interrupted.

geoffreykwan:
## Test

2. Create a Graph component in a step
3. Set the X Axis type to Categories
4. Go down to the Series and add two Data Points
5. Type a number with two digits like 11 into the first data point input "Category One"
6. 11 should show up in the input. Previously, you would end up with 1 in the first data point input and 11 in the second data point input "Categories Two" because it would switch focus to the second data point input for some reason.
7. Try adding some digits into the second data point input "Categories Two"
8. The digits should show up as expected. Previously, it would only let you enter one digit at a time because the input would lose focus after you entered a digit.

Closes #741

### DIFF
--- a/src/assets/wise5/components/graph/graph-authoring/graph-authoring.component.html
+++ b/src/assets/wise5/components/graph/graph-authoring/graph-authoring.component.html
@@ -431,7 +431,7 @@
   <div *ngIf="series.data == null || series.data.length == 0">
     This series has no starter data points. Click the "Add Data Point" button to add a starter data point if you'd like.
   </div>
-  <div *ngFor="let dataPoint of series.data; index as dataPointIndex; first as isFirst; last as isLast">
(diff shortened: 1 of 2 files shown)


breity (approved):
Looks good.

PR #744, opened 2022-08-18
fix(Graph Component): @HostListener interferes with Table interactions

### LINKED ISSUE #734: @HostListener in Graph interferes with Table interactions
## Description
When a Table and connected Graph component are on the same step, pressing backspace doesn't work in table cells, throws error:
```
ERROR TypeError: Cannot read properties of null (reading 'canEdit')
    at kt.canEdit (default-src_app_notebook_notebook-notes_notebook-notes_component_ts-src_app_notebook_notebook-cb2d36.b38fac9fec715a69.js:1:327246)
    at kt.handleDeleteKeyPressed (default-src_app_notebook_notebook-notes_notebook-notes_component_ts-src_app_notebook_notebook-cb2d36.b38fac9fec715a69.js:1:334914)
...
```

### Steps to reproduce
1. Create a step with a Table with some editable cells and Graph component.
2. Connect components by setting Graph component to import work from table.
3. Launch/preview the unit.
4. Type a value into one of the table cells, then try to delete the value by pressing backspace (delete on Mac).

hirokiterashima:
## Changes
- Add hidden button that will gain focus (i.e. set it as document.activeElement) when user clicks anywhere on the graph. This allows the component to throw events (e.g. "keydown.backspace")
- @HostListener in GraphComponent only listens to events from component. Before, it was listening to events from the entire document.

## Test
- In Graph component, add and delete points
   - test on a step with multiple graph components 
- In DataExplorer activity, type a value into one of the table cells, then try to delete the value by pressing backspace. This should delete the value from the cell.

Closes #734

### DIFF
--- a/src/assets/wise5/components/graph/graph-student/graph-student.component.html
+++ b/src/assets/wise5/components/graph/graph-student/graph-student.component.html
@@ -66,6 +66,11 @@
   </mat-form-field>
   <div>
     <button
+        #hiddenButton
+        style="height:0; width:0; padding: 0; border: 0 none"
+        aria-role="none"
+        tabindex="-1"></button>
+    <button
         mat-button
         *ngIf="isResetSeriesButtonVisible && activeSeries != null && activeSeries.canEdit"
         color="accent"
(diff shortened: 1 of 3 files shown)


breity (approved):
Looks good.

PR #748, opened 2022-08-18
fix(Data Explorer): Fix rendering with empty cells

### LINKED ISSUE #747: Data Explorer graph does not render if there are empty values
### Description

Create a Data Explorer step that contains a column that contains empty cells like in the step below

https://wise.berkeley.edu/preview/unit/38172/node214

For "X Data" choose "Demographic"
For "Y Data 1" choose "Percent COVID-19 Deaths"

The graph will not render but it should.

geoffreykwan:
## Changes

When a cell does not have a value, we now add an empty array [] to the data point array instead of ['N/A', 'N/A'].

## Test

Make sure Data Explorer still renders if some cells in a column are empty like in the step below

For "X Data" choose "Demographic"
For "Y Data 1" choose "Percent COVID-19 Deaths"

https://wise.berkeley.edu/preview/unit/38172/node214

Closes #747

### DIFF
--- a/src/assets/wise5/components/graph/graph-student/graph-student.component.ts
+++ b/src/assets/wise5/components/graph/graph-student/graph-student.component.ts
@@ -1708,28 +1708,26 @@ export class GraphStudent extends ComponentStudent {
   }
 
   addPointFromTableIntoData(xCell: any, yCell: any, data: any[]) {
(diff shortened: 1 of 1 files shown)


breity (approved):
Looks good.

PR #749, opened 2022-08-18
feat(Dialog Guidance): Add IdeaCount tokens to Feedback Rule expression

### LINKED ISSUE #739: Add ideaCount*(num) tokens to Dialog Guidance feedback rule expression
The author can use these new tokens when authoring Feedback Rules for Dialog Guidance: 

- ideaCountEquals(num): student has exactly num ideas
- ideaCountMoreThan(num): student has more than num ideas (exclusive)
- ideaCountLessThan(num): student has less than num ideas (exclusive)

A number within the parenthesis should indicate the number of ideas that will match this token. This number should be 0 or greater.

hirokiterashima:
## Changes
- Add ideaCount(MoreThan|Equals|LessThan)(num) tokens to Dialog Guidance feedback expression rule
   - ideaCountMoreThan(num): student has more than num ideas (exclusive)
   - ideaCountEquals(num): student has exactly num ideas
   - ideaCountLessThan(num): student has less than num ideas (exclusive)
- Clean up test

## Test
- Add "ideaCount(MoreThan|Equals|LessThan)(num)" feedback rules for both multiple-score item ("MusicalInstruments_Speed_score-ki_idea-ki_nonscor") and single-score item ("GREENROOF-II") and make sure that they are matched and the corresponding feedback is shown
   - For example, add these rules for MusicalInstruments_Speed_score-ki_idea-ki_nonscor item and test that they're matched
      - ideaCountLessThan(2) is matched by the response "Bubbles. He's under water."
      - ideaCountEquals(2) is matched by the response "Bubbles. Sound travels faster under water compared to air, so it reaches Bubbles first."
      - ideaCountMoreThan(2) is matched by the response "Bubbles. Sound travels faster under water compared to air, so it reaches Bubbles first. It's more dense in the water, so molecules are closer together."

Closes #739

### DIFF
--- a/src/assets/wise5/components/dialogGuidance/DialogGuidanceFeedbackRuleEvaluator.ts
+++ b/src/assets/wise5/components/dialogGuidance/DialogGuidanceFeedbackRuleEvaluator.ts
@@ -132,12 +132,13 @@ export class DialogGuidanceFeedbackRuleEvaluator {
     const matches = term.match(/ideaCount(.*)\((.*)\)/);
     const comparer = matches[1];
     const expectedIdeaCount = parseInt(matches[2]);
-    if (comparer === 'MoreThan') {
-      return response.getDetectedIdeaCount() > expectedIdeaCount;
-    } else if (comparer === 'Equals') {
-      return response.getDetectedIdeaCount() === expectedIdeaCount;
-    } else {
-      return response.getDetectedIdeaCount() < expectedIdeaCount;
+    switch (comparer) {
+      case 'MoreThan':
+        return response.getDetectedIdeaCount() > expectedIdeaCount;
+      case 'Equals':
+        return response.getDetectedIdeaCount() === expectedIdeaCount;
+      case 'LessThan':
+        return response.getDetectedIdeaCount() < expectedIdeaCount;
     }
   }
 
--- a/src/assets/wise5/components/dialogGuidance/DialogGuidanceFeedbackRuleEvaluator.ts
+++ b/src/assets/wise5/components/dialogGuidance/DialogGuidanceFeedbackRuleEvaluator.ts
@@ -117,8 +117,27 @@ export class DialogGuidanceFeedbackRuleEvaluator {
   private evaluateTerm(term: string, response: CRaterResponse): boolean {
     if (this.isHasKIScoreTerm(term)) {
       return this.evaluateHasKIScoreTerm(term, response);
+    } else if (this.isIdeaCountTerm(term)) {
+      return this.evaluateIdeaCountTerm(term, response);
     } else {
       return this.evaluateIdeaTerm(term, response);
+    }
+  }
+
+  private isIdeaCountTerm(term: string): boolean {
+    return /ideaCount(MoreThan|Equals|LessThan)\([\d+]\)/.test(term);
+  }
+
+  private evaluateIdeaCountTerm(term: string, response: CRaterResponse): boolean {
+    const matches = term.match(/ideaCount(.*)\((.*)\)/);
+    const comparer = matches[1];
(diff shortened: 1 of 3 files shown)


src/assets/wise5/components/dialogGuidance/DialogGuidanceFeedbackRuleEvaluator.ts
```diff
@@ -117,11 +117,30 @@ export class DialogGuidanceFeedbackRuleEvaluator {
...
+    const matches = term.match(/ideaCount(.*)\((.*)\)/);
+    const comparer = matches[1];
+    const expectedIdeaCount = parseInt(matches[2]);
+    if (comparer === 'MoreThan') {
+      return response.getDetectedIdeaCount() > expectedIdeaCount;
+    } else if (comparer === 'Equals') {
+      return response.getDetectedIdeaCount() === expectedIdeaCount;
+    } else {
+      return response.getDetectedIdeaCount() < expectedIdeaCount;
+    }
```
geoffreykwan: Should we use a switch here?

```
    switch (comparer) {
      case 'MoreThan':
        return response.getDetectedIdeaCount() > expectedIdeaCount;
      case 'Equals':
        return response.getDetectedIdeaCount() === expectedIdeaCount;
      case 'LessThan':
        return response.getDetectedIdeaCount() < expectedIdeaCount;
      default:
        return false;
    }
```

geoffreykwan (changes requested):
Functionality works well. I have a code change suggestion below.

src/assets/wise5/components/dialogGuidance/DialogGuidanceFeedbackRuleEvaluator.ts
hirokiterashima: Hmm. This increases the complexity of this function from 8 to 10 because of the "default" case. 

How about 
```
   switch (comparer) {
      case 'MoreThan':
        return response.getDetectedIdeaCount() > expectedIdeaCount;
      case 'Equals':
        return response.getDetectedIdeaCount() === expectedIdeaCount;
      default:
        return response.getDetectedIdeaCount() < expectedIdeaCount;
    }
```
since we can assume that those are the only possible values that comparer could be inside this function?

geoffreykwan: I think it would be better to be more explicit and have a 'LessThan' case as well as a default.

hirokiterashima: We would never reach the default case in your code example because the values for comparer can only be "MoreThan", "Equals", or "LessThan". ```evaluateIdeaCountTerm()``` is a private function, and it is only called when ```isIdeaCountTerm()``` is true, which guarantees that the comparer is one of the three possible values.

If you want to add a default case, we can throw an error instead of returning a false value. Returning a false value will allow the program to continue, which may result in unwanted/unexpected behaviors down the line.

```
    switch (comparer) {
      case 'MoreThan':
        return response.getDetectedIdeaCount() > expectedIdeaCount;
      case 'Equals':
        return response.getDetectedIdeaCount() === expectedIdeaCount;
      case 'LessThan':
        return response.getDetectedIdeaCount() < expectedIdeaCount;
      default:
        throw new Error("Unexpected comparer value.");
    }
```

geoffreykwan: Yea, throwing an error there is good.

hirokiterashima:
Changed to a switch statement that checks for the 3 possible values as we discussed during the meeting.

geoffreykwan (approved):
Looks good.

PR #759, opened 2022-08-22
refactor(VLEComponent): Use @HostListener to snip image to notebook

### LINKED ISSUE #758: refactor(VLEComponent): change SnipImageListener to use @HostListener
Right now the VLEComponent calls window.addEventListener to register a listener for snipping images to the notebook, and keeps a reference to the handler in the VLEComponent class.

Instead, use Angular's @HostListener annotation. This will eliminate the need to call window.removeEventListener() as well.

hirokiterashima:
## Changes
- Use @HostListener instead of calling window.addEventListener/removeEventListener
- Fix issue where snip image was not working in preview

## Test
- Snipping image to notebook works in both student and preview. Before, there was an in issue snipping image in preview

Closes #758

### DIFF
--- a/src/assets/wise5/services/studentAssetService.ts
+++ b/src/assets/wise5/services/studentAssetService.ts
@@ -85,7 +85,7 @@ export class StudentAssetService {
 
   getFileNameFromAsset(asset) {
     if (this.ConfigService.isPreview()) {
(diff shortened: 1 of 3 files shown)


geoffreykwan (approved):
Looks good.

PR #762, opened 2022-08-23
refactor(VLEComponent): Remove unused server connection handling

### LINKED ISSUE #761: refactor(VLEComponent): Remove unused server connection handling
### Description

After [this PR](https://github.com/WISE-Community/WISE-Client/pull/656
) when session times out while using VLE, AT, CM, we show message indicating an error and prompt user to login again. Upon logging in, they will be redirected to where they where before. VLEComponent does not handle the server connection status any more. Remove the code.

hirokiterashima:
## Changes
- Remove server connection handling from VLEComponent, which used to show a snack message. But after this PR: https://github.com/WISE-Community/WISE-Client/pull/656, when the student is logged out of the server, they will immediately see the login page. This is handled by http-error-interceptor. The code in VLEComponent is no longer used.
- Remove unused numberProjects variable

## Test
- While using VLE as a student, log out of the session in another tab. Then, try saving work by making changes or going to the next step.
   - They should be redirected to the log in page and see an error message, prompting them to log back in and check recent work
   - When they log in, they should be redirected to where they were before in the app 

Closes #761

### DIFF
--- a/src/assets/wise5/vle/vle.component.html
+++ b/src/assets/wise5/vle/vle.component.html
@@ -7,9 +7,7 @@
   </mat-drawer>
   <mat-drawer-content>
     <top-bar></top-bar>
-    <step-tools *ngIf="layoutState === 'node'"
-        class="control-bg-bg mat-elevation-z1"
-        show-position="numberProject"></step-tools>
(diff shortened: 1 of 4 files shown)


geoffreykwan (approved):
Looks good.

PR #767, opened 2022-08-23
fix(Graph): Second editable series shows tooltip

### LINKED ISSUE #754: Graph second series does not display tooltips
### Description

1. Create a Graph that contains two series
2. Make sure "Can Student Edit" is true for both series
3. Add a data point to each series
4. Preview the Graph
5. Mouseover the point from the first series, the tooltip should be displayed
6. Mouseover the point from the second series, the tooltip will not be displayed even though it should be. If you select the second series as the "Select series to edit", it will display the tooltips on mouseover.

geoffreykwan:
## Changes

Always allow data point mouse over tooltip.

## Test

1. Create a Graph that contains two series
2. Make sure "Can Student Edit" is true for both series
3. Add a data point to each series
4. Preview the Graph
5. Mouseover the point from the first series, the tooltip should be displayed
6. Mouseover the point from the second series, the tooltip should be display. Previously, the tooltip would not be displayed.

Closes #754

### DIFF
--- a/src/assets/wise5/components/graph/graph-student/graph-student.component.ts
+++ b/src/assets/wise5/components/graph/graph-student/graph-student.component.ts
@@ -871,9 +871,8 @@ export class GraphStudent extends ComponentStudent {
   }
 
   setAllSeriesFields(series) {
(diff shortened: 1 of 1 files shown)


breity (approved):
Looks good.

PR #770, opened 2022-08-23
fix(Data Explorer): Fix x axis N/A tick label

### LINKED ISSUE #769: fix(Data Explorer): Fix x axis N/A tick label
### Description

1. Go to a Data Explorer step like this https://wise.berkeley.edu/preview/unit/38172/node210
2. For X Data choose "% 12+ fully vaccinated"
3. For Y Data 1 choose "% 12+ fully vaccinated with a first booster dose"
4. You will see an N/A value in the x axis tick label (between 22 and 26). It should display a number instead of N/A.

geoffreykwan:
1. Go to a Data Explorer step like this https://wise.berkeley.edu/preview/unit/38172/node210
2. For X Data choose "% 12+ fully vaccinated"
3. For Y Data 1 choose "% 12+ fully vaccinated with a first booster dose"
4. There used to be an N/A tick label between 22 and 26 because the 24th row (Modoc County) has N/A in the cells for "% 12+ fully vaccinated" and "% 12+ fully vaccinated with a first booster dose". Now it should properly display the number 24 there instead.
5. Make sure if the x value for a cell is NA or N/A or na or n/a that it does not show up as an x axis tick label.

Closes #769

### DIFF
--- a/src/assets/wise5/components/graph/graph-student/graph-student.component.ts
+++ b/src/assets/wise5/components/graph/graph-student/graph-student.component.ts
@@ -399,7 +399,12 @@ export class GraphStudent extends ComponentStudent {
             studentData.tableData,
             this.value
           );
-          if (textValue !== '' && isNaN(parseFloat(textValue))) {
+          if (
+            textValue !== '' &&
+            textValue !== 'NA' &&
+            textValue !== 'N/A' &&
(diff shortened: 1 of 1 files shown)


breity (approved):
Looks good.

PR #771, opened 2022-08-23
chore(MilestoneWorkgroupItem): Upgrade to Angular

### LINKED ISSUE #765: Upgrade MilestoneWorkgroupItem to Angular

hirokiterashima:
## Note
- @breity please style as you see fit. I tried to copy the styles from your changes in the recent WorkgroupItem PR.
- One issue: in milestone-workgroup-item.component.html, changing "md-display-1" to "mat-display-1" messes up the alignment of the "Change in Score" text.

## Changes
- Upgrade MilestoneWorkgroupItem to Angular
- Code clean up
   - Remove WorkgroupItem.ts (WorkgroupItemController) as this is no longer used
   - Fix indenting issue in workgroup-item.component.html

## Test
- In the CM's milestone report dialog, the workgroup item appear with the correct status (Completed, no work, not assigned, ...). 
   - If this is a milestone report with multiple location, it should properly show the two scores and the change in scores.
- Click on a WorkgroupItem, and it should toggle visibility of the grading view for that workgroup
- Grading (score, comments) works as before

Closes #765


breity (approved):
Looks good.

I think styling is fine for now. We're still using some styles from AngularJS, but I think it makes sense to go ahead with the component upgrade as is for now. Then we cn address removing remaining legacy styles once most of the teacher tools are upgraded.

PR #782, opened 2022-08-29
fix(Notebook): Fix notebook tools to window in VLE

breity:
## Changes
Notebook report, launcher, and notes should now be fixed to the browser window and always visible in the VLE again. (A previous update had attached them to the bottom of the VLE content.)

Closes #735.

## Test
- Check that the notebook tools are always visible when viewing a unit as a student or in preview.
- Make sure the position of the report positioning updates correctly when notes bar is opened and closed, report is expanded to full screen, small view, and collapsed.
- Test with different window sizes (from mobile to desktop widths).
- Make sure editing and saving report and notes content works as before.

PR #789, opened 2022-08-31
feat(VLE): Tabbed Navigation Theme

### LINKED ISSUE #788: Tabbed Navigation Theme
### Description

Add a new theme for the VLE that does the following
- show lessons as tabs at the top
- show prev/next button at the bottom of the content

hirokiterashima:
## Changes
- Add new "tab" theme for the VLE

## Test Prep
- Add ```"theme": "tab"``` as top-level key/value in project.json

## Test
- Default theme: Preview/Student works as before
- Tabbed theme
   - tabs appear at the top of the screen, one for each lesson in the unit
      - click on a tab. It should take you to the first step in that lesson
   - prev/next buttons appear at the bottom of the page after all the components
      - they are disabled if there is no prev/next, including when you can't go next because of constraint 
      - click on prev/next should take you to prev/next step in the unit. 
         - it should always scroll to the top of the page after navigating to the new step 
         - the selected tab should reflect the lesson that the current step is in. 

Closes #788


geoffreykwan (changes requested):
Feature works well in general. I just noticed some minor things.

1. If a lesson is locked, the step before the locked lesson has a clickable "Next Step" button. I think this is ok though because if they click the "Next Step" button, the constraint popup shows up.

2. If the student clicks on a locked lesson, the tab underline highlight still moves under the locked lesson even though the student isn't actually in that lesson.

3. In the Authoring Tool and Classroom Monitor, the WISE logo in the upper left does not load.

4. In the Authoring Tool and Classroom Monitor, if you click the upper right avatar icon, the account menu does not open.

5. Students need to scroll to the bottom of the page to access Notes and Report.

6. The tab area has some extra vertical white space so we could reduce the height of that area to give more space for step content.

7. Nice to have but not required: If there are a lot of lesson tabs, it might be nice if you could scroll through the lesson tabs sideways with the mouse wheel or touchpad.

hirokiterashima:
> Feature works well in general. I just noticed some minor things.
> 
> 1. If a lesson is locked, the step before the locked lesson has a clickable "Next Step" button. I think this is ok though because if they click the "Next Step" button, the constraint popup shows up.
> 2. If the student clicks on a locked lesson, the tab underline highlight still moves under the locked lesson even though the student isn't actually in that lesson.

Locked lesson is no longer clickable. Please test and make sure they become enabled/disabled in real-time as teacher unlocks/locks a lesson in the CM.

> 3. In the Authoring Tool and Classroom Monitor, the WISE logo in the upper left does not load.

Fixed

> 4. In the Authoring Tool and Classroom Monitor, if you click the upper right avatar icon, the account menu does not open.

Fixed

> 5. Students need to scroll to the bottom of the page to access Notes and Report.

Fixed by merging in #782 

> 6. The tab area has some extra vertical white space so we could reduce the height of that area to give more space for step content.
> 7. Nice to have but not required: If there are a lot of lesson tabs, it might be nice if you could scroll through the lesson tabs sideways with the mouse wheel or touchpad.

@breity can you look at 6&7 and other styling issues?

breity (approved):
Looks good. I updated the styles to fix the tab bar to the top and position prev/next buttons.

Some topics I think we should address in subsequent issues:
- UI:
  - No way to jump to different steps in lessons - must go next/previous in order
  - No way to know which step you're on in a lesson or how many there are - mini nav mode would be nice
  - Need to scroll all the way down to navigate is frustrating if you want to quickly go through steps
  - Show lock icon on disabled lessons?
- Theming architecture:
  - Customizations are built into WISE core, need to customize vle.html to add a theme
  - Maybe each theme should have its own vle.html?
  - Components specific to theme are mixed in with default theme components
  - Would be better to group theme assets and components together

geoffreykwan (changes requested):
Clicking next step does not scroll to the top.

1. Create two steps one after the other that have a lot of vertical content (you can just author a prompt with a lot of br tags)
2. Preview the project
3. Go to the first step
4. Scroll down to the bottom of the step
5. Click the "Next Step" button
6. You will go to the second step but will be at the bottom of the step. Clicking "Next Step" should bring you to the top of the step.

Note: If you click "Previous Step" it does properly bring you to the top of the step.

breity:
Thanks @geoffreykwan. I fixed the scroll to top. Take a look and make sure it's working now for both default and tab themes.

geoffreykwan (approved):
Looks good.

PR #796, opened 2022-09-06
chore(Classroom Monitor): Convert NodeGradingView

### LINKED ISSUE #790: Convert NodeGradingView to Angular
### Description

geoffreykwan:
## Changes

Converted NodeGradingView to Angular.

## Test

2. In the "Grade By Step" view, click on a step
3. Make sure everything works like it did before
- Completion percentage
- Mean score
- Showing x/x items
- Info
- Searching for workgroups
- Expand All
- Collapse All
- Expanding a single workgroup
- Sorting by Team
- Sorting by Status
- Sorting by Score

There is some styling that needs to be fixed
- The bar that contains the workgroup search and "Expand all" and "Collapse all" buttons is a little too wide now
- When you expand a workgroup, it now has left and right side padding when it used to have none
- If you mouse over the "Showing x/x items" it shows the outline of a button (this seems to also happen on production so maybe this problem was introduced somewhere else)
- If you make the browser window narrow, some elements used to wrap but now they don't
  - The "Expand All" and "Collapse All" buttons are supposed to wrap under the workgroup search input but it no longer does that
  - The percent completion and mean score are supposed to be centered when the "Showing x/x items" and "Info" buttons wrap underneath them but they aren't centered anymore

Screenshot 2022-09-06 171922

Closes #790


breity (approved):
Looks good. I fixed the styling and layout issues.

PR #798, opened 2022-09-07
fix(Login): Add redirectUrl to Google login url

### LINKED ISSUE #797: Send redirectUrl to Google login
### Description

When the user tries to access a secured path and logs in with Google, the path is not remembered. We need to send this redirectUrl to Google so that the user is taken to the path once they log in.

hirokiterashima:
## Changes
- When google login button is clicked in the homepage, it will append the redirect url to the google login url

## Test
- Test with PR https://github.com/WISE-Community/WISE-API/pull/176

Closes #797

### DIFF
--- a/src/app/login/login-home/login-home.component.ts
+++ b/src/app/login/login-home/login-home.component.ts
@@ -124,8 +124,6 @@ export class LoginHomeComponent implements OnInit {
   }
 
   private appendAccessCodeParameter(url: string): string {
-    return url.includes('?')
-      ? `${url}&accessCode=${this.accessCode}`
-      : `${url}?accessCode=${this.accessCode}`;
+    return `${url}${url.includes('?') ? '&' : '?'}accessCode=${this.accessCode}`;
   }
 }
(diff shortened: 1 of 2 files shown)


geoffreykwan (changes requested):
Works well. I have a code clean up suggestion.

src/app/login/login-home/login-home.component.ts
```diff
@@ -113,13 +113,19 @@ export class LoginHomeComponent implements OnInit {
...
-      redirectUrl = `${redirectUrl}?accessCode=${this.accessCode}`;
+      redirectUrl = this.appendAccessCodeParameter(redirectUrl);
     }
     return redirectUrl;
   }
+
+  private appendAccessCodeParameter(url: string): string {
+    return url.includes('?')
+      ? `${url}&accessCode=${this.accessCode}`
+      : `${url}?accessCode=${this.accessCode}`;
```
geoffreykwan: Should this be rewritten to reduce duplication? Something like
```
return `${url}${url.includes('?') ? '%' : '?'}accessCode=${this.accessCode}`;
```
or
```
const delimiter = url.includes('?') ? '%' : '?';
return `${url}${delimiter}accessCode=${this.accessCode}`;
```

breity (approved):
Looks good.

geoffreykwan (approved):
Looks good.

PR #806, opened 2022-09-09
refactor(StudentGradingToolsComponent): Reduce duplication

### LINKED ISSUE #805: Refactor(StudentGradingToolsComponent): Reduce duplication
### Description

Motivation:
What if we remove workgroups from this.workgroups instead of setting visible=true/false? So this.workgroups will always contain only the visible workgroups. Then the functions would become something like this:
```
getPreviousWorkgroupId(workgroupId: number): number {  
   const workgroupIdIndex = this.getIndexOfWorkgroupId(workgroupId);
   return workgroupIdIndex == 0 ? null : this.workgroups[workgroupIdIndex -1];
}

getNextWorkgroupId(workgroupId: number): number {   
   const workgroupIdIndex = this.getIndexOfWorkgroupId(workgroupId);
   return workgroupIdIndex == this.workgroups.length ? null : this.workgroups[workgroupIdIndex + 1];
}
```

hirokiterashima:
## Changes
- ```this.workgroups``` only contains workgroups for current period
- Code cleanup
   - Add ```private``` keywords
   - Define Workgroup class

## Test
1. Go to the Grading Tool for a run
2. Go to the Grade By Team view
3. Select a workgroup
4. Make sure the workgroup select drop down works like before
- Next workgroup
- Previous workgroup
- Jump to workgroup
- Search for workgroup
- Make sure switching periods properly filters workgroups

Closes #805

### DIFF
--- a/src/assets/wise5/classroomMonitor/classroomMonitorComponents/studentGrading/student-grading-tools/student-grading-tools.component.ts
+++ b/src/assets/wise5/classroomMonitor/classroomMonitorComponents/studentGrading/student-grading-tools/student-grading-tools.component.ts
@@ -5,6 +5,11 @@ import { ConfigService } from '../../../../services/configService';
 import { TeacherDataService } from '../../../../services/teacherDataService';
 import { UtilService } from '../../../../services/utilService';
 
+class Workgroup {
+  periodId: number;
+  workgroupId: number;
+}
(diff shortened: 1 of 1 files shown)


geoffreykwan (approved):
Looks good.

PR #816, opened 2022-09-16
fix(AT): Display shared units in descending id order

### LINKED ISSUE #801: Shared units show up in incrementing order in AT unit list page
### Description

In the AT unit list page, the "Shared Units" units appear in increasing order (oldest first). This should be the other way around (newest first) to reduce scrolling.

hirokiterashima:
## Test
In the authoring tool's unit listing page
- Shared units appear in descending id order
- If no units are shared, you should see the message "You have no shared units"

Closes #801


geoffreykwan (approved):
Looks good.

PR #823, opened 2022-09-19
feat(Google Login): Add sign up with Google buttons when user not found

### LINKED ISSUE #821: Add sign up with Google links to Google account not found page
## Description
Currently, a user is shown a link to the general WISE registration page when they attempt to log in with a Google account that does not exits in WISE. This causes confusion for some users who want to register with Google because they're presented with the regular sign up form and may not notice Register with Google button below.

We should add links that allow them to directly register with Google.

breity:
## Changes
Adds sign up with Google buttons (student and teacher) to the Google user not found page.

Closes #821.

## Test
- Clicking each button allows user to sign up for WISE using Google.
- On success, new student is redirected to `join/student/form`.
- On success, new teacher is redirected to `join/teacher/form`.


hirokiterashima (approved):
Looks good.

PR #828, opened 2022-09-20
feat(Password): Increase password requirements

### LINKED ISSUE #827: Increase password requirements
### Description

Increase password requirements to require at least 8 characters and at least one lowercase, one uppercase, and one number.

Here are the places where users can change passwords
- student create account
- student edit password
- student forgot password
- teacher create account
- teacher edit password
- teacher forgot password
- teacher change student password in classroom monitor

geoffreykwan:
## Changes

Increased password requirements to require new passwords to be at least 8 characters and to have at least one lowercase, one uppercase, and one number character.

## Test

Make sure the new password requirements are enforced in these locations
- student create account
- student edit password
- student forgot password
- teacher create account
- teacher edit password
- teacher forgot password
- teacher change student password in classroom monitor

Closes #827

### DIFF
--- a/src/app/forgot/student/forgot-student-password-change/forgot-student-password-change.component.html
+++ b/src/app/forgot/student/forgot-student-password-change/forgot-student-password-change.component.html
@@ -16,6 +16,12 @@
                  autofocus
                  required />
           <mat-error *ngIf="changePasswordFormGroup.controls['password'].hasError('required')" i18n>Password required</mat-error>
+          <mat-error *ngIf="changePasswordFormGroup.controls['password'].hasError('minlength')">
+            {{ passwordService.minLengthErrorMessage }}
+          </mat-error>
+          <mat-error *ngIf="changePasswordFormGroup.controls['password'].hasError('pattern')">
+            {{ passwordService.patternErrorMessage }}
+          </mat-error>
         </mat-form-field>
       </p>
       <p>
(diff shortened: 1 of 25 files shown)


hirokiterashima (changes requested):
Looking good.

Unlink Google for both student and teacher needs to be updated as well.

I have some questions:
1. In the template, should we write out the message directly and using i18n tag instead of referencing passwordService? I think it'd made the template easier to read.
```
<mat-error *ngIf="changePasswordFormGroup.controls['password'].hasError('minlength')" i18n>
  Password must be at least 8 characters
</mat-error>
<mat-error *ngIf="changePasswordFormGroup.controls['password'].hasError('pattern')" i18n>
  Password must have at least one lowercase, one uppercase, and one number character
</mat-error>
```
2. There's some duplication, in the code (e.g. code to create FormGroup) and in the template (e.g. mat-error). Is there any way we can extract to a common class?

geoffreykwan:
Ah good catch with the unlink Google page. I forgot about that.

> 1. In the template, should we write out the message directly and using i18n tag instead of referencing passwordService? I think it'd made the template easier to read.

Having the message in the passwordService variable reduces duplication. If you ever want to change the message you only need to change it in one place.

> 2. There's some duplication, in the code (e.g. code to create FormGroup) and in the template (e.g. mat-error). Is there any way we can extract to a common class?

I think we might be able to create a new component that contains the new password and confirm new password fields. Then we could use this new component in all the places we change a password. We should do this in a different issue.

hirokiterashima (approved):

hirokiterashima:
New changes look good. Ready to merge. I created new issue to remove duplication: https://github.com/WISE-Community/WISE-Client/issues/831

Issue #829, opened 2022-09-20
feat(OpenResponse): Add FeedbackRuleEvaluator


hirokiterashima:
### Description

Add support for FeedbackRuleEvaluator to the OpenResponse component to provide feedback for CRater items.

If the component content has cRater.feedback, and cRater.feedback.enabled is true, assume that it will use the FeedbackRuleEvaluator to provide feedback instead of other feedback mechanisms (e.g. MutlipleRoundsFeedback). Also assume that the author can specify one or more feedback texts for a FeedbackRule.
```
...
"cRater": {
  "feedback": {
    "enabled": true,
    "rules": [
      {
        "id": "lq3duxi3se",
        "expression": "ideaCountMoreThan(2)",
        "feedback": [
          "ideaCountMoreThan 2: first feedback",
          "ideaCountMoreThan 2: second feedback,
        ]
      },
      {
        "id": "default",
        "expression": "isDefault",
        "feedback": [
          "The default feedback"
        ]
      },
      ...
    ]
    ...
  }
}
...
```

PR #830, opened 2022-09-20
feat(OpenResponse): Add FeedbackRuleEvaluator expressions

hirokiterashima:
Part 1 of work for issue #829 "feat(OpenResponse): Add FeedbackRuleEvaluator". These are planned future PR's for this issue:
- refactoring, organizing folders to share common classes between OR and DG
- adding FeedbackRule authoring to OR

## Changes
- Evaluate FeedbackRules if authored for the OpenResponse content (see "Test Prep" section below). If the component content has cRater.feedbackRules, assume that it will use the FeedbackRuleEvaluator to provide feedback instead of other feedback mechanisms (e.g. MutlipleRoundsFeedback). Also assume that the author can specify one or more feedback texts for a FeedbackRule.
- Refactoring to share classes between DG and OR
   - Created abstract FeedbackRuleComponent that extends ComponentStudent and use this in DialogGuidanceFeedbackRuleEvaluator

## Test Prep
Using component advanced authoring, directly edit the component JSON and add "feedbackRules" block to OpenResponse's "cRater" block. Make sure that the feedback rules contain an id and a feedback array (instead of feedback string)
```
...
"cRater": {
  "feedback": {
    "enabled": true,
    "rules": [
      {
        "id": "lq3duxi3se",
        "expression": "ideaCountMoreThan(2)",
        "feedback": [
          "ideaCountMoreThan 2: first feedback",
          "ideaCountMoreThan 2: second feedback,
        ]
      },
      {
        "id": "default",
        "expression": "isDefault",
        "feedback": [
          "The default feedback"
        ]
      },
      ...
    ]
  }
  ...
}
...
```

## Test
- FeedbackRule evaluation for OR works and the correct feedback is shown to the student. 
   - If the rule has more than one feedback text, it should cycle through them.
- Existing C-Rater items that do not use FeedbackRule should work as before
- DG works as before

### DIFF
--- a/src/assets/wise5/components/dialogGuidance/DialogGuidanceFeedbackRuleEvaluator.ts
+++ b/src/assets/wise5/components/dialogGuidance/DialogGuidanceFeedbackRuleEvaluator.ts
@@ -1,5 +1,5 @@
+import { FeedbackRuleComponent } from '../feedbackRule/FeedbackRuleComponent';
 import { CRaterResponse } from './CRaterResponse';
-import { DialogGuidanceStudentComponent } from './dialog-guidance-student/dialog-guidance-student.component';
 import { FeedbackRule } from './FeedbackRule';
 import { HasKIScoreTermEvaluator } from './TermEvaluator/HasKIScoreTermEvaluator';
 import { IdeaCountTermEvaluator } from './TermEvaluator/IdeaCountTermEvaluator';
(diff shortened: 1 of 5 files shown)


geoffreykwan (changes requested):
Functionality works well. I have two suggestions.

1. FeedbackRuleComponent is completely abstract so it can actually just be an interface. Then the OpenResponseStudent and DialogGuidanceStudent can implement that interface instead of extend it.
2. We have a boolean field enableMultipleAttemptScoringRules so should we also have a boolean field like enableFeedbackRules? This way authors can easily toggle it on and off without having to completely delete the feedbackRules array.

hirokiterashima:
Good suggestions. I made the changes. 

Instead of enableFeedbackRules boolean, I converted FeedbackRules from Array to Object, and added a "enabled" and "rules" fields inside it. See updated descriptions to change the content.

geoffreykwan (approved):
Looks good.

PR #832, opened 2022-09-22
feat(OpenResponse): Add FeedbackRule authoring

hirokiterashima:
Part 2 of work for issue #829 "feat(OpenResponse): Add FeedbackRuleEvaluator". These are planned future PR's for this issue:
- refactoring, organizing folders to share common classes between OR and DG

## Changes
- Add FeedbackRule authoring to OpenResponse advanced component editing view
- Remove ScoringRule authoring from OpenResponse advanced component editing view
- Remove old translation strings

## Test
In the OpenResponse advanced component editing view:
- Verify that "Enable Feedback Rules" checkbox appears iff CRater is enabled
- Verify that "Scoring Rule" authoring section no longer appears when CRater is enabled
- Check "Enable Feedback Rules" checkbox
   - Verify that "isDefault" feedback is displayed as initial rule
   - Verify that the help button launches the Feedback Rule Authoring help dialog
   - Test that you can add, delete, reorder feedback rules as you can in DialogGuidance
      - Multiple feedback per rule is allowed (you can use the "+" to add more feedback per rule) 
- Un-check "Enable Feedback Rules" checkbox
   - Verify that FeedbackRules is not used 
   - If you enable the checkbox again, the rules that you authored before should appear again

In the OpenResponse student view:
- Test that the Feedback Rules that you authored above work as expected

### DIFF
--- a/src/assets/wise5/components/dialogGuidance/edit-dialog-guidance-feedback-rules/edit-dialog-guidance-feedback-rules.component.ts
+++ b/src/assets/wise5/components/dialogGuidance/edit-dialog-guidance-feedback-rules/edit-dialog-guidance-feedback-rules.component.ts
@@ -17,7 +17,7 @@ export class EditDialogGuidanceFeedbackRulesComponent implements OnInit {
   @Input() feedbackRules: FeedbackRule[] = [];
   inputChanged: Subject<string> = new Subject<string>();
   subscriptions: Subscription = new Subscription();
-  @Input() version: number;
+  @Input() version: number = 2;
 
   constructor(
     private dialog: MatDialog,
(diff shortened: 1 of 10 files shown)


geoffreykwan (commented):
Works well.

If an Open Response previously had the old scoringRules should we show the old authoring for it? It's weird how the old scoringRules still work but you can no longer see it in the authoring or edit it.

hirokiterashima:
Good point. Since we want to move from scoringRules -> FeedbackRules, how about we write a converter to convert existing scoringRules to FeedbackRule format? This converter would be invoked when the author opens the advanced authoring view in the OR component. We can do this in a new PR? For now, authors can edit the scoringRules using JSON editing view if needed.

geoffreykwan:
Yea a converter would be good. We can do that in a new PR. For now I think we should show the old scoringRules authoring if the scoringRules array is not empty and feedbackRules is not enabled. I don't think we should have authors editing JSON.

hirokiterashima:
Added back ScoringRule authoring. Will work on converter and removing the ScoringRule authoring in a future issue.

geoffreykwan:
When the Feedback Rules are enabled the Scoring Rules authoring should then be hidden.

geoffreykwan (approved):
Looks good.

PR #834, opened 2022-09-23
feat(Password): Handle password change error responses

### LINKED ISSUE #833: Handle the new password change error responses
### Description

The API will now send back an error response when there is an error changing a password. It used to send back a success response but with error information like this
```
{
  status: "error",
  messageCode: "incorrectPassword"
}
```

geoffreykwan:
Test with
- https://github.com/WISE-Community/WISE-API/pull/178

## Changes

Password change requests now properly utilize the success function and error function of subscribe(). They used to always use the success function even for errors.

## Test

Make sure success and error responses are properly handled when specifying passwords.

Here are the places where users can specify passwords
- student create account
- student forgot password
- teacher create account
- teacher forgot password
- teacher change student password in classroom monitor
- user edit password
- user unlink Google account

Closes #833

### DIFF
--- a/src/app/forgot/student/forgot-student-password-change/forgot-student-password-change.component.html
+++ b/src/app/forgot/student/forgot-student-password-change/forgot-student-password-change.component.html
@@ -9,17 +9,17 @@
         <mat-form-field appearance="fill" fxFlex>
           <mat-label i18n>Password</mat-label>
           <input matInput
-                 id="password"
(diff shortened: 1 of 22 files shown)


hirokiterashima (approved):
Looks good.

Can you add "private" keyword to private functions and return types to functions (e.g. ForgotTeacherPasswordChangeComponent) before merging?

Issue #840, opened 2022-09-30
Implement Dynamic Prompt with rules


geoffreykwan:
### Description

The Dynamic Prompt should use rules to determine which prompt to display.

PR #841, opened 2022-09-30
feat(VLE): Add dynamic prompt component

geoffreykwan:
# Changes

Added Prompt component and DynamicPrompt component.
Note: This does not contain the rules functionality yet.

# Test

- Manually modify the JSON for a component to have a dynamicPrompt field (see below)
- When the Dynamic Prompt is enabled, it should display the prePrompt and postPrompt
- When the Dynamic Prompt is not enabled, it should show the regular prompt
- Make sure old components that do not have the dynamicPrompt field still work like before

This is what the component content looks like.
```
{
    "id": "7mx88u6y60",
    "type": "OpenResponse",
    "prompt": "This is the regular prompt.",
    "dynamicPrompt": {
        "enabled": true,
        "prePrompt": "This is the pre prompt before the dynamic prompt.",
        "postPrompt": "This is the post prompt after the dynamic prompt."
    }
}
```

Work for #840


hirokiterashima (approved):
Looks good

PR #843, opened 2022-10-05
feat(Dynamic Prompt): Add prompt rules

geoffreykwan:
## Changes

Added functionality to evaluate which dynamic prompt to display.

## Test

- Add a dynamic prompt to a component content (see below example)
- Have the dynamic prompt reference component point to an OpenResponse CRater component (you'll need to change the nodeId and componentId in the example below)
- As a student, submit an answer on the Open Response step and then go to the step with the dynamic prompt and make sure it displays the correct dynamic prompt

```
{
    "id": "gkvyzuils1",
    "type": "PeerChat",
    "prompt": "",
    "showSaveButton": false,
    "showSubmitButton": false,
    "peerGroupingTag": "847zgyqp1p",
    "questionBank": [],
    "dynamicPrompt": {
        "enabled": true,
        "prePrompt": "This is the pre prompt",
        "postPrompt": "This is the post prompt",
        "referenceComponent": {
            "nodeId": "node49",
            "componentId": "2aabvbcukq"
        },
        "rules": [
            {
                "id": "votbik4ncz",
                "expression": "hasKIScore(5)",
                "prompt": "Prompt when you got 5"
            },
            {
                "id": "hwp6sxf8vr",
                "expression": "hasKIScore(4)",
                "prompt": "Prompt when you got 4"
            },
            {
                "id": "swxzc0j7q0",
                "expression": "hasKIScore(3)",
                "prompt": "Prompt when you got 3"
            },
            {
                "id": "ebtgds3b1r",
                "expression": "hasKIScore(2)",
                "prompt": "Prompt when you got 2"
            },
            {
                "id": "z4yu37jcis",
                "expression": "hasKIScore(1)",
                "prompt": "Prompt when you got 1"
            },
            {
                "id": "isDefault",
                "expression": "isDefault",
                "prompt": "Default prompt"
            }
        ]
    }
}
```

#840

### DIFF
--- a/src/assets/wise5/directives/dynamic-prompt/dynamic-prompt.component.ts
+++ b/src/assets/wise5/directives/dynamic-prompt/dynamic-prompt.component.ts
@@ -40,11 +40,15 @@ export class DynamicPromptComponent implements OnInit {
   private evaluateOpenResponseComponent(referenceComponent: any): void {
     const nodeId = this.dynamicPrompt.getReferenceNodeId();
     const componentId = referenceComponent.id;
-    const latestComponentState = this.getLatestComponentState(nodeId, componentId);
-    const latestAutoScoreAnnotation = this.getLatestAutoScore(
+    const latestComponentState = this.studentDataService.getLatestComponentStateByNodeIdAndComponentId(
+      nodeId,
+      componentId
+    );
+    const latestAutoScoreAnnotation = this.annotationService.getLatestScoreAnnotation(
       nodeId,
       componentId,
-      this.configService.getWorkgroupId()
+      this.configService.getWorkgroupId(),
+      'autoScore'
     );
     if (latestComponentState != null && latestAutoScoreAnnotation != null) {
       const cRaterResponse = new CRaterResponse({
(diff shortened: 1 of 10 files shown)


src/assets/wise5/directives/dynamic-prompt/dynamic-prompt.component.ts
```diff
@@ -8,8 +16,82 @@ import { DynamicPrompt } from './DynamicPrompt';
...
+    );
+  }
+
+  private getSubmitCounter(componentState: any): number {
+    return componentState.studentData.submitCounter;
+  }
+
+  private setPrompt(prompt: string): void {
+    this.prompt = prompt;
+  }
```
hirokiterashima: Is this necessary? Let's discuss together when @breity gets back

src/assets/wise5/directives/dynamic-prompt/dynamic-prompt.component.ts
```diff
@@ -8,8 +16,82 @@ import { DynamicPrompt } from './DynamicPrompt';
...
+      this.setPrompt(feedbackRule.prompt);
+    }
+  }
+
+  private getLatestComponentState(nodeId: string, componentId: string): any {
+    return this.studentDataService.getLatestComponentStateByNodeIdAndComponentId(
+      nodeId,
+      componentId
+    );
+  }
```
hirokiterashima: Is this necessary? Let's discuss together when @breity gets back

hirokiterashima (changes requested):
Works well, but if the dynamicPrompt does not have a referenceComponent, it throws this JS error:
```
Cannot read properties of undefined (reading 'nodeId')
    at DynamicPrompt.getReferenceNodeId (DynamicPrompt.ts:22:36)
    at DynamicPromptComponent.getReferenceComponent (dynamic-prompt.component.ts:35:34)
    at DynamicPromptComponent.ngOnInit (dynamic-prompt.component.ts:28:37)
```

You can ignore the inline code comments for now. We can discuss them next week.

geoffreykwan:
Would we ever have a Dynamic Prompt that doesn't have a referenceComponent? Maybe if we have a completely random Dynamic Prompt? Right now in DynamicPrompt.ts, the referenceComponent field is not optional.

hirokiterashima:
Yah that's true. Let's keep the ReferenceComponent mandatory for now. We can relax in the future if needed.

geoffreykwan:
Is it ok that the error is thrown when there is no referenceComponent then since it should always be provided?

hirokiterashima:
Yes, I think that's ok. If the referenceComponent is not provided, we should consider that a user-error. As long as the author is using the authoring tool to edit the dynamic prompt, they shouldn't have this issue.

hirokiterashima (approved):

PR #844, opened 2022-10-05
chore(Feedback Rule): Move submitCounter to CRaterResponse

geoffreykwan:
## Changes

Took the submitCounter out of FeedbackRuleComponent and put it in the CRaterResponse.

## Test

Make sure feedback rules still work properly
- When there is a max submit
  - When the student submits and it is not the second to last and is not the last submit
  - When the student submits and it is the second to last submit
  - When the student submits and it is the last submit
- When there is no max submit

#840


hirokiterashima (approved):
Looks good.

PR #849, opened 2022-10-12
feat(Peer Grouping): Add Different Ideas authoring

geoffreykwan:
## Changes

- Added ability to author Peer Grouping with Different Ideas logic
- Moved EditPeerGroupingDialogComponent from PeerGroupingAuthoringModule to ComponentAuthoringModule otherwise the EditPeerGroupingDialogComponent couldn't use \<edit-connected-component-node-select\> and \<edit-connected-component-component-select\>

## Test

- Edit a Peer Grouping to use Different Ideas logic and set the Step and Component
- Make sure the Different Ideas logic and Step and Component are repopulated if you refresh the page
- Edit a Peer Grouping to use Random or Manual and make sure it still saves correctly like before
- Make sure the correct logic values are saved in the database

#840

### DIFF
--- a/src/assets/wise5/authoringTool/peer-grouping/PeerGroupingLogic.ts
+++ b/src/assets/wise5/authoringTool/peer-grouping/PeerGroupingLogic.ts
@@ -3,7 +3,12 @@ export interface PeerGroupingLogic {
   value: string;
 }
 
+export const DIFFERENT_IDEAS_NAME = $localize`Different Ideas`;
+export const DIFFERENT_IDEAS_REGEX = /differentIdeas\("(\w+)",\s*"(\w+)"\)/g;
+export const DIFFERENT_IDEAS_VALUE = 'differentIdeas';
+
 export const availableLogic: PeerGroupingLogic[] = [
   { name: $localize`Random`, value: 'random' },
-  { name: $localize`Manual`, value: 'manual' }
+  { name: $localize`Manual`, value: 'manual' },
+  { name: DIFFERENT_IDEAS_NAME, value: DIFFERENT_IDEAS_VALUE }
 ];
--- a/src/assets/wise5/authoringTool/peer-grouping/PeerGroupingLogic.ts
+++ b/src/assets/wise5/authoringTool/peer-grouping/PeerGroupingLogic.ts
@@ -7,7 +7,7 @@ export const DIFFERENT_IDEAS_NAME = $localize`Different Ideas`;
 export const DIFFERENT_IDEAS_REGEX = /differentIdeas\("(\w+)",\s*"(\w+)"\)/g;
 export const DIFFERENT_IDEAS_VALUE = 'differentIdeas';
 
-export const availableLogic: PeerGroupingLogic[] = [
+export const AVAILABLE_LOGIC: PeerGroupingLogic[] = [
   { name: $localize`Random`, value: 'random' },
   { name: $localize`Manual`, value: 'manual' },
   { name: DIFFERENT_IDEAS_NAME, value: DIFFERENT_IDEAS_VALUE }
--- a/src/assets/wise5/authoringTool/peer-grouping/edit-peer-grouping-dialog/edit-peer-grouping-dialog.component.ts
+++ b/src/assets/wise5/authoringTool/peer-grouping/edit-peer-grouping-dialog/edit-peer-grouping-dialog.component.ts
@@ -2,8 +2,10 @@ import { Component, Inject } from '@angular/core';
 import { MatDialogRef, MAT_DIALOG_DATA } from '@angular/material/dialog';
 import { PeerGrouping } from '../../../../../app/domain/peerGrouping';
 import { PeerGroupingAuthoringService } from '../../../services/peerGroupingAuthoringService';
+import { ProjectService } from '../../../services/projectService';
 import { UtilService } from '../../../services/utilService';
 import { AuthorPeerGroupingDialogComponent } from '../author-peer-grouping-dialog/author-peer-grouping-dialog.component';
+import { DIFFERENT_IDEAS_REGEX, DIFFERENT_IDEAS_VALUE } from '../PeerGroupingLogic';
 
 @Component({
   selector: 'edit-peer-grouping-dialog',
@@ -11,12 +13,16 @@ import { AuthorPeerGroupingDialogComponent } from '../author-peer-grouping-dialo
   styleUrls: ['./edit-peer-grouping-dialog.component.scss']
 })
 export class EditPeerGroupingDialogComponent extends AuthorPeerGroupingDialogComponent {
+  allowedReferenceComponentTypes: string[] = ['OpenResponse'];
+  logicType: string;
+  referenceComponent: any = {};
   stepsUsedIn: string[] = [];
 
   constructor(
     @Inject(MAT_DIALOG_DATA) public peerGrouping: PeerGrouping,
     protected dialogRef: MatDialogRef<EditPeerGroupingDialogComponent>,
     private peerGroupingAuthoringService: PeerGroupingAuthoringService,
+    private projectService: ProjectService,
     private utilService: UtilService
   ) {
     super(dialogRef);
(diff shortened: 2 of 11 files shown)


src/assets/wise5/authoringTool/peer-grouping/edit-peer-grouping-dialog/edit-peer-grouping-dialog.component.ts
```diff
@@ -2,21 +2,27 @@ import { Component, Inject } from '@angular/core';
...
 
 @Component({
   selector: 'edit-peer-grouping-dialog',
   templateUrl: './edit-peer-grouping-dialog.component.html',
   styleUrls: ['./edit-peer-grouping-dialog.component.scss']
 })
 export class EditPeerGroupingDialogComponent extends AuthorPeerGroupingDialogComponent {
+  allowedReferenceComponentTypes: string[] = ['OpenResponse'];
+  logicType: string;
+  referenceComponent: any = {};
```
hirokiterashima: Can we define a ReferenceComponent type?

src/assets/wise5/authoringTool/peer-grouping/edit-peer-grouping-dialog/edit-peer-grouping-dialog.component.ts
```diff
@@ -25,14 +31,63 @@ export class EditPeerGroupingDialogComponent extends AuthorPeerGroupingDialogCom
...
+      this.referenceComponent = this.getDifferentIdeasReferenceComponent(this.peerGrouping.logic);
+    }
+  }
+
+  private getLogicType(logic: string): string {
+    if (new RegExp(DIFFERENT_IDEAS_REGEX).exec(logic) != null) {
+      return DIFFERENT_IDEAS_VALUE;
+    } else {
+      return logic;
+    }
```
hirokiterashima: Use ternary to reduce complexity?
```suggestion
    return new RegExp(DIFFERENT_IDEAS_REGEX).exec(logic) != null ? DIFFERENT_IDEAS_VALUE : logic;
```

src/assets/wise5/authoringTool/peer-grouping/edit-peer-grouping-dialog/edit-peer-grouping-dialog.component.ts
```diff
@@ -25,14 +31,63 @@ export class EditPeerGroupingDialogComponent extends AuthorPeerGroupingDialogCom
   ngOnInit(): void {
     this.peerGrouping = this.utilService.makeCopyOfJSONObject(this.peerGrouping);
     this.stepsUsedIn = this.peerGroupingAuthoringService.getStepsUsedIn(this.peerGrouping.tag);
+    this.logicType = this.getLogicType(this.peerGrouping.logic);
+    if (this.logicType === DIFFERENT_IDEAS_VALUE) {
```
hirokiterashima, line 36: The variable name DIFFERENT_IDEAS_VALUE is a bit confusing to me here, because what is it a value of? Can it just be "differentIdeas" (string, instead of using variable)? Or how about renaming the variable to DIFFERENT_IDEAS?

src/assets/wise5/authoringTool/peer-grouping/edit-peer-grouping-dialog/edit-peer-grouping-dialog.component.ts
```diff
@@ -25,14 +31,63 @@ export class EditPeerGroupingDialogComponent extends AuthorPeerGroupingDialogCom
...
+
+  private getLogicType(logic: string): string {
+    if (new RegExp(DIFFERENT_IDEAS_REGEX).exec(logic) != null) {
+      return DIFFERENT_IDEAS_VALUE;
+    } else {
+      return logic;
+    }
+  }
+
+  private getDifferentIdeasReferenceComponent(logic: string): any {
```
hirokiterashima: Return ReferenceComponent here instead of any?

src/assets/wise5/authoringTool/peer-grouping/edit-peer-grouping-dialog/edit-peer-grouping-dialog.component.ts
```diff
@@ -25,14 +31,63 @@ export class EditPeerGroupingDialogComponent extends AuthorPeerGroupingDialogCom
...
+    const result = new RegExp(DIFFERENT_IDEAS_REGEX).exec(logic);
+    return {
+      nodeId: result[1],
+      componentId: result[2]
+    };
+  }
+
+  referenceComponentNodeIdChanged(event: any): void {
```
hirokiterashima, line 50: This function is very similar to EditDynamicPromptComponent.referenceComponentNodeIdChanged(). Can we extract to a common parent (abstract?) class?

src/assets/wise5/authoringTool/peer-grouping/edit-peer-grouping-dialog/edit-peer-grouping-dialog.component.ts
```diff
@@ -25,14 +31,63 @@ export class EditPeerGroupingDialogComponent extends AuthorPeerGroupingDialogCom
...
       this.dialogRef.close();
     });
   }
 
+  private injectPeerGroupingLogic(peerGrouping: PeerGrouping): void {
+    if (this.logicType === DIFFERENT_IDEAS_VALUE) {
+      peerGrouping.logic = this.generateDifferentIdeasLogic();
+    } else {
+      peerGrouping.logic = this.logicType;
+    }
```
hirokiterashima: Use ternary to reduce complexity and verbosity? Also generate differentIdeasLogic string in this function? I also wonder if this function should be called updatePeerGroupingLogic() without taking in a parameter..."inject" doesn't feel like the right verb; it's more like we're setting/updating the PeerGrouping.logic field.
```suggestion
    this.peerGrouping.logic =
      this.logicType === DIFFERENT_IDEAS_VALUE
        ? `differentIdeas("${this.referenceComponent.nodeId}", "${this.referenceComponent.componentId}")`
        : this.logicType;
```

src/assets/wise5/authoringTool/peer-grouping/edit-peer-grouping-dialog/edit-peer-grouping-dialog.component.ts
```diff
@@ -25,14 +31,63 @@ export class EditPeerGroupingDialogComponent extends AuthorPeerGroupingDialogCom
...
+  }
+
+  private generateDifferentIdeasLogic(): string {
+    return `${DIFFERENT_IDEAS_VALUE}("${this.referenceComponent.nodeId}", "${this.referenceComponent.componentId}")`;
+  }
+
```
hirokiterashima: Make this string directly in the injectPeerGroupingLogic instead?
```suggestion
```

src/assets/wise5/authoringTool/peer-grouping/select-peer-grouping-option/select-peer-grouping-option.component.ts
```diff
@@ -35,9 +35,13 @@ export class SelectPeerGroupingOptionComponent implements OnInit {
...
 
   setPeerGroupingLogicName(): void {
-    this.peerGroupingLogicName = availableLogic.find(
-      (logic) => logic.value === this.peerGrouping.logic
-    ).name;
+    if (new RegExp(DIFFERENT_IDEAS_REGEX).exec(this.peerGrouping.logic) != null) {
+      this.peerGroupingLogicName = DIFFERENT_IDEAS_NAME;
+    } else {
+      this.peerGroupingLogicName = availableLogic.find(
+        (logic) => this.peerGrouping.logic === logic.value
```
hirokiterashima: Put changing variable on the left side of the comparator.
```suggestion
        (logic) => logic.value === this.peerGrouping.logic;
```

src/assets/wise5/authoringTool/peer-grouping/PeerGroupingLogic.ts
```diff
@@ -3,7 +3,12 @@ export interface PeerGroupingLogic {
   value: string;
 }
 
+export const DIFFERENT_IDEAS_NAME = $localize`Different Ideas`;
+export const DIFFERENT_IDEAS_REGEX = /differentIdeas\("(\w+)",\s*"(\w+)"\)/g;
+export const DIFFERENT_IDEAS_VALUE = 'differentIdeas';
+
 export const availableLogic: PeerGroupingLogic[] = [
```
hirokiterashima: Change to upper-case for consistency?
```suggestion
export const AVAILABLE_LOGIC: PeerGroupingLogic[] = [
```

hirokiterashima (changes requested):
Functionality works great. I added some inline code change suggestions.

src/assets/wise5/authoringTool/peer-grouping/edit-peer-grouping-dialog/edit-peer-grouping-dialog.component.ts
geoffreykwan: The DIFFERENT_IDEAS_NAME and DIFFERENT_IDEAS_VALUE are in the PeerGroupingLogic.ts in reference to the name and values in the AVAILABLE_LOGIC. If we just use DIFFERENT_IDEAS wouldn't that be ambiguous because it could be the name or the value?

src/assets/wise5/authoringTool/peer-grouping/edit-peer-grouping-dialog/edit-peer-grouping-dialog.component.ts
geoffreykwan: We can do this in another issue. Should it be a common parent class or a utility function? I think we should rename \<edit-connected-component-node-select\> and \<edit-connected-component-component-select\> now that we use them in more places besides connected component authoring. Then we can refactor these common functions too.

edit-connected-component.ts => automaticallySetConnectedComponentComponentIdIfPossible()
edit-peer-grouping-dialog.ts => referenceComponentNodeIdChanged()
edit-dynamic-prompt.component.ts => referenceComponentNodeIdChanged()

hirokiterashima: Ok. I think a common abstract parent class would be good. I'll create a new issue for it.

hirokiterashima (approved):
LGTM. Remember to generate messages before merging.

PR #850, opened 2022-10-12
feat(Peer Chat): Show Dynamic Prompt in Grading

geoffreykwan:
## Changes

- Display the Dynamic Prompt in the Peer Chat grading if the latest component state has a Dynamic Prompt in the student data
- Removed PeerChatGradingComponent.ngOnInit() because it was making a duplicate call to retrieve the PeerGroup. The parent component (PeerChatShowWorkComponent) already retrieves the PeerGroup in its ngOnInit().

## Test

- Make sure the Dynamic Prompt shows up in the Peer Chat grading
- Make sure the Peer Chat messages show up in the Peer Chat grading like before

#840


hirokiterashima (changes requested):
Good catch on ngOnInit().

When you make the browser window wide, it displays the prompt on the left of the peer chat. Can you make it so that it's always displayed above?
Screen Shot 2022-10-12 at 3 58 16 PM

Also, remember to generate messages.

hirokiterashima (approved):
LGTM

PR #857, opened 2022-10-14
feat(Dynamic Prompt): Make real request for student data

geoffreykwan:
- Test with https://github.com/WISE-Community/WISE-API/pull/182

## Changes

- Dynamic Prompt now makes a real request to the backend for student data
- We now pass the nodeId to the component-header so that it can be passed down to the prompt and dynamic-prompt

## Test

- Make sure the peer group student data is retrieved from the backend and used to evaluate which prompt to show

#840

### DIFF
--- a/src/assets/wise5/directives/dynamic-prompt/dynamic-prompt.component.spec.ts
+++ b/src/assets/wise5/directives/dynamic-prompt/dynamic-prompt.component.spec.ts
@@ -1,7 +1,7 @@
 import { HttpClientTestingModule } from '@angular/common/http/testing';
 import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { MatDialogModule } from '@angular/material/dialog';
-import { Observable, of } from 'rxjs';
+import { of } from 'rxjs';
 import { StudentTeacherCommonServicesModule } from '../../../../app/student-teacher-common-services.module';
 import { AnnotationService } from '../../services/annotationService';
 import { PeerGroupService } from '../../services/peerGroupService';
@@ -75,6 +75,7 @@ function peerGroupingTagDisabled(): void {
 }
 
 function peerGroupingTagEnabled(): void {
+  let retrieveDynamicPromptStudentDataSpy: jasmine.Spy;
   describe('peerGroupingTagEnabled', () => {
     beforeEach(() => {
       fixture = TestBed.createComponent(DynamicPromptComponent);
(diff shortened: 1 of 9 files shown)


src/assets/wise5/directives/dynamic-prompt/dynamic-prompt.component.spec.ts
```diff
@@ -75,32 +75,46 @@ function peerGroupingTagDisabled(): void {
...
     beforeEach(() => {
       fixture = TestBed.createComponent(DynamicPromptComponent);
       component = fixture.componentInstance;
       component.dynamicPrompt = createDynamicPrompt();
       component.dynamicPrompt.peerGroupingTag = 'apple';
       spyOn(TestBed.inject(PeerGroupService), 'retrievePeerGroup').and.returnValue(
-        createPeerGroup()
+        of(createPeerGroup())
+      );
+      retrieveDynamicPromptStudentDataSpy = spyOn(
```
hirokiterashima: Can we make this variable name shorter, like "studentDataSpy" or just "spy"? There is only one spy in this class, so I don't think it needs to be this specific. It would also make the file shorter.

src/assets/wise5/services/peerGroupService.ts
```diff
@@ -69,4 +69,14 @@ export class PeerGroupService {
   removeWorkgroupFromGroup(workgroupId: number, groupId: number): Observable<any> {
     return this.http.delete(`/api/peer-group/membership/${groupId}/${workgroupId}`);
   }
+
+  retrieveDynamicPromptStudentData(
+    peerGroupId: number,
+    nodeId: string,
+    componentId: string
+  ): Observable<any> {
```
hirokiterashima: Can we define a PeerGroupData and use it instead of any? I think it would help make the code a bit clearer.

interface PeerGroupData {
  annotation: any;
  studentWork: any;
  workgroup: any;
}

hirokiterashima (changes requested):
Functionality works well. I added a few code improvement suggestions inline.

src/assets/wise5/directives/dynamic-prompt/dynamic-prompt.component.spec.ts
geoffreykwan: I think it's better for readers if we name the spy so they don't have to go find the declaration to figure out what it's spying on.

hirokiterashima: I don't think there's anything wrong with going to the declaration to figure out what it's spying on. When I read the code, I first need to read the declaration code to know what it's spying on, no matter what the variable is called. The result is store in the *Spy variable, and from there it becomes a matter of readability.

I think the longer variable name actually makes the code harder to read because 1. reader has to read more letters each time, and 2. in this case, the statement spans multiple lines, which makes the whole class longer and require scrolling. Renaming this to "retrieveDataSpy" or "studentDataSpy" would still get the meaning across and make it easier to read for me.

hirokiterashima (approved):
LGTM

PR #858, opened 2022-10-14
Dynamic Prompt with Rules

### LINKED ISSUE #840: Implement Dynamic Prompt with rules

geoffreykwan:
## Changes

- Created Dynamic Prompt component
- Added Dynamic Prompt to Peer Chat
- Implemented Dynamic Prompt authoring
- Implemented Peer Grouping Different Ideas authoring
- Added Dynamic Prompt to Peer Chat grading view

## Test

Make sure to pull the latest from the WISE-API develop branch.

- Authoring Dynamic Prompt in Peer Chat
- Make sure the correct Dynamic Prompt shows up in the VLE for the student Peer Chat group
- Make sure the Dynamic Prompt shows up in the grading view

Closes #840

### DIFF
--- a/src/app/authoring-tool/edit-dynamic-prompt-rules/edit-dynamic-prompt-rules.component.html
+++ b/src/app/authoring-tool/edit-dynamic-prompt-rules/edit-dynamic-prompt-rules.component.html
@@ -1,0 +1,76 @@
+<div class="feedback-rules notice-bg-bg">
+  <h5 fxLayout="row" fxLayoutAlign="start center" fxLayoutGap="4px">
+    <span i18n>Prompt Rules</span>
+    <button mat-icon-button
+        color="primary"
+        matTooltip="Add a new rule"
+        i18n-matTooltip
+        matTooltipPosition="above"
+        (click)="addNewRule('beginning')">
+      <mat-icon>add_circle</mat-icon>
+    </button>
+    <span fxFlex></span>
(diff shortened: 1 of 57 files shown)


breity (commented):
Functionality seems to be working for me so far. 👍 

A few thoughts on the authoring and grouping:
1. I'm wondering now why we need to set the Grouping Logic for dynamic prompts. It probably makes more sense to use the component grouping, no? I can't think of a case where we would want to use different groupings for the chat and the prompt in the same component. Also, authoring the grouping twice (and maybe 3 times once we add dynamic Question Banks?) seems redundant and confusing. Plus you have to have to author the reference component twice too. (And what about the future where dynamic prompts could be used in any component, not necessarily ones with groupings?)
2. I think the labels for pre- and post-prompt content could be improved. Maybe something like ("Text to show before/after dynamic prompt"?). Or maybe add help icon with tooltip?
3. We should probably update the Help content for authoring dynamic prompts to use correct terminology (e.g. "prompt" instead of "feedback"). [Can be addressed in a different issue.]
4. We should think about cases like when there are 3 teams left (with a group size of 2) and all 3 remaining get grouped together. The 3rd team gets added to group even if it has not completed reference component. (Also, what if another team joins the run later on?) [Can be addressed in a different issue.]

PR #878, opened 2022-10-26
fix(Grading): Fix score and max score not showing up

### LINKED ISSUE #877: Fix grading view score and max score
### Description

- Sometimes the student scores do not show up in the grading view. This occurs when the annotations are finished being retrieved before the project.json is finished being retrieved. This is due to a race condition.
- The max score never shows up in the grading view.

geoffreykwan:
## Changes

- Resolved a race condition by removing the activeAnnotations filtering and instead check if the component exists in the shouldIncludeInTotalScore() function
- Properly named maxScore parameter so that it would show up

## Test

Test score
1. Open a run in the Classroom Monitor and go to the Grade By Step view and click on a step
2. Give a student a score and make sure it saves
3. Refresh the Classroom Monitor a bunch of times and make sure the score shows up every time. Previously the score would not show up in the rare occasion when the annotations finished being retrieved before the project.json was finished being retrieved.

Test max score
1. Open a run in the Classroom Monitor and go to the Grade By Step view and click on a step
2. Give a component a max score and make sure it saves
3. Refresh the Classroom Monitor and make sure the max score shows up. Previously the max score would never show up.

Closes #877

### DIFF
--- a/src/app/services/data.service.ts
+++ b/src/app/services/data.service.ts
@@ -60,12 +60,6 @@ export class DataService {
     }
   }
 
-  getActiveAnnototations(annotations): any[] {
-    return annotations.filter((annotation) => {
-      return this.ProjectService.componentExists(annotation.nodeId, annotation.componentId);
-    });
-  }
-
   broadcastCurrentNodeChanged(previousAndCurrentNode: any) {
     this.currentNodeChangedSource.next(previousAndCurrentNode);
   }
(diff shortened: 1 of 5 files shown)


breity (commented):
Looks good. Score and max score show up reliably now on refresh.

I notice another bug. Changing a workgroup's score doesn't reflect in the score display on the row for the workgroup until you leave the step and come back or refresh (see below). This is a bug on production as well, so we can fix in a separate issue.

Screen Shot 2022-10-31 at 9 50 28 AM

breity (approved):

PR #879, opened 2022-10-27
feat(Data Explorer): Sync table row sorting with connected graphs

### LINKED ISSUE #876: feat(Data Explorer): Sync table and graph sort order
## Description
When student sorts a Table component by column, connected Graph components should update to reflect the sort order.

breity:
## Changes
When students sort a Table component by row, connected graphs now update to reflect the sort order. The sort order is now also saved in the Table's studentData and restored when reloading the component.

The Tabulator library was also updated to the latest version.

Closes #876.

## Test
- Run `npm install` to update dependencies.
- Test with a step that includes a Data Explorer activity (Table component + connected graph component).
- Verify that sorting the table data (by clicking on row headers) updates the connected graph with the same order for the data points.
- Enable row selection on the Table component and test that selecting rows in the table still filters the graph data and that filtering works when sorting the table in different ways.
- Verify that sorting and visible rows are saved with the Table studentData and repopulated when revisiting the step.
- Verify that grading view for Table and Graph components reflect that saved sorting and selected rows.

@geoffreykwan There is also a bug in the connected graph. When the table is sorted, the x-axis labels are properly named in the graph initially. But when you leave the step and come back, they don't show the cell values and are replaced with indices (0, 1, 2, etc.). This also happens in the show work/grading view. (See screenshot below.) Could you take a look to see if you can fix? Thanks!

Screen Shot 2022-10-27 at 12 59 34 PM
Screen Shot 2022-10-27 at 12 58 07 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -58,7 +58,7 @@
         "sockjs-client": "^1.6.0",
         "svg.draggable.js": "2.2.0",
         "svg.js": "2.7.1",
-        "tabulator-tables": "^5.2.7",
+        "tabulator-tables": "^5.4.2",
         "tinymce": "^5.10.3",
         "tslib": "^2.3.1",
         "webfontloader": "^1.6.28",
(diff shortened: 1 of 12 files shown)


breity:
@geoffreykwan I added a fix for the import from Table error.

The problem with the x-axis labels still occurs in the Graph show work/grading view though. In the GraphStudentComponent, there's the `setXAxisLabels` function that gets called when handling connected Data Explorer data. The show work view doesn't have this method and doesn't have access to the Data Explorer data either. So maybe the solution is to store the x-axis labels in the Graph componentState?

geoffreykwan (approved):
Functionality works well. I have a function name change suggestion.

In graph-student.component.ts the getSortedTableData() function actually sorts and filters so the function name is misleading.

Maybe we could rename

getFilteredAndSortedTableData to processTableData
getSortedTableData to (getFilteredAndSortedTableData or getSortedAndFilteredTableData)

PR #884, opened 2022-11-02
chore(Classroom Monitor): Convert Notebook Grading

### LINKED ISSUE #883: Convert Notebook Grading to Angular

geoffreykwan:
## Changes

Converted Notebook Grading to Angular.

## Test

1. Open the Classroom Monitor for a run that has Notes and Report enabled (make sure students have created notes and modified the report)
2. Go to the "Student Notebooks" view
3. Click on a workgroup to expand it
4. Make sure you can tab between the Notes and Final Report
5. Make sure you can collapse a workgroup
6. Make sure Expand All works
7. Make sure Collapse All works
8. Make sure sorting by all the different columns works

Some styling still needs to be fixed.
- There is extra padding on the left and right in the top row that contains the search and expand/collapse all buttons
- There is extra padding on the left and right when you expand a workgroup
- When the browser width is small, the expand and collapse buttons should wrap under the search but it doesn't
- When the browser width is small, the sort arrow wraps under the column name but it shouldn't

Screenshot 2022-11-02 171700

Closes #883


breity (approved):
LGTM.

I updated the styles to fix the issues you mentioned and refactored the css a bit.

The sort arrow still wraps under the column name on very small screens some, but I think it's okay for now, as the arrow would overlap the column name otherwise.

PR #886, opened 2022-11-08
refactor(ProjectService): move UtilService dependency to TeacherProjectService

### LINKED ISSUE #885: refactor(ProjectService): move UtilService dependency to TeacherProjectService
Only TeacherProjectService uses the UtilService dependency. Remove it from the parent ProjectService. Also, lower-case variable names.

hirokiterashima:
## Changes
- Move UtilService from the parent ProjectService to TeacherProjectService, since that's where it's only used
- Make variable names lowercase in the affected files

## Test
- VLE, AT, CM load and work as before

Closes #885

### DIFF
--- a/src/assets/wise5/services/projectService.ts
+++ b/src/assets/wise5/services/projectService.ts
@@ -1,7 +1,6 @@
 'use strict';
 
 import { ConfigService } from './configService';
-import { UtilService } from './utilService';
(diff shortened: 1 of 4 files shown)


geoffreykwan (approved):
Looks good.

It might be good to alphabetize the parameters to the constructor in projectService.ts and teacherProjectService.ts.

PR #893, opened 2022-11-15
refactor(UtilService): Move generateKey() to its own class

### LINKED ISSUE #892: refactor(UtilService): Move generateKey() to its own class
### Description

UtilService.generateKey() is used by select file, but each file needs to import the entire UtilService, which contain other non-related functions. Instead, move this function to its own class (RandomKeyService) and update the calls. This function should also be called statically, since it doesn't require any state. This will improve cohesion of UtilService and simplify the file dependencies.

hirokiterashima:
## Changes
- Move UtilService.generateKey() to RandomKeyService.generate() and make the function static
- Update classes that used the old code to call the new static function, and remove UtilService from constructor if no other code in the class depend on it.

## Test
-  AT (check that a random id or uniqueTag is generated for each of the following actions)
   - add new component: should generate random component id for new component
   - copy component: should generate random component id for copied component
   - Match: add choice/bucket: should generate new ids for choice/bucket
   - MC: add choice: should generate random id for choice
   - add new prompt rule: should generate random id for FeedbackRule
   - Milestone:
      - add milestone satisfy criteria
      - add template satisfy criteria
      - add milestone
   - PeerChat: create new PeerGrouping: should generate new uniqueTag for PeerGrouping
   - animation: add new object should generate random id
- Student VLE:
   - Match: Add choice should generate id for newly added choice
   - Creating new note should set localNotebookItemId
   - Saving events, student work create unique request token

Closes #892


geoffreykwan (approved):
Looks good.

You can remove the unused imports in
- achievementService.ts
- peerChatService.ts

PR #900, opened 2022-11-16
refactor(AuthoringTool): Rename authoringComponentContent to componentContent

### LINKED ISSUE #899: refactor(AuthoringTool): Rename authoringComponentContent to componentContent
### Description

Rename ```authoringComponentContent``` to ```componentContent```. This was a legacy naming. This change will make the variable name consistent with the rest of the application.

Also remove resetUI(). This was used in the old AT where we displayed preview inline with the authoring. Now we show the preview in a popup, so this function and related variables are no longer used.

hirokiterashima:
## Changes
- Rename ```authoringComponentContent``` to ```componentContent```. This was a legacy naming. This change will make the variable name shorter and consistent with the rest of the application.
- Remove ComponentAuthoringComponent.resetUI() and related variables. This was used in the old AT where we displayed preview inline with the authoring. Now we show the preview in a popup, so this function and related variables are no longer used.
- Remove ComponentAuthoringComponent.reloadPreview(). This was called only in the OpenResponse component authoring template, but it is no longer used.

## Test
- AT works as before
   - loading, saving
   - create/delete/copy/move/import steps
   - create/delete/copy/move/import components

Closes #899


geoffreykwan (changes requested):
When performing these actions, the component type labels don't update properly.
- Moving a component
  - If you have two components in a step, say 1. Open Response and 2. Multiple Choice, and you move the Open Response after the Multiple Choice, the view will still show 1. Open Response and 2. Multiple Choice
- Copying a component
- Deleting a component

hirokiterashima:
Good catch. Looks like this is a bug on develop branch. I've created a new issue: #902.

geoffreykwan (approved):
Looks good.

PR #912, opened 2022-11-23
feat(QuestionBank): Add authoring support

### LINKED ISSUE #911: feat(QuestionBank): Add authoring support
### Description

Allow authors to enable/disable, add/delete/edit/reorder QuestionBank items in a PeerChat component

hirokiterashima:
## Changes
- Allow author to add/delete/order/edit QuestionBank items, similar to FeedbackRule and DynamicPrompt. QuestionBank allows author to specify one or more questions per rule, similar to v2 of FeedbackRule
- Hide QuestionBank from student/grading view if it is not disabled

## Test Prep
- If you have any existing PeerChat component, be sure to remove the ```questionsBank``` field from the component content's JSON if it's an empty array. The new format is a JSON object

## Test
- Add/delete/order/edit QuestionBank
- Add/delete QuestionBank question
- Enable/disable QuestionBank should show/hide QuestionBank from student/grading views

Closes #911

### DIFF
--- a/src/app/authoring-tool/edit-question-bank-rules/edit-question-bank-rules.component.html
+++ b/src/app/authoring-tool/edit-question-bank-rules/edit-question-bank-rules.component.html
@@ -1,4 +1,4 @@
-<div class="feedback-rules notice-bg-bg">
+<div class="question-rules notice-bg-bg">
         <h5 fxLayout="row" fxLayoutAlign="start center" fxLayoutGap="4px">
           <span i18n>Question Bank Rules</span>
           <button mat-icon-button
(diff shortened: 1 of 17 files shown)


geoffreykwan (changes requested):
1. Question Bank does not show up for the student. QuestionBankContent object does not contain an "enabled" field but peer-chat-student.component.html looks for that field.

2. Question Bank Rules authoring needs padding or margin because the content is right up against the edge.

3. dynamicPromptChangedEvent variable is used in edit-question-bank.component.html but does not exist in the component. An error is thrown when the component select option is changed. If the component select has multiple options, choosing a component will not be saved. The name of the variable is also related to dynamic prompt but should be related to question bank.

4. Add new lines to these html files for \<peer-chat-question-bank\> element.
peer-chat-grading.component.html
peer-chat-show-work.component.html
peer-chat-student.component.html

hirokiterashima:
For #3, I replaced event emitting with a call to projectService.nodeChanged(), to save changes.

geoffreykwan (approved):
Looks good.

PR #918, opened 2022-11-29
chore(Export): Convert data export

### LINKED ISSUE #913: Convert Data Export Controller

geoffreykwan:
## Changes

Converted Data Export to Angular.

## Test

Make sure all the different exports in the Classroom Monitor still work like they used to.

Closes #913


hirokiterashima (approved):
Everything I tested seems to work.

Before merging, can you make all the variable names in constructor of DataExportComponent (AnnotationService, ConfigService, etc) and all their references (maybe just AbstractDataExportStrategy?) lower-case? Otherwise, looks good.

Also, remember to set the Project and status on the PR and the issue.

PR #925, opened 2022-12-02
chore(Google Login): Remove Angular-Socialx-Login dependency

### LINKED ISSUE #924: Remove angular-socialx-login dependency
Replace with in-house Google OAuth solution. This library has an issue that is not allowing us to upgrade to Angular 14.

hirokiterashima:
## Notes
- @breity please style as you see fit. 
   - The Sign in with Google button no longer is centered. It centers for a second, before being forced to left-aligned.
   - In ```join?googleUserNotFound=true``` page, should "Sorry, your Google User ID was not found in WISE." be a bit more prominent, or does it look ok as is.
   
## Changes
- Replace Angular-Socialx-Login dependency with custom GoogleSignInButtonComponent
- Remove LoginGoogleUserNotFoundComponent with message on top of RegisterHomeComponent, which looks for new ```googleUserNotFound=true``` URL param.
- Code cleanup

## Test Prep
- Test with https://github.com/WISE-Community/WISE-API/pull/196

## Test
- Log in with Google from home page
   - if account exists, log in and show homepage
   - if account doesn't exist, show ```join?googleUserNotFound=true``` page and error 
- Register student/teacher with Google
   - if already registered, show ```/join/googleUserAlreadyExists``` page 
   - if not already registered, correctly populates form with firstname, lastname, email (if teacher)
- Team sign-in form when launching VLE works as before

Closes #924


geoffreykwan (commented):
1. If a student tries to add a teammate that is already in the run, there should be an error message but nothing happens and the "Add Teammate" element disappears. This happens if the teammate you are trying to add is a Google user or regular WISE user. This problem exists on production.

Screenshot 2022-12-02 111348

Screenshot 2022-12-02 111407

2. I think if we change this message
`Hi! There is already a WISE user with this Google account.`
to something like
`Hi! This Google user already has a WISE account.`
it might be more straightforward.

hirokiterashima:
@geoffreykwan I changed the wording of the message. Let's work on the team sign in issue in a separate task since it's also an issue in develop.

geoffreykwan (approved):
Looks good.

breity (commented):
Looks good for the most part. I made some minor style updates.

A few issues:
- ~I don't see the centering problem with the Google buttons. They always show up as full width for me.~
- ~For consistency, I'm thinking we should probably use the same Google button in `/login`.~
- The Google button won't show up for adding a new teammate on the student Team Sign In dialog if one of the other teammates that uses Google hasn't signed in yet. I imagine this is because we're trying to show two Google buttons on the same template. Not sure how we should resolve this. Here's a screenshot:

Screen Shot 2022-12-05 at 12 57 12 PM

If the second teammate signs in or you check "absent" for the teammate that hasn't signed in, you can then click "Add Teammate" and the second Google button will render because the first was removed from the DOM.

Screen Shot 2022-12-05 at 12 57 33 PM

Screen Shot 2022-12-05 at 12 57 55 PM

breity (approved):
Looks good.

I was able to get multiple Google api buttons to render on same page by using a ViewChild in the GoogleSignInButton.

I also updated the Google login and sign up button styles to better match the buttons generated by the Google account api. And I was able to see the centering issue that you both saw and fixed it.

hirokiterashima:
@breity your new changes look good. Thx for coming up with a better solution.

geoffreykwan:
Can we change the button to always just show "Sign in with Google" instead of having it detect the user that's already signed in? It's kind of weird to being asking the teammate to click on a button that says "Sign in as x" when x isn't your name.

Screenshot 2022-12-07 142530

geoffreykwan (approved):
Looks good.

PR #930, opened 2022-12-02
fix(Node Authoring): Component type label does not update

### LINKED ISSUE #902: AT component type label update issue
When performing these actions, the component type labels don't update properly.

- Moving a component
   - If you have two components in a step, say 1. Open Response and 2. Multiple Choice, and you move the Open Response after the Multiple Choice, the view will still show 1. Open Response and 2. Multiple Choice
- Copying a component
- Deleting a component

hirokiterashima:
## Changes
- Component type labels update properly when
   - Moving a component
      - If you have two components in a step, say 1. Open Response and 2. Multiple Choice, and you move the Open Response after the Multiple Choice, the view will still show 1. Open Response and 2. Multiple Choice
   - Copying a component
   - Deleting a component
- Clean up related code
   - removed unnecessary null check 

## Test
- Test above scenarios and make sure the component type labels update properly

Closes #902

### DIFF
--- a/src/assets/wise5/authoringTool/node/nodeAuthoring.html
+++ b/src/assets/wise5/authoringTool/node/nodeAuthoring.html
@@ -146,7 +146,7 @@
       <md-checkbox ng-model='$ctrl.componentsToChecked[component.id]'
           ng-disabled='$ctrl.insertComponentMode'
           class='md-primary'>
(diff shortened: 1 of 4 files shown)


geoffreykwan (approved):
Looks good.

PR #934, opened 2022-12-05
feat(FeedbackRuleEvaluator): Add isSubmitNumber(n) token

### LINKED ISSUE #933: feat(FeedbackRuleEvaluator): Add isSubmitNumber(n) token
### Description

Add a new ```isSubmitNumber(n)``` token in FeedbackRuleEvaluator that evaluates to true when this is the user's n'th submit.

hirokiterashima:
## Changes
- Add support for ```isSubmitNumber(n)``` token in FeedbackRuleEvaluator that evaluates to true when it is the student's n-th submit.
- Add explanation for new token in AT FeedbackRule help dialog
- Refactor: extract TermEvaluatorFactory to create appropriate TermEvaluator

## Test
- ```isSubmitNumber(n)``` works
   - by itself
   - in conjunction with another token, e.g. ```isSubmitNumber(2) && ideaCountLessThan(3)```
- Help dialog in AT shows info about the new token

Closes #933

### DIFF
--- a/src/assets/wise5/components/common/feedbackRule/TermEvaluator/IsSubmitNumberEvaluator.spec.ts
+++ b/src/assets/wise5/components/common/feedbackRule/TermEvaluator/IsSubmitNumberEvaluator.spec.ts
@@ -9,11 +9,6 @@ describe('IsSubmitNumberEvaluator', () => {
       const evaluator_2 = new IsSubmitNumberEvaluator('isSubmitNumber(2)');
       expect(evaluator_2.evaluate(response_2nd_submit)).toBeTrue();
       expect(evaluator_2.evaluate(response_3rd_submit)).toBeFalse();
-
-      const evaluator_10 = new IsSubmitNumberEvaluator('isSubmitNumber(10)');
-      expect(evaluator_10.evaluate(response_2nd_submit)).toBeFalse();
-      const response_10th_submit = createResponseWithSubmitNumber(10);
-      expect(evaluator_10.evaluate(response_10th_submit)).toBeTrue();
     });
   });
 });
--- a/src/assets/wise5/components/common/feedbackRule/TermEvaluator/IsSubmitNumberEvaluator.spec.ts
+++ b/src/assets/wise5/components/common/feedbackRule/TermEvaluator/IsSubmitNumberEvaluator.spec.ts
@@ -1,0 +1,25 @@
+import { CRaterResponse } from '../../cRater/CRaterResponse';
+import { IsSubmitNumberEvaluator } from './IsSubmitNumberEvaluator';
+
+describe('IsSubmitNumberEvaluator', () => {
+  describe('evaluate()', () => {
+    it("should return true iff response's submitCounter is the expected submit number", () => {
+      const response_2nd_submit = createResponseWithSubmitNumber(2);
+      const response_3rd_submit = createResponseWithSubmitNumber(3);
+      const evaluator_2 = new IsSubmitNumberEvaluator('isSubmitNumber(2)');
+      expect(evaluator_2.evaluate(response_2nd_submit)).toBeTrue();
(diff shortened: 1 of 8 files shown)


src/assets/wise5/components/common/feedbackRule/TermEvaluator/IsSubmitNumberEvaluator.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+      const response_3rd_submit = createResponseWithSubmitNumber(3);
+      const evaluator_2 = new IsSubmitNumberEvaluator('isSubmitNumber(2)');
+      expect(evaluator_2.evaluate(response_2nd_submit)).toBeTrue();
+      expect(evaluator_2.evaluate(response_3rd_submit)).toBeFalse();
+
+      const evaluator_10 = new IsSubmitNumberEvaluator('isSubmitNumber(10)');
+      expect(evaluator_10.evaluate(response_2nd_submit)).toBeFalse();
+      const response_10th_submit = createResponseWithSubmitNumber(10);
+      expect(evaluator_10.evaluate(response_10th_submit)).toBeTrue();
+    });
```
geoffreykwan: Should this be two tests?

src/assets/wise5/components/common/feedbackRule/feedback-rule-help/feedback-rule-help.component.html
```diff
@@ -46,6 +46,13 @@ <h3 i18n>Terms</h3>
         Ex: <b>ideaCountEquals(2)</b> evaluates to true if the student had exactly 2 ideas.
       </td>
     </tr>
+    <tr>
+      <td class="term">isSubmitNumber(X)</td>
+      <td i18n>
+        Evaluates to true if this is the student’s X-th attempt.<br />
```
geoffreykwan: I noticed that a lot of the apostrophes in this file are fancy apostrophes. Should we change all of them to plain apostrophes?
Fancy `’`
Plain `'`

geoffreykwan (commented):
Functionality works well.

src/assets/wise5/components/common/feedbackRule/TermEvaluator/IsSubmitNumberEvaluator.spec.ts
hirokiterashima: Do you mean separate into
```"should return true iff response's submitCounter is the expected submit number (single-digit)"```
and
```"should return true iff response's submitCounter is the expected submit number (double-digit)"```
?

I thought about that, but wasn't sure, so I put it into one and separated with a newline. In both cases, it's testing the same block of code in ```evaluate()```- the code doesn't handle single and double digit inputs differently, and they return the same output. I think it makes sense to have two tests if the input tests different parts of the code, and you expect different behavior based on the input.

What do you think?

src/assets/wise5/components/common/feedbackRule/feedback-rule-help/feedback-rule-help.component.html
hirokiterashima: Good catch. Must have been from when I did copy-paste from Word.

src/assets/wise5/components/common/feedbackRule/TermEvaluator/IsSubmitNumberEvaluator.spec.ts
geoffreykwan: If it's testing the same block of code then we don't need the evaluator_10 tests then?

hirokiterashima: You're right. It's not necessary.

breity (approved):
Looks good.

I made a minor update to the token explanation in the help dialog (replacing "submit" with "response" to match the other explanations).

geoffreykwan (approved):
Looks good.

PR #938, opened 2022-12-06
feat(Export): Implement Label component export

### LINKED ISSUE #923: Implement Label component export
Have each row represent a workgroup and display a column for each label the student has.

geoffreykwan:
## Changes

Implemented Label component export.

## Test

Make sure the Label component export shows the labels the students saved.
1. Open the Classroom Monitor
2. Go to the "Data Export" view
4. Find a Label component and click "All" and "Latest"

Also make sure changing the settings for "Include Student Names" and "Include Only Submits" are properly reflected in the export file.

Closes #923

### DIFF
--- a/src/assets/wise5/classroomMonitor/dataExport/ComponentRevisionCounter.ts
+++ b/src/assets/wise5/classroomMonitor/dataExport/ComponentRevisionCounter.ts
@@ -1,8 +1,17 @@
 export class ComponentRevisionCounter extends Map<string, number> {
-  incrementCounter(key: string): void {
+  incrementCounter(nodeId: string, componentId: string): void {
+    const key = `${nodeId}_${componentId}`;
     if (!this.has(key)) {
       this.set(key, 1);
     }
     this.set(key, this.get(key) + 1);
+  }
+
+  getCounter(nodeId: string, componentId: string): number {
+    const key = `${nodeId}_${componentId}`;
+    if (!this.has(key)) {
+      this.set(key, 1);
+    }
+    return this.get(key);
   }
 }
--- a/src/assets/wise5/classroomMonitor/dataExport/ComponentRevisionCounter.ts
+++ b/src/assets/wise5/classroomMonitor/dataExport/ComponentRevisionCounter.ts
@@ -1,5 +1,8 @@
 export class ComponentRevisionCounter extends Map<string, number> {
   incrementCounter(key: string): void {
+    if (!this.has(key)) {
+      this.set(key, 1);
+    }
     this.set(key, this.get(key) + 1);
   }
 }
--- a/src/assets/wise5/classroomMonitor/dataExport/ComponentRevisionCounter.ts
+++ b/src/assets/wise5/classroomMonitor/dataExport/ComponentRevisionCounter.ts
@@ -1,17 +1,23 @@
 export class ComponentRevisionCounter extends Map<string, number> {
   incrementCounter(nodeId: string, componentId: string): void {
-    const key = `${nodeId}_${componentId}`;
-    if (!this.has(key)) {
-      this.set(key, 1);
-    }
+    const key = this.getKey(nodeId, componentId);
+    this.initializeKeyIfNecessary(key);
     this.set(key, this.get(key) + 1);
   }
 
   getCounter(nodeId: string, componentId: string): number {
-    const key = `${nodeId}_${componentId}`;
+    const key = this.getKey(nodeId, componentId);
+    this.initializeKeyIfNecessary(key);
+    return this.get(key);
+  }
+
+  private getKey(nodeId: string, componentId: string): string {
+    return `${nodeId}_${componentId}`;
+  }
+
+  private initializeKeyIfNecessary(key: string): void {
     if (!this.has(key)) {
       this.set(key, 1);
     }
-    return this.get(key);
   }
 }
--- a/src/assets/wise5/classroomMonitor/dataExport/ComponentRevisionCounter.ts
+++ b/src/assets/wise5/classroomMonitor/dataExport/ComponentRevisionCounter.ts
@@ -1,0 +1,5 @@
+export class ComponentRevisionCounter extends Map<string, number> {
+  incrementCounter(key: string): void {
+    this.set(key, this.get(key) + 1);
+  }
+}
--- a/src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
+++ b/src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
@@ -1,6 +1,7 @@
 import { Component } from '../../../common/Component';
 import { ComponentDataExportParams } from '../ComponentDataExportParams';
 import { ComponentRevisionCounter } from '../ComponentRevisionCounter';
+import { UserIdsAndStudentNames } from '../UserIdsAndStudentNames';
 import { AbstractDataExportStrategy } from './AbstractDataExportStrategy';
 
 export abstract class AbstractComponentDataExportStrategy extends AbstractDataExportStrategy {
@@ -33,22 +34,11 @@ export abstract class AbstractComponentDataExportStrategy extends AbstractDataEx
     componentState: any
   ): string[];
 
-  generateComponentHeaderRow(
-    columnNames: string[],
-    columnNameToNumber: Map<string, number>
-  ): string[] {
-    this.populateColumnNames(columnNames, columnNameToNumber);
-    const headerRow = [];
-    for (const columnName of columnNames) {
-      headerRow.push(columnName);
-    }
-    return headerRow;
+  generateComponentHeaderRow(columnNames: string[]): string[] {
+    return [...columnNames];
   }
 
-  private populateColumnNames(
-    columnNames: string[],
-    columnNameToNumber: Map<string, number>
-  ): void {
+  populateColumnNames(columnNames: string[], columnNameToNumber: Map<string, number>): void {
     const defaultColumnNames = [
       '#',
       'Workgroup ID',
@@ -108,7 +98,10 @@ export abstract class AbstractComponentDataExportStrategy extends AbstractDataEx
   ): string[] {
     const rows = [];
     const userInfo = this.configService.getUserInfoByWorkgroupId(workgroupId);
-    const extractedUserIDsAndStudentNames = this.extractUserIDsAndStudentNames(userInfo.users);
+    const extractedUserIdsAndStudentNames = new UserIdsAndStudentNames(
+      userInfo.users,
+      this.canViewStudentNames
+    );
     const componentRevisionCounter = new ComponentRevisionCounter();
     const componentStates = this.teacherDataService.getComponentStatesByWorkgroupIdAndComponentId(
       workgroupId,
@@ -120,12 +113,12 @@ export abstract class AbstractComponentDataExportStrategy extends AbstractDataEx
         const row = this.generateComponentWorkRow(
           rowCounter,
           workgroupId,
-          extractedUserIDsAndStudentNames['userId1'],
-          extractedUserIDsAndStudentNames['userId2'],
-          extractedUserIDsAndStudentNames['userId3'],
-          extractedUserIDsAndStudentNames['studentName1'],
-          extractedUserIDsAndStudentNames['studentName2'],
-          extractedUserIDsAndStudentNames['studentName3'],
+          extractedUserIdsAndStudentNames.getUserId(1),
+          extractedUserIdsAndStudentNames.getUserId(2),
+          extractedUserIdsAndStudentNames.getUserId(3),
+          extractedUserIdsAndStudentNames.getStudentName(1),
+          extractedUserIdsAndStudentNames.getStudentName(2),
+          extractedUserIdsAndStudentNames.getStudentName(3),
           userInfo.periodName,
           componentRevisionCounter,
           componentState
@@ -133,28 +126,10 @@ export abstract class AbstractComponentDataExportStrategy extends AbstractDataEx
         rows.push(row);
         rowCounter++;
       } else {
-        this.incrementRevisionCounter(componentRevisionCounter, nodeId, componentId);
+        componentRevisionCounter.incrementCounter(nodeId, componentId);
       }
     }
     return rows;
-  }
-
-  /**
-   * @param users An array of user objects. Each user object contains an id and name.
-   * @returns {object} An object that contains key/value pairs. The key is userIdX
-   * or studentNameX where X is an integer. The values are the corresponding actual
-   * values of user id and student name.
-   */
-  private extractUserIDsAndStudentNames(users: any[]): any {
-    const extractedUserIDsAndStudentNames = {};
-    for (let u = 0; u < users.length; u++) {
-      const user = users[u];
-      extractedUserIDsAndStudentNames['userId' + (u + 1)] = user.id;
-      if (this.canViewStudentNames) {
-        extractedUserIDsAndStudentNames['studentName' + (u + 1)] = user.name;
-      }
-    }
-    return extractedUserIDsAndStudentNames;
   }
 
   private shouldExportRow(
@@ -180,14 +155,6 @@ export abstract class AbstractComponentDataExportStrategy extends AbstractDataEx
       this.workSelectionType === 'exportLatestWork' &&
       componentStateIndex !== numComponentStates - 1
     );
-  }
-
-  private incrementRevisionCounter(
-    componentRevisionCounter: ComponentRevisionCounter,
-    nodeId: string,
-    componentId: string
-  ): void {
-    componentRevisionCounter.incrementCounter(`${nodeId}_${componentId}`);
   }
 
   /**
@@ -233,16 +200,13 @@ export abstract class AbstractComponentDataExportStrategy extends AbstractDataEx
     row[this.columnNameToNumber.get('Project Name')] = this.projectService.getProjectTitle();
     row[this.columnNameToNumber.get('Run ID')] = this.configService.getRunId();
     row[this.columnNameToNumber.get('Student Work ID')] = componentState.id;
-    const formattedDateTime = this.utilService.convertMillisecondsToFormattedDateTime(
-      componentState.serverSaveTime
-    );
-    row[this.columnNameToNumber.get('Server Timestamp')] = formattedDateTime;
+    row[
+      this.columnNameToNumber.get('Server Timestamp')
+    ] = this.utilService.convertMillisecondsToFormattedDateTime(componentState.serverSaveTime);
     const clientSaveTime = new Date(componentState.clientSaveTime);
-    if (clientSaveTime != null) {
-      const clientSaveTimeString =
-        clientSaveTime.toDateString() + ' ' + clientSaveTime.toLocaleTimeString();
-      row[this.columnNameToNumber.get('Client Timestamp')] = clientSaveTimeString;
-    }
+    const clientSaveTimeString =
+      clientSaveTime.toDateString() + ' ' + clientSaveTime.toLocaleTimeString();
+    row[this.columnNameToNumber.get('Client Timestamp')] = clientSaveTimeString;
     row[this.columnNameToNumber.get('Node ID')] = componentState.nodeId;
     row[this.columnNameToNumber.get('Component ID')] = componentState.componentId;
     row[this.columnNameToNumber.get('Step Title')] = this.projectService.getNodePositionAndTitle(
@@ -276,11 +240,9 @@ export abstract class AbstractComponentDataExportStrategy extends AbstractDataEx
        * mapping. this case will happen when we are exporting all student
        * work.
        */
-      row[this.columnNameToNumber.get('Component Revision Counter')] = this.getRevisionCounter(
-        componentRevisionCounter,
-        componentState.nodeId,
-        componentState.componentId
-      );
+      row[
+        this.columnNameToNumber.get('Component Revision Counter')
+      ] = componentRevisionCounter.getCounter(componentState.nodeId, componentState.componentId);
     } else {
       /*
        * use the revision counter from the value in the component state.
@@ -291,11 +253,7 @@ export abstract class AbstractComponentDataExportStrategy extends AbstractDataEx
       row[this.columnNameToNumber.get('Component Revision Counter')] =
         componentState.revisionCounter;
     }
-    this.incrementRevisionCounter(
-      componentRevisionCounter,
-      componentState.nodeId,
-      componentState.componentId
-    );
+    componentRevisionCounter.incrementCounter(componentState.nodeId, componentState.componentId);
     if (componentState.isSubmit) {
       row[this.columnNameToNumber.get('Is Submit')] = 1;
       if (studentData != null) {
@@ -337,25 +295,6 @@ export abstract class AbstractComponentDataExportStrategy extends AbstractDataEx
     if (studentName3 != null && this.includeStudentNames) {
       row[this.columnNameToNumber.get('Student Name 3')] = studentName3;
     }
-  }
-
-  /**
-   * Get the revision number for the next component state revision.
-   * @param componentRevisionCounter The mapping from component to revision
-   * counter.
-   * @param nodeId The node id the component is in.
-   * @param componentId The component id of the component.
-   */
-  private getRevisionCounter(
-    componentRevisionCounter: ComponentRevisionCounter,
-    nodeId: string,
-    componentId: string
-  ): number {
-    const nodeIdAndComponentId = `${nodeId}_${componentId}`;
-    if (!componentRevisionCounter.has(nodeIdAndComponentId)) {
-      componentRevisionCounter.set(nodeIdAndComponentId, 1);
-    }
-    return componentRevisionCounter.get(nodeIdAndComponentId);
   }
 
   generateExportFileName(nodeId: string, componentId: string, componentType: string): string {
--- a/src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
+++ b/src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
@@ -1,3 +1,4 @@
+import { ComponentDataExportParams } from '../ComponentDataExportParams';
 import { AbstractDataExportStrategy } from './AbstractDataExportStrategy';
 
 export abstract class AbstractComponentDataExportStrategy extends AbstractDataExportStrategy {
@@ -6,7 +7,11 @@ export abstract class AbstractComponentDataExportStrategy extends AbstractDataEx
   includeStudentNames: boolean;
   workSelectionType: string;
 
-  constructor(protected nodeId: string, protected component: any, additionalParams: any) {
+  constructor(
+    protected nodeId: string,
+    protected component: any,
+    additionalParams: ComponentDataExportParams
+  ) {
     super();
     this.canViewStudentNames = additionalParams.canViewStudentNames;
     this.includeOnlySubmits = additionalParams.includeOnlySubmits;
@@ -15,7 +20,6 @@ export abstract class AbstractComponentDataExportStrategy extends AbstractDataEx
   }
 
   abstract generateComponentWorkRow(
-    component: any,
     columnNames: string[],
     columnNameToNumber: any,
     rowCounter: number,
@@ -81,7 +85,7 @@ export abstract class AbstractComponentDataExportStrategy extends AbstractDataEx
     columnNames: string[],
     columnNameToNumber: any,
     nodeId: string
-  ): any[] {
+  ): string[] {
     const componentId = component.id;
     let rows = [];
     let rowCounter = 1;
@@ -125,7 +129,6 @@ export abstract class AbstractComponentDataExportStrategy extends AbstractDataEx
       const componentState = componentStates[c];
       if (this.shouldExportRow(componentState, c, componentStates.length)) {
         const row = this.generateComponentWorkRow(
-          component,
           columnNames,
           columnNameToNumber,
           rowCounter,
@@ -174,16 +177,24 @@ export abstract class AbstractComponentDataExportStrategy extends AbstractDataEx
     componentStateIndex: number,
     numComponentStates: number
   ): boolean {
-    let exportRow = true;
-    if (this.includeOnlySubmits && !componentState.isSubmit) {
-      exportRow = false;
-    } else if (
+    return !(
+      this.includeOnlySubmitsAndIsNotSubmit(componentState) ||
+      this.exportLatestWorkAndIsNotLatestWork(componentStateIndex, numComponentStates)
+    );
+  }
+
+  private includeOnlySubmitsAndIsNotSubmit(componentState: any): boolean {
+    return this.includeOnlySubmits && !componentState.isSubmit;
+  }
+
+  private exportLatestWorkAndIsNotLatestWork(
+    componentStateIndex: number,
+    numComponentStates: number
+  ): boolean {
+    return (
       this.workSelectionType === 'exportLatestWork' &&
-      componentStateIndex != numComponentStates - 1
-    ) {
-      exportRow = false;
-    }
-    return exportRow;
+      componentStateIndex !== numComponentStates - 1
+    );
   }
 
   private incrementRevisionCounter(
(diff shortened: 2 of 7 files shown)


src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,393 @@
...
+  constructor(protected nodeId: string, protected component: any, additionalParams: any) {
+    super();
+    this.canViewStudentNames = additionalParams.canViewStudentNames;
+    this.includeOnlySubmits = additionalParams.includeOnlySubmits;
+    this.includeStudentNames = additionalParams.includeStudentNames;
+    this.workSelectionType = additionalParams.workSelectionType;
+  }
+
+  abstract generateComponentWorkRow(
+    component: any,
```
hirokiterashima: Not used, remove?
```suggestion
```

src/assets/wise5/classroomMonitor/dataExport/strategies/LabelComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,78 @@
...
+      );
+      this.addLabelHeaders(rows, this.maxNumLabels);
+      const fileName = this.generateExportFileName(this.nodeId, this.component.id, 'label');
+      this.controller.generateCSVFile(rows, fileName);
+      this.controller.hideDownloadingExportMessage();
+    });
+  }
+
+  generateComponentWorkRow(
+    component: any,
```
hirokiterashima: Not used, remove?
```suggestion
```

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,393 @@
+import { AbstractDataExportStrategy } from './AbstractDataExportStrategy';
+
+export abstract class AbstractComponentDataExportStrategy extends AbstractDataExportStrategy {
+  canViewStudentNames: boolean;
+  includeOnlySubmits: boolean;
+  includeStudentNames: boolean;
+  workSelectionType: string;
+
+  constructor(protected nodeId: string, protected component: any, additionalParams: any) {
```
hirokiterashima: Create new ComponentDataExportParams interface to type ```additionalParams```?
Type ComponentContent?
```suggestion
  constructor(protected nodeId: string, protected component: ComponentContent, additionalParams: ComponentDataExportParams) {
```

src/assets/wise5/classroomMonitor/dataExport/strategies/LabelComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,78 @@
+import { AbstractComponentDataExportStrategy } from './AbstractComponentDataExportStrategy';
+
+export class LabelComponentDataExportStrategy extends AbstractComponentDataExportStrategy {
+  maxNumLabels: number;
+
+  constructor(nodeId: string, component: any, additionalParams: any) {
```
hirokiterashima: Create new ComponentDataExportParams interface to type additionalParams?
Type ComponentContent?
```suggestion
  constructor(nodeId: string, component: ComponentContent, additionalParams: ComponentDataExportParams) {
```

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,393 @@
...
+    this.canViewStudentNames = additionalParams.canViewStudentNames;
+    this.includeOnlySubmits = additionalParams.includeOnlySubmits;
+    this.includeStudentNames = additionalParams.includeStudentNames;
+    this.workSelectionType = additionalParams.workSelectionType;
+  }
+
+  abstract generateComponentWorkRow(
+    component: any,
+    columnNames: string[],
+    columnNameToNumber: any,
```
hirokiterashima: Define new ColumnNameToNumber interface?
```suggestion
    columnNameToNumber: ColumnNameToNumber,
```

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,393 @@
...
+    rowCounter: number,
+    workgroupId: number,
+    userId1: number,
+    userId2: number,
+    userId3: number,
+    studentName1: string,
+    studentName2: string,
+    studentName3: string,
+    periodName: string,
+    componentRevisionCounter: any,
```
hirokiterashima: Define new ComponentRevisionCounter interface?
```suggestion
    componentRevisionCounter: ComponentRevisionCounter,
```

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,393 @@
...
+      columnNames.push(defaultColumnName);
+    }
+  }
+
+  generateComponentWorkRows(
+    component: any,
+    columnNames: string[],
+    columnNameToNumber: any,
+    nodeId: string
+  ): any[] {
```
hirokiterashima: Looks like this returns a string array?
```suggestion
  ): string[] {
```

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,393 @@
...
+    let exportRow = true;
+    if (this.includeOnlySubmits && !componentState.isSubmit) {
+      exportRow = false;
+    } else if (
+      this.workSelectionType === 'exportLatestWork' &&
+      componentStateIndex != numComponentStates - 1
+    ) {
+      exportRow = false;
+    }
+    return exportRow;
```
hirokiterashima: This block is a bit hard to read. Can this be re-written without the exportRow variable, and combine the two false cases?
```
if ((this.includeOnlySubmits && !componentState.isSubmit) ||
    (this.workSelectionType === 'exportLatestWork' && componentStateIndex != numComponentStates - 1)) {
  return false;
}
return true;
```

It would be even better if we could put the ```this.includeOnlySubmits && !componentState.isSubmit``` and ```this.workSelectionType === 'exportLatestWork' && componentStateIndex != numComponentStates - 1``` blocks into their own function with a descriptive function name? Then this function would be one line:
```return !(function1() || function2())```

File: src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts [hunk 9]
```diff
@@ -0,0 +1,393 @@
...
+      componentState.nodeId
+    );
+    const componentPartNumber =
+      this.projectService.getComponentPosition(componentState.nodeId, componentState.componentId) +
+      1;
+    row[columnNameToNumber['Component Part Number']] = componentPartNumber;
+    const component = this.projectService.getComponent(
+      componentState.nodeId,
+      componentState.componentId
+    );
```
hirokiterashima: Is this component the same as ```this.component```?
```suggestion
    const component = this.component;
```

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,393 @@
...
+    if (component != null) {
```
hirokiterashima: Will this ever be null?
```suggestion
```

hirokiterashima (changes requested):
Export works well.

Can you define and assign types to field names, method params and return types in the new classes where possible? I think it would clarify the code and make it easier to read than ```any```. I commented on some but not all instances of ```any``` that I think can be changed.

Also, is the component param to ```generateComponentWorkRow()``` not needed since you can get the component via ```this.component```?

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
geoffreykwan: What would the ColumnNameToNumber interface look like? columnNameToNumber is an object where the key is a column name string and the value is a number.

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
geoffreykwan: What would the ComponentRevisionCounter interface look like? componentRevisionCounter is an object where the key is a component id string and the value is a number.

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,398 @@
...
+          extractedUserIDsAndStudentNames['studentName3'],
+          userInfo.periodName,
+          componentRevisionCounter,
+          componentState
+        );
+        rows.push(row);
+        rowCounter++;
+      } else {
+        // We do not want to add this component state as a row in the export but we still want to
+        // increment the revision counter.
```
hirokiterashima: The comment is not useful (just repeats what the code does and does not offer additional explanation why the code was written), remove?
```suggestion
```

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,398 @@
...
+    );
+  }
+
+  private incrementRevisionCounter(
+    componentRevisionCounter: ComponentRevisionCounter,
+    nodeId: string,
+    componentId: string
+  ): void {
+    componentRevisionCounter.incrementCounter(`${nodeId}_${componentId}`);
+  }
```
hirokiterashima: Directly call ```componentRevisionCounter.incrementCounter(`${nodeId}_${componentId}`);```
wherever this function is called?
```suggestion
 ```

File: src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts [hunk 13]
```diff
@@ -0,0 +1,398 @@
...
+    nodeId: string,
+    componentId: string,
+    rowCounter: number
+  ): string[] {
+    const rows = [];
+    const userInfo = this.configService.getUserInfoByWorkgroupId(workgroupId);
+    const extractedUserIDsAndStudentNames = this.extractUserIDsAndStudentNames(userInfo.users);
+
+    // A mapping from component to component revision counter. The key will be
+    // {{nodeId}}_{{componentId}} and the value will be a number.
```
hirokiterashima: Move this comment to class definition or remove altogether? Class file indicates that it's a Map<string, number> already.
```suggestion
```

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,398 @@
...
+      componentState.nodeId,
+      componentState.componentId
+    );
+    if (componentState.revisionCounter == null) {
+      /*
+       * use the revision counter obtained from the componentRevisionCounter
+       * mapping. this case will happen when we are exporting all student
+       * work.
+       */
+      row[columnNameToNumber.get('Component Revision Counter')] = revisionCounter;
```
hirokiterashima: ```revisionCounter``` is only used in the ```if (componentState.revisionCounter == null)``` block, so move it there?
```suggestion
    if (componentState.revisionCounter == null) {
      /*
       * use the revision counter obtained from the componentRevisionCounter
       * mapping. this case will happen when we are exporting all student
       * work.
       */
      row[columnNameToNumber.get('Component Revision Counter')] =  this.getRevisionCounter(
      componentRevisionCounter,
      componentState.nodeId,
      componentState.componentId
    );
```

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,398 @@
...
+       */
+      row[columnNameToNumber.get('Component Revision Counter')] = componentState.revisionCounter;
+    }
+    this.incrementRevisionCounter(
+      componentRevisionCounter,
+      componentState.nodeId,
+      componentState.componentId
+    );
+    const isSubmit = componentState.isSubmit;
+    if (isSubmit) {
```
hirokiterashima: No need to create isSubmit variable here, since it's only being used once?
```suggestion
    if (componentState.isSubmit) {
```

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,398 @@
...
+      row[columnNameToNumber.get('Server Timestamp')] = formattedDateTime;
+    }
+    if (componentState.clientSaveTime != null) {
+      const clientSaveTime = new Date(componentState.clientSaveTime);
+      if (clientSaveTime != null) {
+        const clientSaveTimeString =
+          clientSaveTime.toDateString() + ' ' + clientSaveTime.toLocaleTimeString();
+        row[columnNameToNumber.get('Client Timestamp')] = clientSaveTimeString;
+      }
+    }
```
hirokiterashima: Remove null-check? clientSaveTime and serverSaveTime are never null. I checked in prod db too.

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,398 @@
...
+
+export abstract class AbstractComponentDataExportStrategy extends AbstractDataExportStrategy {
+  canViewStudentNames: boolean;
+  includeOnlySubmits: boolean;
+  includeStudentNames: boolean;
+  workSelectionType: string;
+
+  constructor(
+    protected nodeId: string,
+    protected component: any,
```
hirokiterashima: Replace with Component (WISE component), which includes nodeId and componentContent?
```suggestion
    protected component: Component,
```

hirokiterashima (changes requested):
The code is getting easier to read. I added additional code change suggestions inline based on the new changes.

Also, what do you think about making ```columnNames``` and ```columnNameToNumber``` into private fields of ```AbstractComponentDataExportStrategy```? It looks like we initialize those fields once, when ```generateComponentHeaderRow()``` is called. Instead, what if we initialize them in the constructor of ```AbstractComponentDataExportStrategy```?

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
geoffreykwan: incrementRevisionCounter() is called in two places so it's better reduce duplication of `${nodeId}_${componentId}` so you don't have to make that string yourself each time you need to increment the counter.

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,367 @@
...
+    row[this.columnNameToNumber.get('Project ID')] = this.configService.getProjectId();
+    row[this.columnNameToNumber.get('Project Name')] = this.projectService.getProjectTitle();
+    row[this.columnNameToNumber.get('Run ID')] = this.configService.getRunId();
+    row[this.columnNameToNumber.get('Student Work ID')] = componentState.id;
+    const formattedDateTime = this.utilService.convertMillisecondsToFormattedDateTime(
+      componentState.serverSaveTime
+    );
+    row[this.columnNameToNumber.get('Server Timestamp')] = formattedDateTime;
+    const clientSaveTime = new Date(componentState.clientSaveTime);
+    if (clientSaveTime != null) {
```
hirokiterashima: Will never be null since componentState.clientSaveTime is not null, remove?
```suggestion
    if (clientSaveTime != null) {
```

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,367 @@
...
+    );
+    row[this.columnNameToNumber.get('Class Period')] = periodName;
+    row[this.columnNameToNumber.get('Project ID')] = this.configService.getProjectId();
+    row[this.columnNameToNumber.get('Project Name')] = this.projectService.getProjectTitle();
+    row[this.columnNameToNumber.get('Run ID')] = this.configService.getRunId();
+    row[this.columnNameToNumber.get('Student Work ID')] = componentState.id;
+    const formattedDateTime = this.utilService.convertMillisecondsToFormattedDateTime(
+      componentState.serverSaveTime
+    );
+    row[this.columnNameToNumber.get('Server Timestamp')] = formattedDateTime;
```
hirokiterashima: No need to store in variable since we don't use it again?
```suggestion
    row[this.columnNameToNumber.get('Server Timestamp')] = this.utilService.convertMillisecondsToFormattedDateTime(
      componentState.serverSaveTime
    );
```

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,367 @@
...
+    return rows;
+  }
+
+  /**
+   * @param users An array of user objects. Each user object contains an id and name.
+   * @returns {object} An object that contains key/value pairs. The key is userIdX
+   * or studentNameX where X is an integer. The values are the corresponding actual
+   * values of user id and student name.
+   */
+  private extractUserIDsAndStudentNames(users: any[]): any {
```
hirokiterashima: Define class to store this object that ```extends Map<string, string>```?
```suggestion
  private extractUserIDsAndStudentNames(users: any[]): UserIdsAndStudentNames {
```

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,367 @@
...
+  generateComponentHeaderRow(
+    columnNames: string[],
+    columnNameToNumber: Map<string, number>
+  ): string[] {
+    this.populateColumnNames(columnNames, columnNameToNumber);
+    const headerRow = [];
+    for (const columnName of columnNames) {
+      headerRow.push(columnName);
+    }
+    return headerRow;
```
hirokiterashima: Write this in one line for brevity (1 line vs 5)?
```suggestion
    return columnNames.map((columnName) => columnName);
    or 
    return [...columnNames];
```

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,367 @@
...
+    periodName: string,
+    componentRevisionCounter: ComponentRevisionCounter,
+    componentState: any
+  ): string[];
+
+  generateComponentHeaderRow(
+    columnNames: string[],
+    columnNameToNumber: Map<string, number>
+  ): string[] {
+    this.populateColumnNames(columnNames, columnNameToNumber);
```
hirokiterashima: I don't think populating column names should be a part of generating header rows routine. Should ```populateColumnNames()``` be called in the constructor?

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,367 @@
...
+    componentRevisionCounter: ComponentRevisionCounter,
+    nodeId: string,
+    componentId: string
+  ): number {
+    const nodeIdAndComponentId = `${nodeId}_${componentId}`;
+    if (!componentRevisionCounter.has(nodeIdAndComponentId)) {
+      componentRevisionCounter.set(nodeIdAndComponentId, 1);
+    }
+    return componentRevisionCounter.get(nodeIdAndComponentId);
+  }
```
hirokiterashima: Should this be a method in the ComponentRevisionCounter class? 
```getRevisionCounter(nodeId: string, componentId: string) { ...```

**Review by hirokiterashima — CHANGES_REQUESTED** [review]
The code is getting easier to read. I added additional code change suggestions inline based on the new changes.

File: src/assets/wise5/classroomMonitor/dataExport/strategies/LabelComponentDataExportStrategy.ts [hunk 24]
```diff
@@ -0,0 +1,77 @@
...
+  constructor(component: Component, additionalParams: ComponentDataExportParams) {
+    super(component, additionalParams);
+    this.maxNumLabels = 0;
+  }
+
+  export(): void {
+    this.controller.showDownloadingExportMessage();
+    const components = [{ nodeId: this.component.nodeId, componentId: this.component.id }];
+    this.dataExportService.retrieveStudentDataExport(components).then((result) => {
+      this.populateColumnNames(this.columnNames, this.columnNameToNumber);
```
hirokiterashima: Does this ```populateColumnNames()``` call need to be here (after waiting for return of ```retrieveStudentDataExport()```), or can it be called sooner, inside the constructor of ```AbstractComponentDataExportStrategy```? Does it need to pass in the two params to the function, or can the func get them from the class fields?

src/assets/wise5/classroomMonitor/dataExport/strategies/LabelComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,77 @@
+import { Component } from '../../../common/Component';
+import { ComponentDataExportParams } from '../ComponentDataExportParams';
+import { ComponentRevisionCounter } from '../ComponentRevisionCounter';
+import { AbstractComponentDataExportStrategy } from './AbstractComponentDataExportStrategy';
+
+export class LabelComponentDataExportStrategy extends AbstractComponentDataExportStrategy {
+  maxNumLabels: number;
```
hirokiterashima: Initialize value here, instead of in the constructor?
```suggestion
  maxNumLabels: number = 0;
```

src/assets/wise5/classroomMonitor/dataExport/strategies/LabelComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  constructor(component: Component, additionalParams: ComponentDataExportParams) {
+    super(component, additionalParams);
+    this.maxNumLabels = 0;
```
hirokiterashima: Initialize along with class variable declaration?
```suggestion
```

src/assets/wise5/classroomMonitor/dataExport/strategies/LabelComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  private updateMaxNumLabelsIfNecessary(labelComponentState: any): void {
+    const numLabels = labelComponentState.studentData.labels.length;
+    if (numLabels > this.maxNumLabels) {
+      this.maxNumLabels = numLabels;
+    }
+  }
+
+  private addLabelHeaders(rows: any[], maxNumLabels: number): void {
```
hirokiterashima: Type to array of string arrays?
```suggestion
  private addLabelHeaders(rows: string[][], maxNumLabels: number): void {
```

src/assets/wise5/classroomMonitor/dataExport/ComponentRevisionCounter.ts
```diff
@@ -0,0 +1,17 @@
+export class ComponentRevisionCounter extends Map<string, number> {
+  incrementCounter(nodeId: string, componentId: string): void {
+    const key = `${nodeId}_${componentId}`;
```
hirokiterashima: Extract to func since it's duplicated in getCounter()?
```suggestion
    const key = this.getKey(nodeId, componentId);
```

src/assets/wise5/classroomMonitor/dataExport/ComponentRevisionCounter.ts
```diff
@@ -0,0 +1,17 @@
...
+    if (!this.has(key)) {
+      this.set(key, 1);
+    }
```
hirokiterashima: Extract to func since it's duplicated in getCounter()?
```suggestion
    this.initializeKey(key);
```

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,306 @@
...
+
+  private generateComponentWorkRowsForWorkgroup(
+    workgroupId: number,
+    nodeId: string,
+    componentId: string,
+    rowCounter: number
+  ): string[] {
+    const rows = [];
+    const userInfo = this.configService.getUserInfoByWorkgroupId(workgroupId);
+    const extractedUserIdsAndStudentNames = new UserIdsAndStudentNames(
```
hirokiterashima: "extracted" is not necessary, rename to "userIdsAndStudentNames"?

src/assets/wise5/classroomMonitor/dataExport/strategies/LabelComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,77 @@
...
+    this.maxNumLabels = 0;
+  }
+
+  export(): void {
+    this.controller.showDownloadingExportMessage();
+    const components = [{ nodeId: this.component.nodeId, componentId: this.component.id }];
+    this.dataExportService.retrieveStudentDataExport(components).then((result) => {
+      this.populateColumnNames(this.columnNames, this.columnNameToNumber);
+      let rows = [this.generateComponentHeaderRow(this.columnNames)];
+      rows = rows.concat(this.generateComponentWorkRows(this.component, this.component.nodeId));
```
hirokiterashima: Pass in component only, handle getting nodeId downstream (i.e. inside generateComponentWorkRows() func and its callees as well)?
```suggestion
      rows = rows.concat(this.generateComponentWorkRows(this.component));
```

src/assets/wise5/classroomMonitor/dataExport/strategies/LabelComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,77 @@
...
+      this.addLabelHeaders(rows, this.maxNumLabels);
+      const fileName = this.generateExportFileName(
+        this.component.nodeId,
+        this.component.id,
+        'label'
+      );
```
hirokiterashima: Pass in component only, handle getting nodeId, componentId, and type inside the ```generateExportFileName()``` function?
```suggestion
      const fileName = this.generateExportFileName(this.component);
```

hirokiterashima (changes requested):
The code is getting easier to read. I added additional code change suggestions inline based on the new changes.

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,309 @@
...
+    nodeId: string,
+    componentId: string,
+    rowCounter: number
+  ): string[] {
+    const rows = [];
+    const userInfo = this.configService.getUserInfoByWorkgroupId(workgroupId);
+    const userIdsAndStudentNames = new UserIdsAndStudentNames(
+      userInfo.users,
+      this.canViewStudentNames
+    );
```
hirokiterashima, line 90: Call this and pass in as 3rd param to this.generateComponentWorkRow() below?
```suggestion
```

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,309 @@
...
+      if (this.shouldExportRow(componentState, c, componentStates.length)) {
+        const row = this.generateComponentWorkRow(
+          rowCounter,
+          workgroupId,
+          userIdsAndStudentNames.getUserId(1),
+          userIdsAndStudentNames.getUserId(2),
+          userIdsAndStudentNames.getUserId(3),
+          userIdsAndStudentNames.getStudentName(1),
+          userIdsAndStudentNames.getStudentName(2),
+          userIdsAndStudentNames.getStudentName(3),
```
hirokiterashima: Pass in ```UserIdsAndStudentNames``` object instead of individual userId1,userId2,...? Do this all the way down the call stack until it's actually needed in setStudentIDsAndNames(), i.e. ```generateComponentWorkRow() -> createStudentWorkExportRow() -> setStudentIDsAndNames()```?
```suggestion
    new UserIdsAndStudentNames(
      userInfo.users,
      this.canViewStudentNames
    ),
```

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,309 @@
...
+      1;
+    row[this.columnNameToNumber.get('Component Part Number')] = componentPartNumber;
+    row[this.columnNameToNumber.get('Component Type')] = this.component.content.type;
+    if (this.component.content.prompt != null) {
+      let prompt = this.utilService.removeHTMLTags(this.component.content.prompt);
+      prompt = prompt.replace(/"/g, '""');
+      row[this.columnNameToNumber.get('Component Prompt')] = prompt;
+    }
+    const studentData = componentState.studentData;
+    if (studentData != null) {
```
hirokiterashima: Will never be null, remove?
```suggestion
```

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,309 @@
...
+      row[this.columnNameToNumber.get('Student Data')] = studentData;
+      const isCorrect = studentData.isCorrect;
+      if (isCorrect != null) {
+        if (isCorrect) {
+          row[this.columnNameToNumber.get('Is Correct')] = 1;
+        } else {
+          row[this.columnNameToNumber.get('Is Correct')] = 0;
+        }
```
hirokiterashima: Replace with ternary for brevity (5 lines vs 1)?
```suggestion
        row[this.columnNameToNumber.get('Is Correct')] = isCorrect ? 1 : 0;
```

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,309 @@
...
+       * because the revision counter needs to be previously calculated
+       * and then set into the component state
+       */
+      row[this.columnNameToNumber.get('Component Revision Counter')] =
+        componentState.revisionCounter;
+    }
+    componentRevisionCounter.incrementCounter(componentState.nodeId, componentState.componentId);
+    if (componentState.isSubmit) {
+      row[this.columnNameToNumber.get('Is Submit')] = 1;
+      if (studentData != null) {
```
hirokiterashima: Will never be null, remove?
```suggestion
```

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,309 @@
...
+      'Student Data',
+      'Component Revision Counter',
+      'Is Submit',
+      'Submit Count'
+    ];
+    for (let c = 0; c < defaultColumnNames.length; c++) {
+      const defaultColumnName = defaultColumnNames[c];
+      this.columnNameToNumber.set(defaultColumnName, c);
+      this.columnNames.push(defaultColumnName);
+    }
```
hirokiterashima: It looks like defaultColumnNames and this.columnNames are the same? Is this also the case with other component types? If so, can we instead declare&instantiate the class's columnNames field with the array of string, and change this function to:
```suggestion
    for (let c = 0; c < this.columnNames.length; c++) {
      this.columnNameToNumber.set(this.columnNames[c], c);
    }
```

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,309 @@
...
+  }
+
+  generateComponentWorkRows(component: Component): string[] {
+    let rows = [];
+    let rowCounter = 1;
+    for (const workgroupId of this.configService.getClassmateWorkgroupIds()) {
+      const rowsForWorkgroup = this.generateComponentWorkRowsForWorkgroup(
+        workgroupId,
+        component.nodeId,
+        component.id,
```
hirokiterashima: Is there a need to separate the component to nodeId and componentId and pass in two params here, or should we let ```generateComponentWorkRowsForWorkgroup()``` handle that?
```suggestion
        component,
```

hirokiterashima (changes requested):
The code is getting easier to read. I added additional code change suggestions inline based on the new changes.

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
```diff
@@ -0,0 +1,309 @@
...
+    for (let c = 0; c < defaultColumnNames.length; c++) {
+      const defaultColumnName = defaultColumnNames[c];
+      this.columnNameToNumber.set(defaultColumnName, c);
+      this.columnNames.push(defaultColumnName);
+    }
+  }
+
+  generateComponentWorkRows(component: Component): string[] {
+    let rows = [];
+    let rowCounter = 1;
```
hirokiterashima: Make the rowCounter into a class field (declare&instantiate to 1 on the same line) instead and remove the need to pass it around in function calls?

hirokiterashima (approved):
Looks good.

PR #942, opened 2022-12-08
chore(MilestoneReportGraph): Upgrade to Angular

### LINKED ISSUE #941: Upgrade MilestoneReportGraph to Angular
Right now it's in AngularJS

hirokiterashima:
## Changes
- Upgrade MilestoneReportGraph from AngularJS to Angular
- Clean up code
   - Remove chart.reflow() call because it didn't adjust width as commented in MilestoneReportGraphController.js. Instead, fix width to 200.
   - Create MilestoneModule. Future task to move other milestone classes into this module. If we were to do it now it would cause a lot of files to be moved around.

## Test
- Milestone report graph appears and works as before

Closes #941


src/app/teacher/milestone/milestone-report-graph/milestone-report-graph.component.ts
```diff
@@ -0,0 +1,132 @@
...
+      },
+      yAxis: {
+        title: {
+          text: ''
+        },
+        labels: {
+          enabled: false
+        }
+      },
+      series: this.series
```
geoffreykwan, line 90: We can hide the Download image button and `Highcharts.com` link at the bottom right of the graph if we add this to the chartConfig.

```
      exporting: {
        enabled: false
      },
      credits: {
        enabled: false
      },
      series: this.series
```

geoffreykwan (approved):
Looks good.

src/app/teacher/milestone/milestone-report-graph/milestone-report-graph.component.ts
hirokiterashima: Added. Thanks for the tip.

PR #944, opened 2022-12-08
feat(Label Component): Order labels from oldest to newest

### LINKED ISSUE #943: feat(Label Component): Order labels from oldest to newest
- Order the labels from oldest to newest. Currently new labels are added to the front of the array of labels but should be added to the end.
- Add isStarterLabel field to labels.

geoffreykwan:
## Changes

- Timestamp is saved in labels
- Labels are now ordered from oldest to newest using the timestamp when saved
- Added isStarterLabel field to labels
- When a new starter label is authored, the isStarterLabel field is set to true

## Test

- Make sure labels are saved from oldest to newest in the labels array in the student data
- Make sure isStarterLabel field is saved for labels

Closes #943

### DIFF
--- a/src/assets/wise5/components/label/label-student/label-student.component.ts
+++ b/src/assets/wise5/components/label/label-student/label-student.component.ts
@@ -303,14 +303,12 @@ export class LabelStudent extends ComponentStudent {
   }
 
   private setStarterLabels(componentContent: any): void {
+    // Make sure starter labels have isStarterLabel set to true. Starter labels from old Label
+    // component content did not have this field.
     this.setIsStarterLabelTrue(componentContent.labels);
     this.addLabelsToCanvas(componentContent.labels);
   }
 
-  /**
-   * Make sure starter labels have isStarterLabel set to true. Starter labels from old Label
-   * component content did not have this field.
-   */
   private setIsStarterLabelTrue(labels: any[]): void {
     for (const label of labels) {
       label.isStarterLabel = true;
--- a/src/assets/wise5/components/label/label-student/label-student.component.ts
+++ b/src/assets/wise5/components/label/label-student/label-student.component.ts
@@ -292,13 +292,27 @@ export class LabelStudent extends ComponentStudent {
       this.setStudentWork(componentState);
     } else if (this.component.hasConnectedComponent()) {
       this.handleConnectedComponents();
-      if (this.componentContent.labels != null) {
-        this.addLabelsToCanvas(componentContent.labels);
+      if (componentContent.labels != null) {
+        this.setStarterLabels(componentContent);
       }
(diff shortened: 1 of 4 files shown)


src/assets/wise5/components/label/label-student/label-student.component.ts
```diff
@@ -346,9 +360,14 @@ export class LabelStudent extends ComponentStudent {
     this.canvas.getObjects('i-text').forEach((object: any) => {
       labels.push(this.getLabelJSONObjectFromText(object));
     });
+    labels.sort(this.sortByTimestamp);
     return labels;
   }
 
+  sortByTimestamp(labelA: any, labelB: any): number {
```
hirokiterashima: Make function private and name more descriptive?
```suggestion
  private sortByTimestampOldestFirst(labelA: any, labelB: any): number {
```

src/assets/wise5/components/label/label-student/label-student.component.ts
```diff
@@ -292,13 +292,27 @@ export class LabelStudent extends ComponentStudent {
...
+  }
+
+  private setStarterLabels(componentContent: any): void {
+    this.setIsStarterLabelTrue(componentContent.labels);
+    this.addLabelsToCanvas(componentContent.labels);
+  }
+
+  /**
+   * This is for old Label component content before we had the isStarterLabel field in the labels
+   */
```
hirokiterashima: This comment is a bit confusing because the function can be called with labels that have isStarterLabel field already set to true (e.g. newly-authored content). Rephrase or remove?

hirokiterashima (changes requested):
Functionality works well. 

I added two comments inline. I understand the want to add a comment about old content not having isStarterLabel set to true, but the way it was written was confusing, and I'm not sure if it's really necessary.

Also- remember to set project details on this PR and issue.

src/assets/wise5/components/label/label-student/label-student.component.ts
hirokiterashima: Revised comment is clearer, but I think it should be moved to just above the call to the ```setIsStarterLabelTrue()``` inside ```setStarterLabels()```, instead of above the ```setIsStarterLabelTrue()``` function definition.

hirokiterashima (commented):
Revised comment is clearer, but I think it should be moved to just above the call to the ```setIsStarterLabelTrue()``` inside ```setStarterLabels()```, instead of above the ```setIsStarterLabelTrue()``` function definition. Putting the comment above the function call describes why you're having to call the function ("Make sure starter labels have isStarterLabel set to true...")

What do you think?

hirokiterashima (approved):
Looks good.

PR #954, opened 2022-12-18
fix(Export): Request uri too large error

### LINKED ISSUE #953: fix(Export): Request uri too large error
When the teacher selects too many components to export, a request-uri too large error is thrown.

The teacher can select which components to export in these exports
- One Workgroup Per Row
- Latest Work
- All Work
- Raw data

geoffreykwan:
Test with
- https://github.com/WISE-Community/WISE-API/pull/199

## Changes

When the teacher requests student data for specific components, we now compress the components list param string so that we no longer receive the "request uri too large" error.

## Test

Run npm install to get the new dependency.

Test these exports and make sure they work like before. Before the fix, if you chose "Select Steps to Export" and then selected 50 or more steps (by clicking "Select All") this would create an error and no data would be retrieved and no export would be generated.
- One Workgroup Per Row (try both Export All Steps and Select Steps To Export)
- Latest Work (try both Export All Steps and Select Steps To Export)
- All Work (try both Export All Steps and Select Steps To Export)
- Component Data
- Raw Data (try both Export All Steps and Select Steps To Export)

Also test these pages that use the same component data request for retrieving data
- Node Grading View
- Milestone Grading View

Closes #953

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -52,6 +52,7 @@
         "iframe-resizer": "^3.6.6",
         "jquery": "^3.6.0",
         "jwt-decode": "^3.1.2",
+        "lz-string": "^1.4.4",
         "mathjax": "^3.2.2",
         "ng-file-upload": "^12.2.13",
         "ng-recaptcha": "^9.0.0",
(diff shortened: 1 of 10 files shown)


hirokiterashima (approved):
LGTM. Remember to add project info on the PR.

Can you remove the unused "result" variable in ```.then((result) => {``` blocks in ```OneWorkgroupPerRowDataExportStrategy``` and ```RawDataExportStrategy``` before merging?

PR #960, opened 2023-01-09
fix(User Registration): Form now properly shows up in Safari browser

### LINKED ISSUE #959: Register user form does not show up when using Safari browser
When using the Safari browser, on the student and teacher registration page where you enter in your account information such as first name, last name, password, etc., an error is thrown in the console and the form does not show up at all so all you see is the WISE logo and a blank page.

Here is the error
```
Error: Uncaught (in promise): SyntaxError: Invalid regular expression: invalid group specifier name RegExp@[native code]
```

geoffreykwan:
## Changes

In the user registration pages, changed name validation regex to not use lookahead and lookbehind because Safari does not support them yet.

## Test

- Make sure you can register new student accounts and new teacher accounts when using Safari
- Make sure you can use spaces and dashes in the first and last name
- Make sure the name validation still works how we want it to

Closes #959


hirokiterashima (approved):
LGTM. Remember to set the project on the PR

PR #962, opened 2023-01-11
feat(QuestionBank): Choose questions from multiple matched rules up to max

### LINKED ISSUE #961: feat(QuestionBank): Choose questions from multiple matched rules up to max
### Description

Add a new ```maxQuestionsToShow: number``` field to  QuestionBank content. Use this number to choose questions to show in the QuestionBank, up to the specified number.

First find all rules that match (including the default rule), and then choose the questions from the rules in round-robin order

Ex:
```
Rule 1 questions: ["q1", "q2", "q3"]
Rule 2 questions: ["q4"]
Rule 3 questions: ["q5", "q6"]
```
If ```maxQuestionsToShow=4```, it should choose 
```
Rule 1 questions: ["q1", "q2"]
Rule 2 questions: ["q4"]
Rule 3 questions: ["q5"]
```

If maxQuestionsToShow is larger than the number of matched questions, show all questions.

hirokiterashima:
## Changes
- Add a new ```maxQuestionsToShow: number``` field to  QuestionBank content. Use this number to choose questions to show in the QuestionBank, up to the specified number. The algorithm works as follows
   - First find all rules that match (including the default rule)
   - Next, choose one question from each matched rule in round-robin order until maxQuestionsToShow is reached, or until all the questions have been used
   
For example, if matched rules looked like this:
```
Rule 1 questions: ["q1", "q2", "q3"]
Rule 2 questions: ["q4"]
Rule 3 questions: ["q5", "q6"]
```
and ```maxQuestionsToShow=4```, it should choose 
```
Rule 1 questions: ["q1", "q2"]
Rule 2 questions: ["q4"]
Rule 3 questions: ["q5"]
```
- Store matched rules and chosen questions within those rules along with PeerChat student work

## Test
- In PeerChat authoring, enable QuestionBank and set "Number of Questions to Show" to a number
- As a student, go to the PeerChat step and verify that QuestionBank shows up to that many questions
   - The questions should be chosen in round-robin fashion from all the rules that match
   - Type some text in the chatbox to trigger saving the QuestionBank items 
- As a teacher, verify that the QuestionBank questions appear in the grade-by-step view 

Closes #961

### DIFF
--- a/src/app/authoring-tool/edit-question-bank/edit-question-bank.component.html
+++ b/src/app/authoring-tool/edit-question-bank/edit-question-bank.component.html
@@ -9,6 +9,12 @@
     </mat-checkbox>
   </div>
   <div *ngIf="componentContent.questionBank?.enabled">
+    <div class="max-questions-to-show-div">
+      <mat-form-field class="form-field-no-hint" appearance="fill">
+        <mat-label i18n>Number of Questions to Show</mat-label>
+        <input matInput [(ngModel)]="componentContent.questionBank.maxQuestionsToShow" (ngModelChange)='saveChanges()' />
+      </mat-form-field>
+    </div>
     <div class="reference-component-div"
         fxLayout="row wrap"
         fxLayoutAlign="start center"
--- a/src/app/authoring-tool/edit-question-bank/edit-question-bank.component.html
+++ b/src/app/authoring-tool/edit-question-bank/edit-question-bank.component.html
@@ -13,6 +13,7 @@
       <mat-form-field appearance="fill">
         <mat-label i18n>Max Number of Questions</mat-label>
         <input matInput
+            type="number"
             [(ngModel)]="componentContent.questionBank.maxQuestionsToShow"
             (ngModelChange)='saveChanges()' />
       </mat-form-field>
--- a/src/app/authoring-tool/edit-question-bank/edit-question-bank.component.html
+++ b/src/app/authoring-tool/edit-question-bank/edit-question-bank.component.html
@@ -1,5 +1,5 @@
 <div class="section-container">
-  <div class="enable-question-bank-div">
+  <div>
     <mat-checkbox
         [checked]="componentContent.questionBank?.enabled"
         (change)="toggleComponent($event)"
@@ -9,15 +9,15 @@
     </mat-checkbox>
   </div>
   <div *ngIf="componentContent.questionBank?.enabled">
-    <div class="max-questions-to-show-div">
-      <mat-form-field class="form-field-no-hint" appearance="fill">
-        <mat-label i18n>Max Number of Questions to Show</mat-label>
+    <div class="max-questions">
+      <mat-form-field appearance="fill">
+        <mat-label i18n>Max Number of Questions</mat-label>
         <input matInput
             [(ngModel)]="componentContent.questionBank.maxQuestionsToShow"
             (ngModelChange)='saveChanges()' />
       </mat-form-field>
     </div>
-    <div class="reference-component-div"
+    <div class="reference-component"
         fxLayout="row wrap"
         fxLayoutAlign="start center"
         fxLayoutGap="20px">
@@ -36,7 +36,7 @@
           [componentContent]="componentContent.questionBank">
       </edit-component-peer-grouping-tag>
     </div>
-    <div class="feedback-rules-div">
+    <div class="feedback-rules">
       <edit-question-bank-rules [feedbackRules]="componentContent.questionBank.rules">
       </edit-question-bank-rules>
     </div>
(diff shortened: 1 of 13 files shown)


src/app/authoring-tool/edit-question-bank/edit-question-bank.component.html
```diff
@@ -9,6 +9,12 @@
     </mat-checkbox>
   </div>
   <div *ngIf="componentContent.questionBank?.enabled">
+    <div class="max-questions-to-show-div">
+      <mat-form-field class="form-field-no-hint" appearance="fill">
+        <mat-label i18n>Number of Questions to Show</mat-label>
+        <input matInput [(ngModel)]="componentContent.questionBank.maxQuestionsToShow" (ngModelChange)='saveChanges()' />
```
geoffreykwan: Add a line break to this line because it's more than 100 columns wide.

src/assets/wise5/components/peerChat/peer-chat-question-bank/peer-chat-question-bank.component.ts
```diff
@@ -95,4 +121,8 @@ export class PeerChatQuestionBankComponent implements OnInit {
       })
     );
   }
+
+  private setQuestions(): void {
+    this.questions = this.displayedQuestionBankRules.flatMap((rule) => rule.questions);
+  }
```
geoffreykwan: Would it be better to pass in displayedQuestionBankRules as a parameter?

src/assets/wise5/components/peerChat/peer-chat-question-bank/peer-chat-question-bank.component.ts
```diff
@@ -44,38 +47,61 @@ export class PeerChatQuestionBankComponent implements OnInit {
...
+    const maxQuestionsToShow = this.content.questionBank.maxQuestionsToShow;
+    while (numAdded < maxQuestionsToShow && numAdded != totalNumQuestions) {
+      if (questionBankRules[ruleIndex].questions.length > 0) {
+        const question = questionBankRules[ruleIndex].questions.shift();
+        filteredRules[ruleIndex].questions.push(question);
+        numAdded++;
+      }
+      ruleIndex = (ruleIndex + 1) % questionBankRules.length;
+    }
+    return filteredRules.filter((rule) => rule.questions.length > 0);
```
geoffreykwan: Would it be better to pass in maxQuestionsToShow as a parameter to filterQuestions() so it can be a pure function?

```JSON.parse(JSON.stringify(questionBankRules))```
can be changed to
```this.utilService.makeCopyOfJSONObject(questionBankRules)```

geoffreykwan (changes requested):
Functionality works well.

Should we change "Number of Questions to Show" to "Max Number of Questions to Show" to be more descriptive?

If the author does not specify the "Number of Questions to Show" it won't show any questions. Instead, if they don't specify a value, should we show all the questions that are satisfied? Right now if they want to show all the questions that are satisfied, they would have to set the "Number of Questions to Show" to a large number like 100.

src/app/authoring-tool/edit-question-bank/edit-question-bank.component.html
hirokiterashima: Should we set the default maxQuestionsToShow to something like 3 instead of "all"? That would probably be the more likely case?

src/assets/wise5/components/peerChat/peer-chat-question-bank/peer-chat-question-bank.component.ts
hirokiterashima: I don't think we should add a UtilService dependency to this class just to make a copy of the object.

src/assets/wise5/components/peerChat/peer-chat-question-bank/peer-chat-question-bank.component.ts
hirokiterashima: Yes, I made the change. PTAL

src/app/authoring-tool/edit-question-bank/edit-question-bank.component.html
geoffreykwan: I think it would make sense to make an empty maxQuestionsToShow have it show all the questions. The maxQuestionsToShow is an optional added restriction. I think it would be ok to leave it blank by default and let the author specify but it may depend on how often authors choose to use that field and what they usually set it to.

src/assets/wise5/components/peerChat/peer-chat-question-bank/peer-chat-question-bank.component.ts
geoffreykwan: Calling JSON.parse(JSON.stringify()) is adding duplication. The point of the UtilService is to provide common functions that our components use.

breity (approved):

geoffreykwan (approved):
Looks good.

PR #964, opened 2023-01-12
feat(Authoring): Add JSON authoring view for when project is broken

### LINKED ISSUE #922: Authoring edit JSON view when project is broken
When a project is broken, allow the Authoring Tool to open the view where you can edit the JSON.

geoffreykwan:
## Changes

- Added recovery view to allow editing the project JSON without loading the project
- The recovery view also tries to detect and display potential problems

## Test

1. Go to the recovery authoring view for a project.
```http://localhost:81/teacher/edit/unit/<unit id>/recovery```
2. Modify the JSON so that it becomes invalid, the upper right message should say "JSON Invalid" and the "Save" button should be disabled.
3. Modify the JSON so that it becomes valid, the upper right message should say "JSON Valid" and the "Save" button should be enabled.
4. Modify the project so that a node has a null transition and make sure the potential problem is detected.
```
        {
            "id": "node1",
            "transitionLogic": {
                "transitions": [
                    {
                        "to": null
                    }
                ]
            }
        }
```
5. Modify the project so that a group references a node id that does not exist and make sure the potential problem is detected.
```
        {
            "id": "group1",
            "ids": [
                "node1",
                "node9999"
            ]
        }
```
6. Modify the project so that a group references a node id multiple times and make sure the potential problem is detected.
```
        {
            "id": "group1",
            "ids": [
                "node1",
                "node1"
            ]
        }
```
7. Make sure the "Go to Authoring View" button works if the project has no problems.

Closes #922

### DIFF
--- a/src/app/domain/nodeRecoveryAnalysis.ts
+++ b/src/app/domain/nodeRecoveryAnalysis.ts
@@ -1,0 +1,30 @@
+export class NodeRecoveryAnalysis {
+  hasTransitionToNull: boolean = false;
+  nodeId: string;
+  referencedIdsThatAreDuplicated: string[] = [];
+  referencedIdsThatDoNotExist: string[] = [];
+
+  constructor(nodeId: string) {
+    this.nodeId = nodeId;
+  }
+
+  addReferencedIdThatIsDuplicated(nodeId: string): void {
+    this.referencedIdsThatAreDuplicated.push(nodeId);
+  }
+
+  addReferencedIdThatDoesNotExist(nodeId: string): void {
+    this.referencedIdsThatDoNotExist.push(nodeId);
+  }
+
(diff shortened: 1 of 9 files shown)


hirokiterashima:
Works as described. 

Should we allow saving when there are issues (null transition, node id that dne, etc)?

Can you see if you can reduce number of code issues in this change? https://codeclimate.com/github/WISE-Community/WISE-Client/pull/964#issues

geoffreykwan:
I think it might be ok to allow saving if there are issues. Theoretically if you're going to the recovery view it's because the project is already broken. Then if you require no issues in order to save, you have to fix everything all at once which might make it more difficult on the user.

hirokiterashima (approved):
LGTM

PR #966, opened 2023-01-13
chore(ng generate): Create components without app prefix

### LINKED ISSUE #965: Have ng generate component create components without a prefix
Right now whenever we create a component with ```ng generate component``` it creates the component and sets the selector to have the prefix of "app-" but we don't want that prefix to be automatically added every time.

geoffreykwan:
## Changes

When generating a new component with `ng generate component`, the `app-` prefix is no longer added to the component selector.

## Test

- Run `ng generate component` and make sure it creates a component without `app-` in the selector.
- Run `ng generate directive` and make sure it creates a directive without `app-` in the selector.

Closes #965

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -237,11 +237,11 @@
   },
   "schematics": {
     "@schematics/angular:component": {
-      "prefix": "app",
+      "prefix": "",
       "style": "scss"
     },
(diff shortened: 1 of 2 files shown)


hirokiterashima (approved):
LGTM

PR #969, opened 2023-01-19
refactor(UtilService): Move makeCopyOfJSONObject() to its own file

### LINKED ISSUE #963: refactor(UtilService): Move makeCopyOfJSONObject() to its own file
- Create Object.ts
- Move UtilService.makeCopyOfJSONObject() to copy() in the new file
- Change all calls to UtilService.makeCopyOfJSONObject() to instead call copy()
- Change all calls to JSON.parse(JSON.stringify()) to instead call copy()

hirokiterashima:
## Changes
- Create Object.ts
- Move UtilService.makeCopyOfJSONObject() to copy() in the new file
- Change all calls to UtilService.makeCopyOfJSONObject() to instead call copy()
- Change all calls to JSON.parse(JSON.stringify()) to instead call copy()

## Test
- VLE, AT, CM (including raw export) work as before

Closes #963


src/assets/wise5/common/object/object.spec.ts
```diff
@@ -0,0 +1,14 @@
...
+    const arr = [1, 2, 3];
+    expect(copy(arr)).toEqual(arr);
+    const obj = { name: 'WISE', address: 'Berkeley' };
+    expect(copy(obj)).toEqual(obj);
+  });
+  it('should handle null and undefined inputs', () => {
+    expect(copy(null)).toEqual(null);
+    expect(copy(undefined)).toEqual(undefined);
+  });
+});
```
geoffreykwan: I think these 2 tests should be broken into 4 tests. I think the array and object test should make a copy and then modify the original object and make sure the new copy does not receive that modification.

geoffreykwan (changes requested):

geoffreykwan (approved):
Looks good.

PR #977, opened 2023-01-23
fix(Preview): Fix preview step without constraints

### LINKED ISSUE #975: Preview step without constraints does not load the preview unit properly
1. Load a unit in the Authoring Tool
2. Click on a step
3. Click "Preview Step Without Constraints"
4. When the preview tab opens, you will see the Project Plan view and a broken WISE Logo image and broken styling. What should happen is you should see the step loaded and "Constraints Are Off" in the upper right of the screen.

geoffreykwan:
## Changes

Fixed preview step without constraints.

## Test

1. Load a unit in the Authoring Tool
2. Click on a step
3. Click "Preview Step Without Constraints"
4. You should see the step load and "Constraints Are Off" in the upper right of the screen. Previously, the VLE would not load properly. You would see the Project Plan view and a broken WISE Logo image and broken styling.
5. Also make sure preview normally with constraints also still works

Closes #975

### DIFF
--- a/src/assets/wise5/authoringTool/node/nodeAuthoringComponent.ts
+++ b/src/assets/wise5/authoringTool/node/nodeAuthoringComponent.ts
@@ -151,15 +151,7 @@ class NodeAuthoringController {
     this.subscriptions.unsubscribe();
   }
 
-  previewStepInNewWindow() {
-    this.savePreviewEventAndOpenPreview(true);
-  }
-
-  previewStepWithoutConstraintsInNewWindow() {
-    this.savePreviewEventAndOpenPreview(false);
-  }
-
-  private savePreviewEventAndOpenPreview(constraints: boolean) {
+  protected previewStepInNewWindow(constraints: boolean): void {
     this.saveStepPreviewedEvent(constraints);
     window.open(this.createPreviewURL(this.nodeId, constraints));
   }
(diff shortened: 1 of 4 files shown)


src/assets/wise5/authoringTool/node/nodeAuthoringComponent.ts
```diff
@@ -152,18 +152,29 @@ class NodeAuthoringController {
...
 
   previewStepWithoutConstraintsInNewWindow() {
-    const data = { constraints: false };
+    this.savePreviewEventAndOpenPreview(false);
+  }
+
+  private savePreviewEventAndOpenPreview(constraints: boolean) {
+    this.saveStepPreviewedEvent(constraints);
+    window.open(this.createPreviewURL(this.nodeId, constraints));
+  }
```
hirokiterashima: Change this function to protected (Angular 14 new feature that works with calls from templates), and replace calls (in the template) to ```previewStepWithoutConstraintsInNewWindow()``` and ```previewStepInNewWindow()``` to directly call this function? Is there a need to have those two functions if all they do is call this function?

```suggestion
  protected savePreviewEventAndOpenPreview(constraints: boolean): void {
    this.saveStepPreviewedEvent(constraints);
    window.open(this.createPreviewURL(this.nodeId, constraints));
  }
```

hirokiterashima (changes requested):
Functionality looks good. I added a code suggestion.

hirokiterashima (approved):
LGTM

PR #1039, opened 2023-02-14
refactor(UtilService): Move getIntersectOfArrays() to own file

### LINKED ISSUE #995: refactor(UtilService): Move getIntersectOfArrays()

geoffreykwan:
## Changes

- Moved getIntersectOfArrays() function from utilService.ts to array.ts
- Deleted two functions in teacherDataService.ts that were no longer called

## Test

- Make sure student work and scores show up properly in the Classroom Monitor
- Make sure exports work properly

Closes #995

### DIFF
--- a/src/app/services/utilService.spec.ts
+++ b/src/app/services/utilService.spec.ts
@@ -12,7 +12,6 @@ describe('UtilService', () => {
   convertStringToNumberTests();
   arrayHasNonNullElementTests();
   calculateMeanTests();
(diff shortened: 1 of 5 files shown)


hirokiterashima (commented):
Functionality works well.

Should the test's ```it()``` statements be more specific, e.g.
"should return an empty array when there are no common items"
"should return an array with common items when there are common items"
...etc?

hirokiterashima (approved):
LGTM

PR #1048, opened 2023-02-17
refactor(UtilService): Move wordWrap()

### LINKED ISSUE #1025: refactor(UtilService): Move wordWrap()

breity:
## Changes
- Move UtilService.wordWrap() to string.ts and update references

## Test
- Add a label to a Label component with more than 20 characters and make sure the text wraps so no lines are longer than 20 characters.
- Create a Label component that imports from an Open Response. Make sure the background image that is generated from the OR content wraps at the specified number of characters.

Closes #1025

### DIFF
--- a/src/assets/wise5/common/string/string.spec.ts
+++ b/src/assets/wise5/common/string/string.spec.ts
@@ -1,4 +1,4 @@
-import { isValidJSONString } from './string';
+import { isValidJSONString, wordWrap } from './string';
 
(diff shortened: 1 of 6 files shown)


hirokiterashima (approved):
LGTM.

I think having another test case for "it should not break string into multiple lines if the string is same/shorter than the wrap limit" would be useful. Can you add that before merging?

PR #1060, opened 2023-02-23
fix(Embedded): Auto comment does not show up immediately

### LINKED ISSUE #1059: fix(Embedded): Auto comment does not show up immediately
1. Sign in as a student to a run for the [Thermodynamics Challenge unit](https://wise.berkeley.edu/preview/unit/39370/node80)
2. Go to Step 2.5
3. Click a cell in the grid (like Hot Liquid + Aluminum)
4. Click "Get Feedback"
5. You should immediately see the feedback show up below the "Get Feedback" button but right now you don't see any feedback.

geoffreykwan:
- Test with https://github.com/WISE-Community/WISE-API/pull/214

## Changes

- Set the latestAnnotations to a new object when it needs to be updated
- Call detectChanges() to have the feedback annotation show. We still need to call detectChanges() because Angular does not know to detect any changes since the feedback is originally triggered from within the Embedded iframe.

## Test

1. Sign in as a student to a run for the [Thermodynamics Challenge unit](https://wise.berkeley.edu/preview/unit/39370/node80)
2. Go to Step 2.5
3. Click a cell in the grid (like Hot Liquid + Aluminum)
4. Click "Get Feedback"
5. You should immediately see the feedback show up below the "Get Feedback" button. It used to not show up.
6. Refresh the VLE and make sure the feedback still shows up.

Closes #1059

### DIFF
--- a/src/assets/wise5/components/component-student.component.ts
+++ b/src/assets/wise5/components/component-student.component.ts
@@ -727,14 +727,6 @@ export abstract class ComponentStudent {
     );
   }
 
-  updateLatestScoreAnnotation(annotation: any): void {
-    this.latestAnnotations.score = annotation;
-  }
-
-  updateLatestCommentAnnotation(annotation: any): void {
-    this.latestAnnotations.comment = annotation;
-  }
-
   registerNotebookItemChosenListener(): void {
     this.subscriptions.add(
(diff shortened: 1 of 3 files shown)


breity (approved):
LGTM.

PR #1067, opened 2023-02-27
refactor(UtilService): Remove arrayHasNonNullElement()

### LINKED ISSUE #1058: refactor(UtilService): Move arrayHasNonNullElement()
This class does not depend on any other fields in UtilService. Move to its own class or write the function inline (instead of defining a class).

geoffreykwan:
## Changes

Removed UtilService.arrayHasNonNullElement(). The function was only called in two places and the check it was performing is no longer needed.

In this scenario the array it receives has already had null values filtered out.

https://github.com/WISE-Community/WISE-Client/blob/951686f083026ef8f1edbc988fb0e6d4684581ca/src/assets/wise5/vle/node/node.component.ts#L250-L258

https://github.com/WISE-Community/WISE-Client/blob/951686f083026ef8f1edbc988fb0e6d4684581ca/src/assets/wise5/vle/node/node.component.ts#L334-L344

In this scenario an auto saved is triggered when the student leaves the step and only components that have dirty work will call this function so it should be called with non-null values.
https://github.com/WISE-Community/WISE-Client/blob/951686f083026ef8f1edbc988fb0e6d4684581ca/src/assets/wise5/vle/node/node.component.ts#L505-L509

## Test

1. Create a step that contains two components (like Open Response)
2. Enable the save buttons on both components and the step
3. Make sure saving works properly and no errors are thrown in these scenarios
- 1 component has dirty work and you click the component save button
- 1 component has dirty work and you click the step save button
- 1 component has dirty work and you go to the next step so the work is auto saved

Closes #1058

### DIFF
--- a/src/app/services/utilService.spec.ts
+++ b/src/app/services/utilService.spec.ts
@@ -10,7 +10,6 @@ describe('UtilService', () => {
     service = TestBed.inject(UtilService);
   });
   convertStringToNumberTests();
-  arrayHasNonNullElementTests();
   calculateMeanTests();
   trimToLength();
   removeHTMLTags();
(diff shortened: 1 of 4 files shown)


hirokiterashima (approved):
LGTM

PR #1068, opened 2023-02-27
refactor(UtilService): Move hasNodeEnteredEvent() to ComponentService

### LINKED ISSUE #1057: refactor(UtilService): Move hasNodeEnteredEvent() to ComponentService
This function is only used by component service classes.

geoffreykwan:
## Changes

Moved hasNodeEnteredEvent() from UtilService to ComponentService.

## Test

Make sure the places that use hasNodeEnteredEvent still work properly. It is used to check if a component is completed by the student. To Test it out
1. Create two steps and in the first step add a component like Discussion.
2. In the second step add the same component type (in this case Discussion) and set the connected component to point to the Discussion in the first step and make it Show Work.
3. Preview the project.
4. Submit a post in the first Discussion step.
5. Go to the second Discussion step. You should see the completed check mark near the step select drop down. This can be tested with all 4 component types below.
- Discussion
- Embedded
- Graph
- Label

Closes #1057

### DIFF
--- a/src/assets/wise5/components/componentService.ts
+++ b/src/assets/wise5/components/componentService.ts
@@ -153,4 +153,10 @@ export class ComponentService {
   isSubmitRequired(node: any, component: any) {
     return node.showSubmitButton || (component.showSubmitButton && !node.showSaveButton);
   }
+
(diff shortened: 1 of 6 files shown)


hirokiterashima (approved):
LGTM

PR #1077, opened 2023-03-02
refactor(UtilService): Move is[Image/Video/Audio]()

### LINKED ISSUE #1000: refactor(UtilService): Move is[Image/Video/Audio]()
These functions don't need to be in UtilService. Move to its own file.

breity:
## Changes
- Move UtilService.is\[Image/Video/Audio\]() functions to own file
- Add RegEx to ensure extension matches are at end of filename string
- Add tests

## Test
- Make sure these tasks work as before:
  - Upload files (images) to student assets in a run (via Discussion) and use in discussion post
  - Upload and preview files in Authoring Tool File Manager
  - Inset files to HTML component using TinyMCE and File Manager

Closes #1000


src/assets/wise5/common/file/file.spec.ts
```diff
@@ -0,0 +1,33 @@
+import { isAudio, isImage, isVideo } from './file';
+
+describe('isImage()', () => {
+  it('should return true if string ends in an image extension type', () => {
+    const filename = 'image.jpg';
+    expect(isImage(filename)).toBeTruthy();
```
hirokiterashima: I think it's fine to pass in the string directly to the function. 
```suggestion
    expect(isImage('image.jpg')).toBeTruthy();
```

hirokiterashima (approved):
LGTM.

You can remove these since they're no longer used:
- StudentAssetService.hasSuffix()
- ProjectAssetAuthoringController.UtilService dependency

Also, see inline code suggestion to improve spec file readability. Looks like the spec file will introduce some new code duplication. Can you think of a way to prevent this? If not, you can go ahead and merge.

PR #1079, opened 2023-03-05
fix(Unit Library): Create run does not redirect to Class Schedule

### LINKED ISSUE #1066: Creating a run does not redirect to Class Schedule
1. Go to the Unit Library
2. Create a run and you should be shown the Run Code
3. Close the create run dialog popup and you should be redirected to the Class Schedule view but nothing happens

geoffreykwan:
## Changes

When a run is created and the Create Run dialog is closed, we no longer broadcastRunChanges() but instead navigate to the Class Schedule route. The TeacherRunListComponent is no longer rendered when the user is in the Unit Library. Maybe in the past it used to be but now it no longer is maybe because it's now lazy loaded. This means the TeacherRunListComponent subscription to the teacherService.runs$ does not exist at that time, and therefore the subscription that contains the code that used to handle changing to the Class Schedule and highlighting the new run no longer gets called.

## Test

1. Go to the Unit Library
2. Create a run and you should be shown the Run Code
3. Close the create run dialog popup and you should be redirected to the Class Schedule view and the new run card should be highlighted

Closes #1066

### DIFF
--- a/src/app/teacher/create-run-dialog/create-run-dialog.component.spec.ts
+++ b/src/app/teacher/create-run-dialog/create-run-dialog.component.spec.ts
@@ -29,8 +29,6 @@ export class MockTeacherService {
     });
   }
 
-  broadcastRunChanges() {}
-
   setTabIndex() {}
 
   checkClassroomAuthorization(): Observable<string> {
--- a/src/app/teacher/create-run-dialog/create-run-dialog.component.ts
+++ b/src/app/teacher/create-run-dialog/create-run-dialog.component.ts
@@ -8,6 +8,7 @@ import { UserService } from '../../services/user.service';
 import { ConfigService } from '../../services/config.service';
 import { ListClassroomCoursesDialogComponent } from '../list-classroom-courses-dialog/list-classroom-courses-dialog.component';
 import { TeacherRun } from '../teacher-run';
+import { Router } from '@angular/router';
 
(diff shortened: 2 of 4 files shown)


hirokiterashima (changes requested):
Functionality works. I have some comments about the code.

1. Remove ```broadcastRunChanges()``` from create-run-dialog.component.spec.ts

2. Right now ```highlightNewRun()``` is called in ```processRuns()```, which is called in ```getRuns()``` callback and ```updateExistingRun()```, but shouldn't it only be called once, when the page loads, in ngOnInit()? Also, I think the check (to see if new run should be highlighted and if the runId is a valid number) should be done outside of the highlightNewRunId().

```
ngOnInit() {
    this.getRuns();
    this.subscribeToRuns();
    // pseudo-code
    if (check route to see if a new run was added) {
       newRunId = get run Id as number
       highlightNewRun(newRunId) // <- only call with valid run id number parameter
    }
  }

private highlightNewRun(newRunId: number): void {
  const newRun = this.runs.find((run) => {
    return run.id === newRunId;
  });
  newRun.isHighlighted = true;
  // remove the newRunId parameter from the url
  this.router.navigate(['/teacher/home/schedule'], { queryParams: { newRunId: null } });
}
```

hirokiterashima (approved):
LGTM

PR #1082, opened 2023-03-06
fix(News): Only show Discourse news on home page if discourse_url is set

### LINKED ISSUE #1081: Only show Discourse news on home page if discourse_url is set
Right now the home page always shows the Discourse news for WISE Berkeley even if the discourse_url is not set in application.properties.

Handle these situations
- discourse_url is not in application.properties
- discourse_url is in application.properties but has no value
- discourse_url is in application.properties and has a value

geoffreykwan:
## Changes

Only show Discourse news on home page if Discourse property values have been set in the API application.properties.

## Test

Test with
- https://github.com/WISE-Community/WISE-API/pull/221

Test these situations
- when discourse_url and discourse_news_category are not in application.properties, then you should not see the news on the home page
- when discourse_url and discourse_news_category are in application.properties but have no values, then you should not see the news on the home page
- when discourse_url and discourse_news_category are in application.properties and have values, then you should see the news on the home page

Closes #1081

### DIFF
--- a/src/app/home/home.component.ts
+++ b/src/app/home/home.component.ts
@@ -99,15 +99,13 @@ export class HomeComponent implements OnInit {
       if (config != null) {
         this.discourseUrl = this.configService.getDiscourseURL();
         this.discourseNewsCategory = this.configService.getDiscourseNewsCategory();
-        if (
-          this.discourseUrl != null &&
-          this.discourseUrl !== '' &&
-          this.discourseNewsCategory != null &&
-          this.discourseNewsCategory !== ''
-        ) {
-          this.isDiscourseNewsAvailable = true;
-        }
+        this.isDiscourseNewsAvailable =
+          this.isSet(this.discourseUrl) && this.isSet(this.discourseNewsCategory);
       }
     });
+  }
+
+  private isSet(str: string): boolean {
+    return str != null && str !== '';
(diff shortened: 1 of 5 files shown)


src/app/home/home.component.ts
```diff
@@ -87,7 +92,22 @@ export class HomeComponent implements OnInit {
...
+        this.discourseUrl = this.configService.getDiscourseURL();
+        this.discourseNewsCategory = this.configService.getDiscourseNewsCategory();
+        if (
+          this.discourseUrl != null &&
+          this.discourseUrl !== '' &&
+          this.discourseNewsCategory != null &&
+          this.discourseNewsCategory !== ''
+        ) {
+          this.isDiscourseNewsAvailable = true;
+        }
```
hirokiterashima: I think this statement can be written a bit more cleanly, something like this?

```suggestion
        this.isDiscourseNewsAvailable =
          this.isSet(this.discourseUrl) && this.isSet(this.discourseNewsCategory);
          ...
   }
        
        private isSet(str: string): boolean {
          return str != null && str !== '';
        }
```

hirokiterashima (approved):
LGTM.

Added code suggestion inline to reduce ngOnInit()'s complexity.

PR #1087, opened 2023-03-07
refactor(UtilService): Move arraysContainSameValues()

### LINKED ISSUE #1072: refactor(UtilService): Move arraysContainSameValues()

geoffreykwan:
## Changes

Moved arraysContainsSameValues() from UtilService to GraphStudent.

## Test

1. Create a Graph component with "Enable Trials" set to true
2. Preview the step
3. Add some points to the trial
4. Click "+ New Trial"
5. Add some points to the new trial
6. Go to a different step
7. Come back to the Graph step
8. Click the "Select trials to show" drop down and uncheck one of the checkboxes. This should call arraysContainSameValues() and see that the trials checked are different and then call studentDataChanged() which should then get the component state saved.
9. Go to a different step
10. Come back to the Graph step and the trials you have checked should be the same as when you last left

Closes #1072

### DIFF
--- a/src/assets/wise5/components/graph/graph-student/graph-student.component.ts
+++ b/src/assets/wise5/components/graph/graph-student/graph-student.component.ts
@@ -19,7 +19,6 @@ import { GraphContent } from '../GraphContent';
 import { RandomKeyService } from '../../../services/randomKeyService';
 import { copy } from '../../../common/object/object';
 import { convertToPNGFile } from '../../../common/canvas/canvas';
-import { CSVToArray } from '../../../common/array/array';
 
 const Draggable = require('highcharts/modules/draggable-points.js');
 Draggable(Highcharts);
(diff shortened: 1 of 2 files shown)


src/assets/wise5/components/graph/graph-student/graph-student.component.ts
```diff
@@ -2033,6 +2032,30 @@ export class GraphStudent extends ComponentStudent {
...
+   * array2=['abcdefghij', '1234567890']
+   * If these two arrays are passed in as the two arguments, this function
+   * will return true.
+   * Note: This may only work if the elements are strings, numbers or
+   * booleans. If the elements are objects, this function may or may not work.
+   * @param array1 an array of strings, numbers, or booleans
+   * @param array2 an array of strings, numbers, or booleans
+   * @return whether the arrays contain the same values
+   */
+  private arraysContainSameValues(array1: any[], array2: any[]): boolean {
```
hirokiterashima: Looks like array1 and array2 will always be string arrays? If so, change ```any[]``` to ```string[]``` and update the function comments to reflect expected inputs. Otherwise, go ahead and merge.

```suggestion
  /**
   * Check if two string arrays contain the same values. The order of the
   * elements is not compared, only the actual values. This means the elements
   * can be in different orders but still contain the same values.
   * Example:
   * array1=['1234567890', 'abcdefghij']
   * array2=['abcdefghij', '1234567890']
   * If these two arrays are passed in as the two arguments, this function
   * will return true.
   * @param array1 an array of strings
   * @param array2 an array of strings
   * @return whether the arrays contain the same values
   */
  private arraysContainSameValues(array1: string[], array2: string[]): boolean {
```

hirokiterashima (approved):
LGTM. 

See inline comments.

PR #1104, opened 2023-03-11
feat(AT): Directly add component to step if step is empty

### LINKED ISSUE #1100: feat(AT): Directly add component if step is empty
When adding a component to an empty step in the authoring tool, don't ask authors to choose where to put it. The only option is "Insert as First Component". Instead, directly add it to the step.

hirokiterashima:
## Changes
- In add component to step view, if the step has no components, add the chosen component to the step directly. Previously, we were showing a view where they select "insert as first component" instead.
 
## Test
- Adding a component to an empty step directly adds the component
- Adding a component to a non-empty step asks the author to choose where to put the new component

Closes #1100


geoffreykwan (approved):
Looks good.

PR #1125, opened 2023-03-21
fix(Grade By Step): Duplicate student work request

### LINKED ISSUE #1124: fix(Grade By Step): Duplicate student work request
When a step is loaded in the Grade By Step view, we make two identical requests to retrieve the student data for that step when we should only make one. It looks like we're calling this.teacherDataService.retrieveStudentDataForNode() when ngOnInit() and ngOnChanges() are called.

geoffreykwan:
## Changes

Do not retrieve student work on the first ngOnChanges() of the nodeId in NodeGradingViewComponent.

## Test

1. Log in as a teacher
2. Open the Classroom Monitor for a run
3. Go to the "Grade By Step" view
4. Open the browser developer tools
5. Go to the "Network" tab
6. Click on a step to load the student work for that step
7. In the "Network" tab make sure there is only one request made to `https://wise.berkeley.edu/api/teacher/data`. Previously there would be two identical requests.

Closes #1124


hirokiterashima (approved):
LGTM

PR #1130, opened 2023-03-23
feat(Table): Hide reset button when there are no editable cells

### LINKED ISSUE #1122: feat(Table): Only show reset button when there are editable cells
## Description
Reset button shows up on all Table components, even if there is nothing to reset. Hide the reset button when there are no editable cells in the table.

breity:
## Changes
Reset button on Table components only appears for students when the table contains editable cells.

## Test
- Make sure reset button does not appear no cells in Table component are editable.
- Make sure reset button appears and works as before when there are editable cells in component.

Closes #1122.


hirokiterashima (approved):
LGTM

Issue #1135, opened 2023-03-27
feat(Component): Add constraints


hirokiterashima:
### Description

Add support for constraints and removal criteria functions for Components. Use the same content and code (ConstraintService) to author and evaluate constraints as we have for Nodes.

PR #1136, opened 2023-03-27
feat(Component): Evaluate constraints

hirokiterashima:
## Notes
- This is a feature branch PR, so will not go into develop branch yet. We still need to discuss what to save in StudentStatus.
 
## Changes
- Implement evaluate constraints on component, which affects component visibility
- ConstraintService now handles both Node and Component constraint evaluation
- Clean up NodeComponent by replacing ```this.nodeId``` with ```this.node.id``` and removing unused endedAndLockedMessage variable

## Test Setup
- Try setting constraints array on a component with the action, ```makeThisComponentNotVisible```. You should be able to use any removal criteria (isCompleted, choiceChosen, fillXNumberOfRows, etc)
   - It may be easier if you author the constraint on a node first and copy the JSON and modify it.
   - Alternatively, you can import [this unit zip file](https://github.com/WISE-Community/WISE-Client/files/11081232/509.zip) with pre-authored component constraints

## Test
- Component visibility is correctly toggled based on the constraint evaluation result
- Node constraint evaluation works as before

WIP on #1135

### DIFF
--- a/src/app/services/constraintService.spec.ts
+++ b/src/app/services/constraintService.spec.ts
@@ -41,7 +41,7 @@ let dataService: StudentDataService;
 const groupId1 = 'group1';
 const groupId2 = 'group2';
 const groupId3 = 'group3';
-let nodeConstraintTwoRemovalCriteria: any;
+let constraintTwoRemovalCriteria: any;
 const nodeId1 = 'node1';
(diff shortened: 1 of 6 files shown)


geoffreykwan (approved):
Looks good.

PR #1143, opened 2023-03-28
feat(Data Explorer): Tooltip header column

### LINKED ISSUE #1142: feat(Data Explorer): Data point tooltip header show value from specific column
When the student hovers over a data point, the tooltip will display the value from a specified column along with the x and y values. This column should be authorable in the Data Explorer authoring and should be optional.

geoffreykwan:
## Changes

- Allow the author to specify dataExplorerTooltipHeaderColumn. When the student hovers over a data point, the value from this column will be displayed along with the x and y values.
- Data point tooltips no longer display the tooltip header if the header is the same as the y axis title. It will just display the x and y labels and data so that the y label is no longer duplicated like it used to be.

## Test

- Make sure Data Explorer Graph tooltips work with the Tooltip Header Column specified
- Make sure Data Explorer Graph tooltips work with the Tooltip Header Column not specified
- Make sure regular Graph data point tooltips work, particularly with multiple series in a trial

Closes #1142

### DIFF
--- a/src/assets/wise5/components/graph/graph-student/graph-student.component.spec.ts
+++ b/src/assets/wise5/components/graph/graph-student/graph-student.component.spec.ts
@@ -244,9 +244,9 @@ function convertRowDataToSeriesData() {
       };
       const seriesData = component.convertRowDataToSeriesData(rows, params);
       expect(seriesData).toEqual([
-        [0, 0],
-        [10, 5],
-        [20, 10]
+        { x: 0, y: 0 },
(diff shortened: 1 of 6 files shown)


breity (approved):
Functionality works well.

Wondering if maybe we should specify in the authoring info tooltip that the header will only apply to scatterplots. What do you think?

Also, GraphStudentComponent and it's spec file are really long and have a lot of duplication, complex functions. We should probably create a new issue to refactor.

Otherwise, looks good.

PR #1164, opened 2023-04-04
feat(Grading): Look at component statuses to show components

### LINKED ISSUE #1157: feat(Grading): Look at component statuses to determine which components to show

geoffreykwan:
## Changes

Grade by Step and Grade by Team now look at component statuses for a workgroup to determine which components to show in the grading view.

## Test

1. Author a project that has component visibility constraints

For example a step that contains 3 components
  - A Multiple Choice component that has choices A and B
  - A component that is only visible when the student chooses A
  - A component that is only visible when the student chooses B
3. Have students work on the project
4. Go to the Grade by Step view and make sure the correct components are visible or hidden
5. Go to the Grade by Team view and make sure the correct components are visible or hidden

Closes #1157

### DIFF
--- a/src/assets/wise5/classroomMonitor/classroomMonitorComponents/shared/grading-helpers/grading-helpers.ts
+++ b/src/assets/wise5/classroomMonitor/classroomMonitorComponents/shared/grading-helpers/grading-helpers.ts
@@ -1,16 +1,14 @@
-import { ComponentContent } from '../../../../common/ComponentContent';
 import { ComponentStatus } from '../../../../common/ComponentStatus';
 
 export function calculateComponentVisibility(
-  components: ComponentContent[],
   componentIdToHasWork: { [componentId: string]: boolean },
   componentStatuses: { [componentId: string]: ComponentStatus }
 ): { [componentId: string]: boolean } {
   const componentIdToIsVisible = {};
-  for (const component of components) {
-    componentIdToIsVisible[component.id] =
-      componentIdToHasWork[component.id] &&
-      isComponentVisibleToStudent(componentStatuses, component.id);
+  for (const componentId in componentIdToHasWork) {
+    componentIdToIsVisible[componentId] =
+      componentIdToHasWork[componentId] &&
+      isComponentVisibleToStudent(componentStatuses, componentId);
   }
   return componentIdToIsVisible;
 }
--- a/src/assets/wise5/classroomMonitor/classroomMonitorComponents/shared/grading-helpers/grading-helpers.ts
+++ b/src/assets/wise5/classroomMonitor/classroomMonitorComponents/shared/grading-helpers/grading-helpers.ts
@@ -1,15 +1,15 @@
+import { ComponentContent } from '../../../../common/ComponentContent';
 import { ComponentStatus } from '../../../../common/ComponentStatus';
-import { TeacherProjectService } from '../../../../services/teacherProjectService';
 
 export function calculateComponentVisibility(
-  nodeId: string,
-  componentStatuses: { [componentId: string]: ComponentStatus },
-  projectService: TeacherProjectService
+  components: ComponentContent[],
+  componentIdToHasWork: { [componentId: string]: boolean },
+  componentStatuses: { [componentId: string]: ComponentStatus }
 ): { [componentId: string]: boolean } {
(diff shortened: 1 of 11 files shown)


src/assets/wise5/classroomMonitor/classroomMonitorComponents/shared/grading-helpers/grading-helpers.ts
```diff
@@ -0,0 +1,10 @@
+export function isComponentVisibleToStudent(nodeStatus: any, component: any): boolean {
+  let result = true;
+  if (nodeStatus.componentStatuses != null) {
+    const componentStatus = nodeStatus.componentStatuses[component.id];
+    if (componentStatus != null) {
+      result = componentStatus.isVisible;
+    }
+  }
+  return result;
+}
```
hirokiterashima: This function has a lot of checks, which makes it difficult to read. It looks like we don't need the nodeStatus object here. What if we pass in the componentStatuses object with a default empty object? Can you also write a test for this function?
```suggestion
export function isComponentVisibleToStudent(
  componentStatuses: { [componentId: string]: ComponentStatus }  = {},
  componentId: string
): boolean {
  const componentStatus = componentStatuses[componentId];
  return componentStatus == null || componentStatus.isVisible;
}
```

hirokiterashima (changes requested):
It works the first time the step is loaded in the grading views, but if I go to another step and come back, all the components in the step are shown.

The component numberings should match that of the "Showing X/X items" dropdown.

Also, I added a comment about the isComponentVisibleToStudent() function.

src/assets/wise5/classroomMonitor/classroomMonitorComponents/shared/grading-helpers/grading-helpers.ts
```diff
@@ -1,10 +1,24 @@
...
+): { [componentId: string]: boolean } {
+  const componentIdToIsVisible = {};
+  for (const component of projectService.getComponents(nodeId)) {
+    componentIdToIsVisible[component.id] =
+      projectService.componentHasWork(component) &&
+      isComponentVisibleToStudent(componentStatuses, component.id);
   }
-  return result;
+  return componentIdToIsVisible;
+}
```
hirokiterashima: I'm not sure that passing in the entire TeacherProjectService to this function is a good idea, since doing this exposes all of TeacherProjectService's public functions to this function. This function just needs the components and componentsWithWork for the node. What if we re-wrote it to something like this? I think it'll make it easier to test too (remember to write tests)

```
export function calculateComponentVisibility(
  componentStatuses: { [componentId: string]: ComponentStatus },
  components: ComponentContent[],
  componentsWithWork: ComponentContent[]
): { [componentId: string]: boolean } {
```

hirokiterashima (changes requested):
Fixes look good. I added inline comments about the new ```calculateComponentVisibility()``` code.

src/assets/wise5/classroomMonitor/classroomMonitorComponents/shared/grading-helpers/grading-helpers.ts
```diff
@@ -0,0 +1,24 @@
...
+  components: ComponentContent[],
+  componentIdToHasWork: { [componentId: string]: boolean },
+  componentStatuses: { [componentId: string]: ComponentStatus }
+): { [componentId: string]: boolean } {
+  const componentIdToIsVisible = {};
+  for (const component of components) {
+    componentIdToIsVisible[component.id] =
+      componentIdToHasWork[component.id] &&
+      isComponentVisibleToStudent(componentStatuses, component.id);
+  }
```
hirokiterashima: If you use for...in loop to loop through the componentIdToHasWork object, we won't need to pass in the components param to this function?
```suggestion
  for (const componentId in componentIdToHasWork) {
    componentIdToIsVisible[componentId] =
      componentIdToHasWork[componentId] &&
      isComponentVisibleToStudent(componentStatuses, componentId);
  }
```

hirokiterashima (changes requested):
Does calculateComponentVisibility() need to take in an array of components? See inline comments.

hirokiterashima (approved):
LGTM

PR #1170, opened 2023-04-07
refactor(StudentDataService): Move visitedNodesHistory to NodeStatusService

### LINKED ISSUE #1169: refactor(StudentDataService): Move visitedNodesHistory to NodeStatusService
This cache doesn't have much to do with StudentData and should be in NodeStatusService.

hirokiterashima:
## Changes
- Move StudentDataService.visitedNodesHistory to NodeStatusService.nodeIdToIsVisited and convert it from an array into a map of [nodeId, isVisited] values, which is populated/updated when
   - The student data is retrieved (for student and preview) when the VLE launches
   - Student visits a node

## Test
- IsVisited values in NodeStatuses are calculated and saved as before. Check the StudentStatus that is saved each time you visit a step and verify that on the steps that you've visited has isVisited set to true.

Closes #1169

### DIFF
--- a/src/assets/wise5/services/nodeStatusService.ts
+++ b/src/assets/wise5/services/nodeStatusService.ts
@@ -9,6 +9,8 @@ import { CompletionService } from './completionService';
 
 @Injectable()
 export class NodeStatusService {
+  private nodeIdToIsVisited: { [nodeId: string]: boolean } = {};
+
   constructor(
     private completionService: CompletionService,
     private constraintService: ConstraintService,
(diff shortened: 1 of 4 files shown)


src/assets/wise5/services/nodeStatusService.ts
```diff
@@ -25,6 +31,16 @@ export class NodeStatusService {
     });
   }
 
+  private populateNodeIdToIsVisited(events: any[]): void {
+    events
+      .filter((event) => event.event === 'nodeEntered')
+      .forEach((event) => this.setNodeIsVisited(event.nodeId));
+  }
```
geoffreykwan, line 38: This can be done in a single loop.

geoffreykwan (changes requested):
When the student loads a run for the first time, the first step that is loaded has isVisited set to false in the student status even though it should be true. This problem exists on the develop branch so if it's not a simple fix we can fix it in another issue.

hirokiterashima:
Let's address the issue in a separate task. This PR should focus on refactoring.

geoffreykwan (approved):
Looks good.

PR #1175, opened 2023-04-10
fix(Graph): Multiple regression lines

### LINKED ISSUE #1123: bug(Graph): Multiple regression lines are broken
## Description
When a Graph is connected to a Table component using Data Explorer, only the first regression line is drawn correctly. This happens when the graph type is set to scatterplot and multiple y columns are selected in the Data Explorer settings.

geoffreykwan:
## Changes

- Multiple regression lines now use the correct Y axis
- Fixed data point tooltips not showing up when the number of series is greater than the number of Y axes
- Fixed series color not matching Y axis color when Series 1 is set to be on Y Axis 2 and Series 2 is set to be on Y Axis 1
- You can no longer mouse over the regression lines so they no longer get in the way of mousing over data points

## Test

- Make sure all regression lines work properly when the Data Explorer is set to have multiple series and multiple Y Axes. Make sure to set a matching number of Y Axes in the connected Graph component.
- If you have 3 series and 2 Y axes, make sure the tooltips for the data points in the 3rd series work. They used to not show up on mouse over.
- If you set 2 series and 2 Y Axes and set Series 1 to be on Y Axis 2 and Series 2 to be on Y Axis 1, the series colors should match the Y axis colors and label. They used to not match.
- Make sure that when you mouse over a data point that is very close to a regression line, that you can still see the data point tooltip. Previously the regression line would block mousing over data points.
- Make sure regular graph with trials and multiple series have tooltips and colors that still work like before

Closes #1123

### DIFF
--- a/src/assets/wise5/components/graph/graph-student/graph-student.component.spec.ts
+++ b/src/assets/wise5/components/graph/graph-student/graph-student.component.spec.ts
@@ -1887,7 +1887,8 @@ function setYAxisLabelsWhenMultipleYAxes() {
       }
     ];
     const studentData = {
-      dataExplorerYAxisLabels: ['Count', 'Price']
+      dataExplorerYAxisLabels: ['Count', 'Price'],
+      dataExplorerSeries: [{ yAxis: 0 }, { yAxis: 1 }]
     };
     component.setYAxisLabels(studentData);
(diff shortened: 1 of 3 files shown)


breity (approved):
Looks good. 👍

PR #1178, opened 2023-04-11
fix(Node Completion): Look at NodeStatuses to determine completion

hirokiterashima:
## Changes
- Look at NodeStatutes to determine a node's completion, instead of calculating all of its children components' completion.
- This fixes the issue where a node with invisible components were not deemed completed because it was calculating all of its components.

## Test
- Import this unit and verify that the completion constraints unlock properly: [515.zip](https://github.com/WISE-Community/WISE-Client/files/11203578/515.zip). This unit contains two types completion constraints that depend on completion of node with invisible components
- Completion works as before in other cases (unit without invisible components, unit with branches, etc)
- Test existing units in student and teacher modes to make sure that they work as before


geoffreykwan (approved):
Looks good.

PR #1192, opened 2023-04-20
feat(Authoring): When adding a new component, move immediately to it

### LINKED ISSUE #1191: feat(Authoring): When adding a new component, move immediately to it
Right now when an author adds or copies a component, we wait a couple seconds and then animate scroll to the new component. We want to change this to jump immediately to the new component.

geoffreykwan:
## Changes

When adding or copying a component, we scroll immediately to the new component without waiting or animating.

## Test

Make sure the view scrolls to the new component immediately when
- Adding a new component
- Copying a component

Closes #1191

### DIFF
--- a/src/assets/wise5/authoringTool/node/nodeAuthoringComponent.ts
+++ b/src/assets/wise5/authoringTool/node/nodeAuthoringComponent.ts
@@ -494,46 +494,25 @@ class NodeAuthoringController {
    * @param newComponents an array of the new components we have just added
    */
   highlightNewComponentsAndThenShowComponentAuthoring(newComponents) {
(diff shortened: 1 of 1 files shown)


hirokiterashima (commented):
Works well. Can we remove this check? When would it return false?
```
if (newComponents != null && newComponents.length > 0) {
```

geoffreykwan:
newComponents are null when you're moving components. We can clean that up in another issue so that the place where we move components doesn't need to run that code.

hirokiterashima:
I see. Sounds good.

hirokiterashima (approved):

PR #1195, opened 2023-04-20
fix(ProjectService): Filter Dangling Groups

### LINKED ISSUE #1194: bug(ProjectService): Dangling group
When a group is in the project.nodes array but is not used, we need to make sure that it is not returned by call to projectService.getGroupNodes(). This breaks code downstream where we calculate nodeStatus completion.

hirokiterashima:
## Changes
- Filter dangling groups from projectService.groupNodes
- Add test for parseProject() to test for this case with dangling group

## Test
- Unit with dangling group node loads and does not throw error in NodeProgressService.updateStepNodeProgress() when going to another step
- Normal units (without dangling group node) load and work as before

Closes #1194

### DIFF
--- a/src/app/services/projectService.spec.ts
+++ b/src/app/services/projectService.spec.ts
@@ -57,6 +57,7 @@ describe('ProjectService', () => {
   getMaxScoreForComponent();
   getMaxScoreForNode();
   getPeerGrouping();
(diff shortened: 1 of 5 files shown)


geoffreykwan (approved):
Looks good.

PR #1208, opened 2023-04-24
fix(VLE): Step save button now works with Dynamic Components

### LINKED ISSUE #1207: bug(VLE): Step save button does not work with Dynamic Components
1. Create a step that contains a Multiple Choice component and two Open Response components
2. Enable the save button on the step and all components
3. Add two choices to the Multiple Choice component
4. Make the first Open Response only show up if the student chooses the first choice in the Multiple Choice component
5. Make the second Open Response only show up if the student chooses the second choice in the Multiple Choice component
6. Preview the step
7. Select a choice in the Multiple Choice component
8. Click the step save button (the one on the very bottom). The step save button will not perform a save and will not gray out even though it should do both.

geoffreykwan:
## Changes

When the step save button is clicked, only request student work from components that are visible.

## Test

1. Create a step that contains a Multiple Choice component and two Open Response components
2. Enable the save button on the step and all components
3. Add two choices to the Multiple Choice component
4. Make the first Open Response only show up if the student chooses the first choice in the Multiple Choice component
5. Make the second Open Response only show up if the student chooses the second choice in the Multiple Choice component
6. Preview the step
7. Select a choice in the Multiple Choice component
8. Click the step save button (the one on the very bottom). Previously, the step save button would not save and would not gray out. Now it should perform the save and should gray out.

Also make sure component save and submit buttons work as well as the step submit button.

Closes #1207

### DIFF
--- a/src/assets/wise5/vle/node/node.component.ts
+++ b/src/assets/wise5/vle/node/node.component.ts
@@ -369,7 +369,12 @@ export class NodeComponent implements OnInit {
       }
       return components;
     } else {
-      return this.getComponents();
+      const nodeStatus = this.studentDataService.getNodeStatusByNodeId(this.node.id);
+      return this.getComponents().filter(
+        (component) =>
+          this.workComponents.includes(component.type) &&
+          nodeStatus.componentStatuses[component.id].isVisible
+      );
     }
   }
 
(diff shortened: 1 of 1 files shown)


hirokiterashima (approved):
LGTM

PR #1227, opened 2023-05-03
feat(Constraint Removal Authoring): Filter MC components 

### LINKED ISSUE #1147: feat(Choice Chosen Constraint Authoring): Only show MC steps and components as available options
Choice Chosen constraint removal criteria authoring should only show MC steps and components as available options. Right now it shows all steps and components.

hirokiterashima:
## Changes
- Only show MC steps and components as available options if author choses the "Choice Chosen" removal criteria

## Test
- In constraint removal authoring (for both node and component)
   - if you choose "Choice chosen" for Removal Criteria Name, it should only show steps that contain MC components, and components that are MC.
   - if you choose any other option for Removal Criteria Name, it should show all steps and components like before

Closes #1147

### DIFF
--- a/src/assets/wise5/authoringTool/constraint/edit-constraint-removal-criteria/edit-constraint-removal-criteria.component.html
+++ b/src/assets/wise5/authoringTool/constraint/edit-constraint-removal-criteria/edit-constraint-removal-criteria.component.html
@@ -59,15 +59,15 @@
         <mat-option [value]="''">
           <span class="red" i18n>Please Choose a Component</span>
         </mat-option>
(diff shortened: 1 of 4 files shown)


geoffreykwan (changes requested):
Can we do this without using function calls in the template?

hirokiterashima:
How would we do that? The nodeId is selected by the author (i.e. dynamic), so wouldn't we have to then call a function to get the components in for that selected node?

geoffreykwan (approved):
Looks good.

PR #1228, opened 2023-05-03
refactor(UtilService): Move calculateMean()

### LINKED ISSUE #1162: refactor(UtilService): Move calculateMean()
This function is only used in GraphStudentComponent, so move it there as a private function.

breity:
## Changes
Moved `calculateMean()` to GraphStudentComponent, since it is only used there.

## Test
Make sure generating regression lines for Graph component scatterplots still work as before.

Closes #1162.

### DIFF
--- a/src/app/services/utilService.spec.ts
+++ b/src/app/services/utilService.spec.ts
@@ -9,19 +9,4 @@ describe('UtilService', () => {
     });
     service = TestBed.inject(UtilService);
   });
-  calculateMeanTests();
 });
-
-function calculateMeanTests() {
-  describe('calculateMean()', () => {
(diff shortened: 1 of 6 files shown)


hirokiterashima (changes requested):
Works well. 

Since calculateMean() doesn't require any interaction with GraphStudentComponent's data or functions it doesn't need to be defined in the class. I see it more like the copy() function, in that it's a utility function, and belongs to no one class. Can you move it outside of the GraphStudentComponent class, like at the bottom of the graph-student-component.ts file as a standalone function, or in a separate file in the same directory?

hirokiterashima (approved):
LGTM

PR #1234, opened 2023-05-05
fix(Manage Students): Drag student to workgroup puts wrong student into workgroup

### LINKED ISSUE #1219: bug(Manage Students): Drag student to workgroup puts wrong student into workgroup
1. Create a run
2. Have 4 student accounts add the run but do not click "Launch". For example aa0101, bb0101, cc0101, dd0101.
3. Go to the Manage Students view for the run. You should see 4 students in the "Students without a team".
4. Click "New Team" and then add the first student aa0101 and click "Create".
5. Choose a student that is not the first student in "Students without a team". In my case bb0101 is the first student in my "Students without a team" so I will choose cc0101 or dd0101.
- If I drag cc0101 into the team and click "Proceed", it will add dd0101 to the team instead
- If I drag dd0101 into the team and click "Proceed", it will add cc0101 to the team instead
- If I drag bb0101 into the team and click "Proceed", it will add bb0101 to the team correctly

hirokiterashima:
## Changes
- Fix issue where dragging student to workgroup was putting the wrong student into workgroup. I did this by adding the user object in the dragged item's data and using that instead of looking up the user by index.
- Code clean up

## Test Fix
1. Create a run
2. Have 4 student accounts add the run but do not click "Launch". For example aa0101, bb0101, cc0101, dd0101.
3. Go to the Manage Students view for the run. You should see 4 students in the "Students without a team".
4. Click "New Team" and then add the first student aa0101 and click "Create".
5. Choose a student that is not the first student in "Students without a team". In my case bb0101 is the first student in my "Students without a team" so I will choose cc0101 or dd0101.
- If I drag cc0101 into the team and click "Proceed", it will add dd0101 to the team instead
- If I drag dd0101 into the team and click "Proceed", it will add cc0101 to the team instead
- If I drag bb0101 into the team and click "Proceed", it will add bb0101 to the team correctly

## Test Regression
- Moving students between teams and creating new team work as before

Closes #1219

### DIFF
--- a/src/assets/wise5/classroomMonitor/classroomMonitorComponents/manageStudents/manage-team/manage-team.component.html
+++ b/src/assets/wise5/classroomMonitor/classroomMonitorComponents/manageStudents/manage-team/manage-team.component.html
@@ -34,7 +34,7 @@
       <ng-container *ngFor="let user of team.users">
         <li
           cdkDrag
-          [cdkDragData]="team.workgroupId"
+          [cdkDragData]="{ user: user, workgroupId: team.workgroupId }"
           (cdkDragEntered)="dragEnter($event)"
           (cdkDragExited)="dragExit($event)"
           [cdkDragPreviewContainer]="'parent'"
--- a/src/assets/wise5/classroomMonitor/classroomMonitorComponents/manageStudents/manage-team/manage-team.component.ts
+++ b/src/assets/wise5/classroomMonitor/classroomMonitorComponents/manageStudents/manage-team/manage-team.component.ts
@@ -94,10 +94,13 @@ export class ManageTeamComponent {
         this.team.periodId
       )
       .subscribe(() => {
+        const previousIndex = event.previousContainer.data.findIndex(
+          (user) => user === event.item.data.user
+        );
         transferArrayItem(
           event.previousContainer.data,
           event.container.data,
-          event.previousIndex,
+          previousIndex,
           event.currentIndex
         );
         this.configService.retrieveConfig(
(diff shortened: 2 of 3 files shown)


geoffreykwan:
Functionality works. There still seems to be something off about the drop previousIndex but it might be a bug in the CdkDragDrop. When I drop a student, I see another student's name quickly flash but then it changes back to the correct student name. I'm wondering if we should apply this change I found here.

https://github.com/angular/components/issues/14873#issuecomment-863800685

```
        const previousIndex = event.previousContainer.data.findIndex(
          (user) => user === event.item.data.user
        );
        transferArrayItem(
          event.previousContainer.data,
          event.container.data,
          previousIndex,
          event.currentIndex
        );
```

hirokiterashima:
I don't see this behavior. Is there a breakpoint that I can set to see this intermediate state? I tried breaking in moveUser() but it worked for me.

geoffreykwan:
I don't know if there's a place to break point that will stop in that state. If you perform the steps to reproduce the original problem (where the wrong student is placed in the workgroup) you should be able to see the flash. You might be able to use the code below to detect when this will happen. Put this code at the top of the moveUser() function. When it hits the breakpoint, focus your eyes on the workgroup in the Manage Students view and then click the debugger resume button and then you should see it quickly flash with the wrong student name but then quickly switch to the correct student name.

```
const previousIndex = event.previousContainer.data.findIndex((user) => user === event.item.data.user);
if (event.previousIndex !== previousIndex) {
  console.log('Previous index is incorrect');
  debugger; // set breakpoint here
}
```

hirokiterashima:
Weird. I added the code, but it doesn't break at the debugger line for me (it never goes in the if case). I tried all the workgroup combinations that used to cause the original issue.

hirokiterashima:
PTAL

geoffreykwan (approved):
Looks good.

PR #1236, opened 2023-05-05
feat(Match): Reuse choices

### LINKED ISSUE #1248: feat(Match): Reuse Choices
Allow users to reuse choices multiple times from the source bucket.

- Add choiceReuseEnabled: boolean to MatchContent
- Add Match standard authoring option to enable/disable choiceReuseEnabled.
- If choiceReuseEnabled is true, allow student to move choice into multiple buckets, but you can't have the same choice multiple times in the same bucket
   - Source bucket -> target bucket: make a copy of the choice and move it
   - target bucket -> target bucket: move the choice
   - target bucket -> source bucket: move the choice out from target and remove it once it's in the source bucket (to ensure there's only 1 copy of the choice in the source bucket)

hirokiterashima:
## Changes
- Add choiceReuseEnabled: boolean to MatchContent
- Add Match standard authoring option to enable/disable choiceReuseEnabled.
- If choiceReuseEnabled is true, allow student to move choice into multiple buckets, but you can't have the same choice multiple times in the same bucket

## Test
- Prep: Author Match with choiceReuseEnabled=true
- You can move choice into multiple buckets, but you can't have the same choice multiple times in the same bucket
- Correctness
- Leave step and come back: should show previous choices and allow you to continue to make changes.
- Grading shows student responses

## Test Regression
- Exiting Match components (choiceReuseEnabled=false) work as before

Closes #1248


geoffreykwan (approved):
Looks good.

breity (approved):
LGTM.

PR #1238, opened 2023-05-05
feat(Match): Create Choice class

hirokiterashima:
## Changes
- Create Choice class
- Move createChoiceFromNotebookItem() out of MatchStudentComponent and add tests

## Test Regression
- Match works as before, in particular using private notes as choices


geoffreykwan (approved):
Looks good.

PR #1242, opened 2023-05-09
refactor(MatchStudent): Split to MatchStudentDefault and MatchStudentChoiceReuse

hirokiterashima:
## Changes
- Split MatchStudent to MatchStudentDefault and MatchStudentChoiceReuse
- MatchStudentDefault reverted drop() function to original behavior MatchStudentChoiceReuse overrides drop() function to allow reusing choice
- Add choiceReuseEnabled field to MatchContent

## Test MatchStudentChoiceReuse
Prep: using component JSON authoring, add ```choiceReuseEnabled: true``` to a match component

- A choice can be in multiple buckets. Each choice can be in the same bucket only once
- Show Student Work correctly shows students' work

## Test MatchStudentDefault
This is the existing MatchStudent, where a choice can only exist in one bucket. Make sure it works as before


geoffreykwan (approved):
Looks good.

PR #1243, opened 2023-05-09
feat(Match): Add authoring option to enable choice reuse

hirokiterashima:
## Changes
- Add authoring option to enable choice reuse
- Lower-case variable names in MatchAuthoringComponent

## Test
- Create new match component. It should have ```choiceReuseEnabled``` set to false
- Enable/Disable the ```choiceReuseEnabled``` option by checking/unchecking the option "The student can put each choice in multiple buckets". Preview the component and verify that it lets you reuse choices/not.

### DIFF
--- a/src/assets/wise5/components/match/match-authoring/match-authoring.component.html
+++ b/src/assets/wise5/components/match/match-authoring/match-authoring.component.html
@@ -24,6 +24,17 @@
     Show Choices After Target Buckets
   </mat-checkbox>
 </p>
+<p>
+  <mat-checkbox
+    color="primary"
+    class="checkbox"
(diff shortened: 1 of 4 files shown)


geoffreykwan (approved):
Looks good. The label for the checkbox should match the uppercasing like the other checkbox labels.

Show Source and Target Buckets Side-by-Side
Show Choices After Target Buckets
The student can put each choice in multiple buckets

I also don't think we need the word "The". Maybe change it to "Student Can Put Each Choice In Multiple Buckets".

hirokiterashima:
Changed to: “Students can put each choice in multiple buckets” as discussed during meeting

PR #1254, opened 2023-05-11
feat(Match): Implement component export for choice reuse

### LINKED ISSUE #1251: feat(Match): Implement component export for choice reuse
If a choice is in multiple buckets, have the cell contain the list of all the buckets it is in, separated by a comma.
If there is correctness, merge the correctness values for a choice. For example, if a choice is placed in two buckets and it's correct in the first bucket but incorrect in the second bucket, mark the choice as incorrect.

geoffreykwan:
## Changes

Match component export now properly exports data when a Match component can reuse choices.

## Test

- Create a Match component with choice reuse enabled as well as correctness and correct positions
- Generate some work for that Match component with choice reuse enabled
- Generate the Match component export
  - Classroom Monitor > Data Export view > Export Component Data > Find the Match component and click "All" button
- In the export, check
  - If a choice is in multiple buckets, those bucket names should show up in the cell separated by commas
  - Make sure the correctness value is correct when a choice is placed in multiple buckets based on the matrix below. The previous and next is in reference to when we loop through the buckets to check for correctness.
Examples
    - If choice "Apple" was placed in bucket "A", which is correct, and also placed in bucket "Wildcard", which is also correct, then we would say the previous value was 1 (because it was correct in bucket "A") and the new value is 1 (because it was correct in bucket "Wildcard") and the result correctness value that is shown in the export for the choice "Apple" is 1.
    - If choice "Apple" was placed in bucket "A", which is correct, and also placed in bucket "B", which is incorrect, then we would say the previous value was 1 (because it was correct in bucket "A") and the new value is 0 (because it was incorrect in bucket "B") and the result correctness value that is shown in the export for the choice "Apple" is 0.
```
e = empty
0 = incorrect
1 = correct
2 = correct bucket but incorrect position

       Previous Value
       e 0 1 2
      --------
N  0 | 0 0 0 0
e  1 | 1 0 1 2
w  2 | 2 0 2 2

V
a
l
u
e
```
- Make sure Match component export without choice reuse still works properly

Closes #1251

### DIFF
--- a/src/assets/wise5/classroomMonitor/dataExport/data-export/data-export.component.ts
+++ b/src/assets/wise5/classroomMonitor/dataExport/data-export/data-export.component.ts
@@ -23,6 +23,8 @@ import { LabelComponentDataExportStrategy } from '../strategies/LabelComponentDa
 import { Component as WISEComponent } from '../../../common/Component';
 import { removeHTMLTags } from '../../../common/string/string';
 import { millisecondsToDateTime } from '../../../common/datetime/datetime';
+import { Choice } from '../../../components/match/choice';
+import { Bucket } from '../../../components/match/bucket';
 
 @Component({
   selector: 'data-export',
@@ -1171,11 +1173,7 @@ export class DataExportComponent implements OnInit {
     componentRevisionCounter: any,
     matchComponentState: any
   ): string[] {
-    /*
-     * Populate the cells in the row that contain the information about the
-     * student, project, run, step, and component.
-     */
-    let row = this.createStudentWorkExportRow(
+    const row = this.createStudentWorkExportRow(
       columnNames,
       columnNameToNumber,
       rowCounter,
@@ -1190,8 +1188,22 @@ export class DataExportComponent implements OnInit {
       componentRevisionCounter,
       matchComponentState
     );
-    for (const bucket of matchComponentState.studentData.buckets) {
-      for (const item of bucket.items) {
+    if (component.choiceReuseEnabled) {
+      this.insertMatchChoiceReuseValues(row, columnNameToNumber, component, matchComponentState);
+    } else {
+      this.insertMatchDefaultValues(row, columnNameToNumber, component, matchComponentState);
+    }
+    return row;
+  }
+
+  private insertMatchDefaultValues(
+    row: string[],
+    columnNameToNumber: any,
+    component: any,
+    matchComponentState: any
+  ): void {
+    matchComponentState.studentData.buckets.forEach((bucket: Bucket) => {
+      bucket.items.forEach((item: Choice) => {
         row[columnNameToNumber[item.id]] = this.getBucketValueById(component, bucket.id);
         if (
           this.includeCorrectnessColumns &&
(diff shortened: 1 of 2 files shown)


src/assets/wise5/classroomMonitor/dataExport/data-export/data-export.component.ts
```diff
@@ -1190,18 +1188,127 @@ export class DataExportComponent implements OnInit {
...
+   * @param row: any[]
+   * @param columnNameToNumber: any
+   * @param columnName: string
+   * @param newValue: number
+   */
+  private mergeCorrectnessValue(
+    row: any,
+    columnNameToNumber: any,
+    columnName: string,
+    newValue: any
```
hirokiterashima:
```suggestion
    newValue: number
```

src/assets/wise5/classroomMonitor/dataExport/data-export/data-export.component.ts
```diff
@@ -1190,18 +1188,127 @@ export class DataExportComponent implements OnInit {
...
+        });
+      });
+  }
+
+  private setBucketValueWithAppending(
+    row: string[],
+    columnNameToNumber: any,
+    component: any,
+    bucket: any,
+    item: any
```
hirokiterashima:
```suggestion
    bucket: Bucket,
    item: Choice
```

src/assets/wise5/classroomMonitor/dataExport/data-export/data-export.component.ts
```diff
@@ -1190,18 +1188,127 @@ export class DataExportComponent implements OnInit {
...
+    if (row[columnNameToNumber[item.id]] == null || row[columnNameToNumber[item.id]] === '') {
+      row[columnNameToNumber[item.id]] = this.getBucketValueById(component, bucket.id);
+    } else {
+      row[columnNameToNumber[item.id]] = `${
+        row[columnNameToNumber[item.id]]
+      }, ${this.getBucketValueById(component, bucket.id)}`;
+    }
+  }
+
+  private setCorrectnessValueForChoiceReuse(row: any[], columnNameToNumber: any, item: any): void {
```
hirokiterashima:
```suggestion
  private setCorrectnessValueForChoiceReuse(row: any[], columnNameToNumber: any, item: Choice): void {
```

src/assets/wise5/classroomMonitor/dataExport/data-export/data-export.component.ts
```diff
@@ -1190,18 +1188,127 @@ export class DataExportComponent implements OnInit {
...
+    columnNameToNumber: any,
+    columnName: string,
+    newValue: any
+  ): void {
+    const previousValue = row[columnNameToNumber[columnName]];
+    if (previousValue === '') {
+      row[columnNameToNumber[columnName]] = newValue;
+    } else if (this.bothValuesAreX(previousValue, newValue, 1)) {
+      row[columnNameToNumber[columnName]] = 1;
+    } else if (this.eitherValuesAreX(previousValue, newValue, 0)) {
```
hirokiterashima: Do this check inline and remove extra function call?
```suggestion
    } else if ([previousValue, newValue].includes(0))) {
```

src/assets/wise5/classroomMonitor/dataExport/data-export/data-export.component.ts
```diff
@@ -1190,18 +1188,127 @@ export class DataExportComponent implements OnInit {
...
+      row[columnNameToNumber[columnName]] = 0;
+    } else if (this.eitherValuesAreX(previousValue, newValue, 2)) {
```
hirokiterashima: Do this check inline and remove extra function call?
```suggestion
    } else if ([previousValue, newValue].includes(2))) {
```

src/assets/wise5/classroomMonitor/dataExport/data-export/data-export.component.ts
```diff
@@ -1190,18 +1188,127 @@ export class DataExportComponent implements OnInit {
...
+    }
+  }
+
+  private bothValuesAreX(value1: number, value2: number, expectedValue: number): boolean {
+    return value1 === expectedValue && value2 === expectedValue;
+  }
+
+  private eitherValuesAreX(value1: number, value2: number, expectedValue: number): boolean {
+    return value1 === expectedValue || value2 === expectedValue;
   }
```
hirokiterashima: Remove function?
```suggestion
```

src/assets/wise5/components/match/bucket.ts
```diff
@@ -0,0 +1,12 @@
+import { Choice } from './choice';
+
+export class Bucket {
+  id: string;
+  items?: Choice[];
```
hirokiterashima, line 5: Are there buckets that will never have any items?
```suggestion
  items: Choice[];
```

src/assets/wise5/classroomMonitor/dataExport/data-export/data-export.component.ts
```diff
@@ -1190,18 +1188,127 @@ export class DataExportComponent implements OnInit {
...
+    bucket: any,
+    item: any
+  ): void {
+    if (row[columnNameToNumber[item.id]] == null || row[columnNameToNumber[item.id]] === '') {
+      row[columnNameToNumber[item.id]] = this.getBucketValueById(component, bucket.id);
+    } else {
+      row[columnNameToNumber[item.id]] = `${
+        row[columnNameToNumber[item.id]]
+      }, ${this.getBucketValueById(component, bucket.id)}`;
+    }
```
hirokiterashima: Use ternary expression to simplify block
```suggestion
    row[columnNameToNumber[item.id]] =
      row[columnNameToNumber[item.id]] == null || row[columnNameToNumber[item.id]] === '' ?
        this.getBucketValueById(component, bucket.id) :
        `${row[columnNameToNumber[item.id]]}, ${this.getBucketValueById(component, bucket.id)}`; 
```

src/assets/wise5/classroomMonitor/dataExport/data-export/data-export.component.ts
```diff
@@ -1190,18 +1188,127 @@ export class DataExportComponent implements OnInit {
...
+    } else if (this.eitherValuesAreX(previousValue, newValue, 0)) {
+      row[columnNameToNumber[columnName]] = 0;
+    } else if (this.eitherValuesAreX(previousValue, newValue, 2)) {
+      row[columnNameToNumber[columnName]] = 2;
+    }
+  }
+
+  private bothValuesAreX(value1: number, value2: number, expectedValue: number): boolean {
+    return value1 === expectedValue && value2 === expectedValue;
+  }
```
hirokiterashima: Make this into a specific function to test if both values are 1 (correct) instead?
```suggestion
  private bothValuesAreCorrect(value1: number, value2: number): boolean {
    return value1 === 1 && value2 === 1;
  }
```

src/assets/wise5/classroomMonitor/dataExport/data-export/data-export.component.ts
```diff
@@ -1190,18 +1188,127 @@ export class DataExportComponent implements OnInit {
...
+  private mergeCorrectnessValue(
+    row: any,
+    columnNameToNumber: any,
+    columnName: string,
+    newValue: any
+  ): void {
+    const previousValue = row[columnNameToNumber[columnName]];
+    if (previousValue === '') {
+      row[columnNameToNumber[columnName]] = newValue;
+    } else if (this.bothValuesAreX(previousValue, newValue, 1)) {
```
hirokiterashima: Make this check more clear by stating that we're checking if both are correct?
```suggestion
    } else if (this.bothValuesAreCorrect(previousValue, newValue)) {
```

src/assets/wise5/classroomMonitor/dataExport/data-export/data-export.component.ts
hirokiterashima:
```suggestion
    newValue: number
```

src/assets/wise5/classroomMonitor/dataExport/data-export/data-export.component.ts
```diff
@@ -1190,18 +1188,127 @@ export class DataExportComponent implements OnInit {
...
+            this.includeCorrectnessColumns &&
+            this.matchService.componentHasCorrectAnswer(component)
+          ) {
+            this.setCorrectnessValueForChoiceReuse(row, columnNameToNumber, item);
+          }
+        });
+      });
+  }
+
+  private setBucketValueWithAppending(
```
hirokiterashima: This "WithAppending" part of the function name is a bit confusing. How about just ```setBucketValue()```?
```suggestion
  private setBucketValue(
```

hirokiterashima (changes requested):
Export seems to work as described.

I added some code comments inline.

src/assets/wise5/components/match/bucket.ts
geoffreykwan: The bucket object in the authored content does not have items.

hirokiterashima (approved):
LGTM

PR #1258, opened 2023-05-15
feat(FeedbackRule): accumulatedIdeaCount[LessThan|Equals|MoreThan] tokens

### LINKED ISSUE #1257: feat(FeedbackRule): New accumulatedIdeaCount[LessThan|Equals|MoreThan] token
Add new tokens that can be used in the FeedbackRules:
1. accumulatedIdeaCountLessThan(3): returns true if they had less than 3 ideas across all (both past and current) responses
2. accumulatedIdeaCountEquals(3): returns true if they had exactly 3 ideas across all (both past and current) responses
3. accumulatedIdeaCountMoreThan(3): returns true if they had more than 3 ideas across all (both past and current) responses

hirokiterashima:
## Changes
- Add support for accumulatedIdeaCount[LessThan|Equals|MoreThan] tokens, (**Note: currently only works in the DialogGuidance component** We need to do a bit more extra work to get it to work in OpenResponse feedback and DynamicPrompt because we don't store c-rater response in those cases that's needed to compute)
- FeedbackRuleEvaluator now takes in an array of CRaterResponses instead of just the latest
- DialogGuidanceStudent now passes in all CRaterResponses to the FeedbackRuleEvaluator
- OpenResponse and DynamicPrompt now passes in an array, but it only contains the latest CRaterResponse. So accumulatedIdeaCount tokens won't work in these places
- Clean up code
   - Define CRaterResponses interface
   - Abstract common idea count evaluation code to AbstractIdeaCountTermEvaluator

## Test accumulatedIdeaCount[LessThan|Equals|MoreThan] tokens
- Add accumulatedIdeaCount[LessThan|Equals|MoreThan] tokens and see that they are matched
   - accumulatedIdeaCountLessThan(3): returns true if they had less than 3 ideas across all (both past and current) responses
   - accumulatedIdeaCountEquals(3): returns true if they had exactly 3 ideas across all (both past and current) responses
   - accumulatedIdeaCountMoreThan(3): returns true if they had more than 3 ideas across all (both past and current) responses
- Feedback Rule help popup in the AT shows info about the new tokens 

## Test Regression
Feedback rule evaluation should work as before in these places. Test existing items and make sure that they still work.
- Single Student
  - Dialog Guidance 
  - OpenResponse annotation using Feedback Rules
  - Dynamic Prompt 
- Multiple Students
   - Dynamic Prompt
   - Question Bank

Closes #1257

### DIFF
--- a/src/assets/wise5/components/common/cRater/CRaterResponses.ts
+++ b/src/assets/wise5/components/common/cRater/CRaterResponses.ts
@@ -1,0 +1,12 @@
+/**
+ * CRaterResponses is the response object from the CRater API.
+ * This is different from CRaterResponse, which is the object that WISE creates to store the
+ * response.
+ */
+export interface CRaterResponses {
+  advisories: string[];
+  feedback: any;
+  response_id: string;
+  scores?: any;
+  trait_scores: any;
+}
--- a/src/assets/wise5/components/common/feedbackRule/feedback-rule-help/feedback-rule-help.component.html
+++ b/src/assets/wise5/components/common/feedbackRule/feedback-rule-help/feedback-rule-help.component.html
@@ -1,3 +1,9 @@
+<ng-template #accumulatedIdeaCountSupportMessage
+  ><i i18n>
+    Note: this term currently works in the Dialog Guidance component only. Support in the Open
+    Response component and the Dynamic Prompt will be coming soon.</i
+  ></ng-template
+>
 <h2 class="mat-dialog-title" i18n>Feedback Rule Authoring</h2>
 <mat-dialog-content class="dialog-content-scroll">
   <h3 i18n>Introduction</h3>
@@ -35,7 +41,8 @@
         Evaluates to true if more than X ideas were found in all (both past and current) of
         student's responses.<br />
         Ex: <b>accumulatedIdeaCountMoreThan(2)</b> evaluates to true if the student had more than 2
-        ideas in all of their responses.
+        ideas in all of their responses.<br />
+        <ng-template [ngTemplateOutlet]="accumulatedIdeaCountSupportMessage"></ng-template>
       </td>
     </tr>
     <tr>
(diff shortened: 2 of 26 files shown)


src/assets/wise5/services/studentDataService.ts
```diff
@@ -8,6 +8,8 @@ import { HttpClient, HttpParams } from '@angular/common/http';
 import { Observable, Subject } from 'rxjs';
 import { DataService } from '../../../app/services/data.service';
 import { generateRandomKey } from '../common/string/string';
+import { CRaterResponse } from '../components/common/cRater/CRaterResponse';
+import { DialogResponse } from '../components/dialogGuidance/DialogResponse';
```
geoffreykwan: We can remove these unused imports.

src/assets/wise5/components/common/cRater/CRaterResponses.ts
```diff
@@ -0,0 +1,12 @@
...
+ * This is different from CRaterResponse, which is the object that WISE creates to store the
+ * response.
+ */
+export interface CRaterResponses {
+  advisories: string[];
+  feedback: any;
+  response_id: string;
+  scores?: any;
+  trait_scores: any;
+}
```
geoffreykwan: Having this interface name be plural is slightly confusing. Would it be better to name it RawCRaterResponse or CRaterRawResponse or something like that?

src/assets/wise5/components/common/feedbackRule/feedback-rule-help/feedback-rule-help.component.html
```diff
@@ -29,6 +29,33 @@ <h3 i18n>Terms</h3>
...
+    </tr>
+    <tr>
+      <td class="term">accumulatedIdeaCountEquals(X)</td>
+      <td i18n>
+        Evaluates to true if exactly X ideas were found in all (both past and current) of student's
+        responses.<br />
+        Ex: <b>accumulatedIdeaCountEquals(2)</b> evaluates to true if the student had exactly 2
+        ideas in all of their responses.
+      </td>
+    </tr>
```
geoffreykwan: This shows up in the Help section for Open Response and Dynamic Prompt even though it doesn't work in those places. Should we have it not show up in those places or add a footnote or something?

geoffreykwan (changes requested):
Functionality works well.

hirokiterashima:
@geoffreykwan made changes. PTAL

geoffreykwan (approved):
Looks good.

I think the Note wording might be a little better if it was
"Note: this term currently only works in the Dialog Guidance component."

breity (approved):
Looks good. 👍

PR #1261, opened 2023-05-16
feat(FeedbackRule): idea count token with response index parameter

### LINKED ISSUE #1260: feat(FeedbackRule): idea count rule with response index parameter
Add a new response index parameter to the ideaCount[MoreThan|LessThan|Equals] token
1. ideaCountLessThan(3, 2): returns true if they had less than 3 ideas in their second response
2. ideaCountLessThan(3, 2): returns true if they had exactly 3 ideas in their second response
3. ideaCountLessThan(3, 2): returns true if they had more than 3 ideas in their second response

hirokiterashima:
## Changes
- Add support to pass in a second parameter to ideaCount[LessThan|Equals|MoreThan] tokens for specifying the response index number, starting at 1.
   - Syntax: ```ideaCount[LessThan|Equals|MoreThan](detectedIdeas: number, responseIndex: number)``` 
   - **Note: the token currently only works in the DialogGuidance component.** We need to do a bit more extra work to get it to work in OpenResponse feedback and DynamicPrompt because we don't store c-rater response in those cases that's needed to compute

## Test ideaCount[LessThan|Equals|MoreThan] tokens with second parameter
- Add various ideaCount[LessThan|Equals|MoreThan] tokens with a second parameter and see that they are matched
   - ideaCountLessThan(3, 2): returns true if they had less than 3 ideas in their second response
   - ideaCountLessThan(3, 2): returns true if they had exactly 3 ideas in their second response
   - ideaCountLessThan(3, 2): returns true if they had more than 3 ideas in their second response
- Feedback Rule help popup in the AT shows info about the new tokens 

## Test Regression
Feedback rule evaluation should work as before in these places. Test existing items and make sure that they still work.
- Single Student
  - Dialog Guidance 
  - OpenResponse annotation using Feedback Rules
  - Dynamic Prompt 
- Multiple Students
   - Dynamic Prompt
   - Question Bank

Closes #1260

### DIFF
--- a/src/assets/wise5/components/common/feedbackRule/FeedbackRuleEvaluator.ts
+++ b/src/assets/wise5/components/common/feedbackRule/FeedbackRuleEvaluator.ts
@@ -114,7 +114,7 @@ export class FeedbackRuleEvaluator<T extends CRaterResponse[]> {
 
   protected evaluateTerm(term: string, responses: T): boolean {
     const evaluator: TermEvaluator = this.factory.getTermEvaluator(term);
-    return TermEvaluator.isAccumulatedIdeaCountTerm(term)
+    return TermEvaluator.requiresAllResponses(term)
       ? evaluator.evaluate(responses)
       : evaluator.evaluate(responses[responses.length - 1]);
   }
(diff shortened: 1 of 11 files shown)


geoffreykwan (approved):
Looks good.

The description for the terms have a mix of number literals and number words. It would be good to just use one or the other.

Current
"if the student had less than 2 ideas in their third response"

Only number literals
"if the student had less than 2 ideas in their 3rd response"

Only number words
"if the student had less than two ideas in their third response"

hirokiterashima:
I fixed the issue where the feedback was unexpected when computer avatar is enabled and initial response is authored. @geoffreykwan PTAL.

geoffreykwan:
Looks good.

PR #1265, opened 2023-05-18
refactor(Graph): Move plot line code to its own file

### LINKED ISSUE #1264: refactor(Graph): Move plot line code to its own file

geoffreykwan:
## Changes

Moved plot line code to plot-line-manager.ts.

## Test

- Make sure "X Axis Plot Line" and "Y Axis Plot Line" show up (author this in Graph advanced authoring)
- Make sure "Show Mouse X Plot Line" and "Show Mouse Y Plot Line" works (author this in Graph advanced authoring)
- Make sure basic Graph functionality still works

Closes #1264


hirokiterashima (commented):
What if, instead of defining PlotLineManager.removeAndAddPlotLine(), the PlotLine has a reference to the axis and defines its own removeAndAdd() function?

```
export class PlotLine {
  axis: any;
  ...
  constructor(id: string, axis: any, value: number, text: string) {
  ...
  removeAndAdd(): void {
    this.axis.removePlotLine(this.id);
    this.axis.addPlotLine(this);
  }
```

Then, anywhere that calls PlotLineManager.showXPlotLine() can be replaced with
```
    new XPlotLine(GET_X_AXIS, x, text).removeAndAdd();
```
and anywhere that calls PlotLineManager.showYPlotLine()  can be replaced with
```
    new YPlotLine(GET_Y_AXIS, y, text).removeAndAdd();
```

This would remove the need to define and test PlotLineManager.showXPlotLine() and PlotLineManager.showYPlotLine(), which is where the code duplication is happening right now. You would only need to write test for PlotLine.removeAndAdd()

geoffreykwan:
I removed this code because we don't need it anymore. Test to make sure the plot line labels still work.

```
  applyHighchartsPlotLinesLabelFix() {
    Highcharts.wrap(Highcharts.Axis.prototype, 'getPlotLinePath', function (proceed) {
      var path = proceed.apply(this, Array.prototype.slice.call(arguments, 1));
      if (path) {
        path.flat = false;
      }
      return path;
    });
  }
```

hirokiterashima (approved):
LGTM

PR #1269, opened 2023-05-18
fix(Authoring): Can't load step you were previously on

### LINKED ISSUE #1268: bug(Authoring): Can't load step you were previously on
2. Click on a step to go into it
3. Click the "Unit Home" button at the upper left to go back to the project view
4. Click on the same step you were previously on. You would not be taken into the step but you should.

geoffreykwan:
## Changes

Call teacherDataService.setCurrentNode(null) in node-authoring.component.ts ngOnDestroy() if the current node id is the same as this node id.

## Test

2. Click on a step to go into it
3. Click the "Unit Home" button at the upper left to go back to the project view
4. Click on the same step you were previously on. Before, you would not be taken into the step but now you should.
5. Make sure the step drop down and previous step and next step buttons still work properly

Closes #1268

### DIFF
--- a/src/assets/wise5/authoringTool/node/node-authoring/node-authoring.component.ts
+++ b/src/assets/wise5/authoringTool/node/node-authoring/node-authoring.component.ts
@@ -123,7 +123,7 @@ export class NodeAuthoringComponent implements OnInit {
   }
 
   ngOnDestroy(): void {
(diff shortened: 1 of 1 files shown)


hirokiterashima (approved):
LGTM

PR #1272, opened 2023-05-19
refactor(NodeAuthoring): Extract ComponentAuthoringComponent 

### LINKED ISSUE #1271: refactor(NodeAuthoring): Extract ComponentAuthoringComponent
Create a component for dynamically showing the right component authoring view. Instead of using Switch based on component.type in node-authoring.component.html, do a lookup in this new class and use Angular's createComponent() to dynamically pick the right [OR|MC|HTML...]AuthoringComponent to use.

hirokiterashima:
## Changes
- Rename existing ComponentAuthoring class to AbstractComponentAuthoring
- Create a component called ComponentAuthoringComponent to dynamically choose and render the specific component authoring view. 
   - Instead of using Switch based on component.type in node-authoring.component.html, lookup the component in ComponentAuthoringComponent and use Angular's createComponent() to dynamically create the right [OR|MC|HTML...]AuthoringComponent to show.
- Pass in componentContent from NodeAuthoringComponent to ComponentAuthoringComponent, and into individual [OR|MC|HTML...]AuthoringComponents. Remove the need to get the componentContent again in child components
- Clean up surrounding code
   - Lower-case variable names
    
## Test
- In the AT, the correct component authoring views load and work as before for all components.

Closes #1271


geoffreykwan (approved):
Looks good.

PR #1273, opened 2023-05-19
fix(Branching): Fix step numbering when a branch spans multiple groups

### LINKED ISSUE #1262: bug(Branching): If a branch spans multiple groups, the branch path step numbering is incorrect
1. Author a unit with two lessons
2. Create a branch where the branch point step is in the first lesson and the paths and merge point step is in the second lesson
3. The step numbering for the branch path steps will not be correct

geoffreykwan:
## Changes

Correctly number branch steps when a branch spans multiple groups.

## Test

- Create a branch where the branch point is in one group and the merge point is in another group
- Make sure the step numbers are correct
- Make sure step numbering is correct for projects that do not have a branch

Closes #1262

### DIFF
--- a/src/app/services/projectService.spec.ts
+++ b/src/app/services/projectService.spec.ts
@@ -441,6 +441,12 @@ function parseProject(): void {
 }
 
 function calculateNodeNumbers() {
+  calculateNodeNumbersWhenNoBranches();
+  calculateNodeNumbersWhenBranchInOneActivity();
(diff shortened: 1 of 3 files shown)


hirokiterashima (approved):
LGTM.

ProjectService.calculateNodeNumbersHelper() is complex and hard to understand. Let's try to refactor this in a separate task. I'll create an issue.

PR #1276, opened 2023-05-22
refactor(PreviewComponent): use createComponent() instead of ngSwitch

### LINKED ISSUE #1275: refactor(PreviewComponent): Use createComponent instead of ngSwitch
Create a component for dynamically showing the right component preview view. Instead of using Switch based on component.type in preview-component.component.html, do a lookup in the class and use Angular's createComponent() to dynamically pick the right [OR|MC|HTML...]Component to use.

hirokiterashima:
## Changes
- Change PreviewComponent to dynamically show the right component preview view. Instead of using Switch based on component.type in preview-component.component.html, do a lookup in the class and use Angular's createComponent() to dynamically pick the right [OR|MC|HTML...]Component to use.
- Code clean up
   - Pass in WISEComponent to SaveStarterStateComponent and PreviewComponentDialogComponent instead of nodeId/componentId
   - Add ```protected``` and ```private``` and function return types where applicable

## Test
- AT node editing: preview component works as before
   - save/delete starter states for ConceptMap, Draw, Label
- CM grade by step: Step Info dialog correctly shows components
   - Summary shows summary for current period 

Closes #1275


geoffreykwan (changes requested):
Saving starter state doesn't work sometimes. Deleting starter state doesn't work sometimes either.

hirokiterashima:
I moved component mappings to separate file. This avoids the code duplication issue. Can you test out component authoring and preview again, and make sure they all work?

The saving starter state issue exists in the develop branch. It happens when you go straight to component preview (before expanding the component). The component authoring's ngOnInit() hasn't been called yet, which listens to the starter state changes. The workaround is to expand the component first, then click on the preview. 

I'll write this up and we can address it in a separate issue? We should discuss how we want to handle this- don't show the preview button until it's been expanded, like the advanced? Or move the listener outside of the component authoring?

geoffreykwan (approved):
Looks good.

In Components.ts we might want to order the authoring and student fields alphabetically in the objects.

We can fix the save/delete starter state bug in a different issue.

PR #1277, opened 2023-05-22
fix(Constraint): Constraint in branch prevents visiting any step

### LINKED ISSUE #1249: bug(Branch): MakeAllNodesAfterThisNotVisitable constraint
### Description
When a unit has a branch and the branch has a step with "Make all nodes after this not visitable" constraint, you cannot go to any step in the unit even when you are on a node that is before the branch.

### Steps to reproduce
1. Download and import this unit: [530.zip](https://github.com/WISE-Community/WISE-Client/files/11446419/530.zip)
2. Preview the unit. From step 1.1, try going to step 1.2. You will see the "Item Locked" dialog with the message "To visit 1.2 you need to complete 1.3A: a"

### Notes
If the constraint is on a non-branch step, it doesn't have this problem.

geoffreykwan:
## Changes

Check if a node is in a path before checking if it comes before another node.

## Test

1. Download and import this unit: [530.zip](https://github.com/WISE-Community/WISE-Client/files/11446419/530.zip)
2. Preview the unit. From step 1.1, try going to step 1.2. You used to see the "Item Locked" dialog with the message "To visit 1.2 you need to complete 1.3A: a" but now you should be able to go to the step.

Closes #1249

### DIFF
--- a/src/assets/wise5/services/projectService.ts
+++ b/src/assets/wise5/services/projectService.ts
@@ -20,7 +20,7 @@ import { ReferenceComponent } from '../../../app/domain/referenceComponent';
 export class ProjectService {
   achievements: any = [];
   additionalProcessingFunctionsMap: any = {};
-  allPaths: string[] = [];
+  allPaths: string[][] = [];
   applicationNodes: any = [];
   flattenedProjectAsNodeIds: any = null;
   groupNodes: any[] = [];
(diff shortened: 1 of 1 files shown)


hirokiterashima (approved):
LGTM. change ```pathIncludesNodesAndOneComesBeforeTwo()``` to private function before merge?

PR #1281, opened 2023-05-23
refactor(ComponentComponent): use createComponent() instead of ngSwitch

### LINKED ISSUE #1280: refactor(ComponentComponent): Use createComponent instead of ngSwitch
Instead of using Switch based on component.type in component.component.html, do a lookup in the class and use Angular's createComponent() to dynamically pick the right [OR|MC|HTML...]Component to use.

hirokiterashima:
## Note
- This PR introduces a new code duplication between ComponentComponent and PreviewComponent. Addressing it will require a bit of changes, so I think we should handle this in a separate task to limit the scope of this PR.

## Changes
- Instead of using ngSwitch based on component.type in component.component.html, do a lookup in the class and use Angular's createComponent() to dynamically pick the right [OR|MC|HTML...]Component to use. This affects the components that are used in the Student VLE and Unit Preview.

## Test
- Student VLE
   - Components are displaying correctly
   - Components save and load work correctly
      - via save, submit, and auto-save  
- Preview unit
   - Node and component rubric appears (if authored) 
- Teacher CM
   - Check export to ensure that student data (component data, latest/all work) is saved correctly, including revisions (save, submit, auto-save). Ensure that only the data that you expect is saved, and no redundant saving is done as a result of the change

Closes #1280


geoffreykwan (approved):
Looks good.

PR #1290, opened 2023-05-26
refactor(Graph): Move table data point and multiple y axes code to util

### LINKED ISSUE #1289: refactor(Graph): Move table data point and multiple y axes code to util

geoffreykwan:
## Changes

Move table data point to graph functions and multiple y axes check to util.

## Test

- Make sure importing table data to a graph works
- Make sure multiple Y axes works

Closes #1289


hirokiterashima (approved):
LGTM. It'd be good to start defining some of these data classes/interfaces like Cell and Point instead of using ```any```. I think this will make the code easier to read. Something to keep in mind for future refactoring tasks.

PR #1291, opened 2023-05-26
refactor(NodeService): Move MoveComponent() to Node

### LINKED ISSUE #1226: refactor(NodeService): Move MoveComponent() to Node
NodeService.moveComponent(), extractComponents(), and insertComponentsAfter() should not be a function in the NodeService. It should instead be a function in the Node class.

hirokiterashima:
## Changes
- Move NodeService.moveComponent() to Node.moveComponents() and cleaned up the code a bit.
- Remove parameter from call to this.highlightNewComponentsAndThenShowComponentAuthoring(). It wasn't being called with a parameter in the previous version either.
- Add tests for Node.moveComponents().

## Test
In the AT node authoring view, test the following
- Move component(s) to the top
- Move component(s) after an existing component

Closes #1226

### DIFF
--- a/src/assets/wise5/authoringTool/node/node-authoring/node-authoring.component.ts
+++ b/src/assets/wise5/authoringTool/node/node-authoring/node-authoring.component.ts
@@ -375,18 +375,14 @@ export class NodeAuthoringComponent implements OnInit {
         alert($localize`You are not allowed to insert the selected items after itself.`);
       }
     } else {
-      const newComponents = this.nodeService.moveComponent(
-        this.nodeId,
-        selectedComponentIds,
-        componentId
-      );
+      this.projectService.getNode(this.nodeId).moveComponents(selectedComponentIds, componentId);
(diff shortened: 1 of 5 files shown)


geoffreykwan (changes requested):
Moving components works well.

After the components are moved, they should be highlighted. It looks like it's a bug on PROD where it does not highlight the components after you move them. It should be pretty easy to pass in the components to highlightNewComponentsAndThenShowComponentAuthoring() so that they're highlighted after being moved.

hirokiterashima:
Calling highlightNewComponentsAndThenShowComponentAuthoring() also expands the components' authoring views. Is this the expected behavior? Do authors plan to edit the component(s) immediately after moving them? Alternatively, we can only do the highlighting?

geoffreykwan:
Just highlighting without expanding should be fine.

geoffreykwan (approved):
Looks good.

PR #1299, opened 2023-06-02
refactor(StudentDataService): Move NodeClickLock to StudentNodeService

### LINKED ISSUE #1298: refactor(StudentDataService): Move NodeClickLock to StudentNodeService
Handle NodeClickLocked event in StudentNodeService. This is the only place where this event occurs. Since it's an event related to node, it makes more sense to handle it in NodeService instead of DataService.

hirokiterashima:
## Changes
- Handle NodeClickLocked event in StudentNodeService. Instead of firing an event, handle the dialog launch directly in StudentNodeService. This is the only place where this event occurs. Since it's an event that is related to Node, it makes more sense to handle it in StudentNodeService instead of StudentNodeService
- Remove NodeClickLockService
- Clean up getConstraintMessage() by removing unnecessary null checks

## Test
- In Student VLE and Preview, make sure that trying to visit a node that is not visitable (due to constraints) shows the "Item Locked" dialog with the message about how to remove the restriction.
   - If there are multiple constraints, they should all appear in bullet points
- Unit without constraints load and work as before

Closes #1298


geoffreykwan (approved):
Looks good.

PR #1301, opened 2023-06-05
refactor(ConfigService): Move and rename sortClassmateUserInfosAlphabeticallyByNameHelper()

### LINKED ISSUE #1250: refactor(ConfigService): Move sortClassmateUserInfosAlphabeticallyByNameHelper()
This function is not using any data/functions in ConfigService and can be moved into its own file (user.ts?). Come up with a shorter name?

breity:
## Changes
Moved function to `user.ts` and renamed to shorten and better identify as a comparator function.

## Test
- Make sure that apps (VLE, Teacher Tools, Authoring) still work as before and that `ConfigService.classmateUserInfos` is sorted by username once config is retrieved.
- Test in a run with a student that is not in a workgroup.

Closes #1250.

### DIFF
--- a/src/assets/wise5/common/user/user.ts
+++ b/src/assets/wise5/common/user/user.ts
@@ -1,8 +1,7 @@
 export function usernameComparator(object1: { username: string }, object2: { username: string }) {
-  if (object1.username == null || object2.username == null) {
-    return 0;
-  }
-  const username1 = object1.username.toLowerCase();
-  const username2 = object2.username.toLowerCase();
-  return username1.localeCompare(username2);
+  const username1 = object1.username;
+  const username2 = object2.username;
+  return object1.username == null || object2.username == null
+    ? 0
+    : username1.toLowerCase().localeCompare(username2.toLowerCase());
 }
--- a/src/assets/wise5/common/user/user.ts
+++ b/src/assets/wise5/common/user/user.ts
@@ -1,0 +1,10 @@
+export function usernameComparator(object1: { username: string }, object2: { username: string }) {
+  const username1 = object1.username.toLowerCase();
+  const username2 = object2.username.toLowerCase();
+  if (username1 < username2) {
+    return -1;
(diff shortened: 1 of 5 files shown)


src/assets/wise5/common/user/user.ts
```diff
@@ -0,0 +1,10 @@
+export function usernameComparator(object1: { username: string }, object2: { username: string }) {
+  const username1 = object1.username.toLowerCase();
+  const username2 = object2.username.toLowerCase();
+  if (username1 < username2) {
+    return -1;
+  } else if (username1 > username2) {
+    return 1;
+  }
+  return 0;
```
hirokiterashima: Can you use localeCompare() here?
```suggestion
  return username1.localeCompare(username2);
```

hirokiterashima (commented):
Can we use [String.localeCompare()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) in the string comparison?

Also, can you look into the failing unit tests?

src/assets/wise5/common/user/user.ts
```diff
@@ -0,0 +1,8 @@
+export function usernameComparator(object1: { username: string }, object2: { username: string }) {
+  if (object1.username == null || object2.username == null) {
+    return 0;
+  }
+  const username1 = object1.username.toLowerCase();
+  const username2 = object2.username.toLowerCase();
+  return username1.localeCompare(username2);
```
hirokiterashima: You can use a ternary operator here to remove a return and reduce complexity by 1.
```suggestion
  const username1 = object1.username;
  const username2 = object2.username;
  return username1 == null || username2 == null
    ? 0
    : username1.toLowerCase().localeCompare(username2.toLowerCase());
```

hirokiterashima (approved):
LGTM. I added a code comment inline. Can you consider it before merging?

PR #1311, opened 2023-06-08
refactor(TeacherProjectService): Move deleteComponent() to Node

### LINKED ISSUE #1305: refactor(TeacherProjectService): Move deleteComponent() to Node
This function should be a member function in the Node class instead.

hirokiterashima:
## Changes
- Move TeacherProjectService.deleteComponent() to Node.deleteComponent()

## Test
- In node authoring view, deleting single component and multiple components work as before

Closes #1305

### DIFF
--- a/src/assets/wise5/common/Node.spec.ts
+++ b/src/assets/wise5/common/Node.spec.ts
@@ -8,6 +8,7 @@ let node = new Node();
 
 describe('Node', () => {
   copyComponents();
+  deleteComponent();
   getNodeIcon();
   getNodeIdComponentIds();
   insertComponents();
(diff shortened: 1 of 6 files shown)


src/assets/wise5/common/Node.spec.ts
```diff
@@ -28,6 +27,24 @@ function copyComponents() {
   });
 }
 
+function resetComponents() {
+  node = Object.assign(new Node(), {
+    components: [{ id: componentId1 }, { id: componentId2 }]
+  });
```
geoffreykwan: This code is duplicated in the insertComponents() function in the same file. There are also other places in the file where the components are initialized so the resetComponents() function name seems a little inaccurate. Maybe the top describe function should have a beforeEach() and inside of it the node should be initialized to new Node(). Then in the test functions they can set node.components to whatever they need.

geoffreykwan (changes requested):

src/assets/wise5/common/Node.spec.ts
hirokiterashima: Good point. I made some changes. PTAL

geoffreykwan (approved):
Looks good.

PR #1316, opened 2023-06-16
fix(Login): Users in China can't sign in

### LINKED ISSUE #1315: bug(Login): Users in China can't sign in
We recently added Recaptcha verification for sign in but this doesn't work for users in China because the Google domain is blocked.

geoffreykwan:
## Changes

When making the Recaptcha verification request, use https://recaptcha.net/recaptcha/api.js instead of https://www.google.com/recaptcha/api.js.

## Test

Make sure you can still sign in when Recaptcha is enabled.

Closes #1315


hirokiterashima (approved):
LGTM

Issue #1320, opened 2023-06-28
fix(Authoring): Outside Resource model images extend beyond card


geoffreykwan:

geoffreykwan:
Fixed in #1321

PR #1321, opened 2023-06-28
fix(Authoring): Outside Resource model images extend beyond card

geoffreykwan:
## Changes

Changed margin styling for model card title, image, and actions.

## Test

Open the authoring view for an Outside Resource component and make sure the model images do not extend beyond the card. Also make sure the card title and actions have padding so that they are not right up against the edge of the card.

#1320


hirokiterashima (approved):
LGTM

PR #1325, opened 2023-06-28
chore(Unit Authoring): UI Improvements

### LINKED ISSUE #1324: chore(Unit Authoring): UI Improvements
### Description
- Remove Preview without constraints button
- Move rubric button to advanced

hirokiterashima:
## Changes
- Remove "Preview without constraints" button in unit authoring view
- Move rubric button to advanced unit authoring view
- Tooltip texts on buttons now appear in advanced unit editing view
- Clean up surrounding code

## Test
- Clicking on review unit button launches preview with constraints in a new tab
- Unit rubric editing 
   - is in unit advanced view and can be toggled via button
   - saves and loads rubric changes

Closes #1324

### DIFF
--- a/src/assets/wise5/authoringTool/advanced/advanced-project-authoring.component.html
+++ b/src/assets/wise5/authoringTool/advanced/advanced-project-authoring.component.html
@@ -4,42 +4,52 @@
     color="primary"
     id="backToProjectButton"
     class="topButton"
+    matTooltip="Back to Unit Plan"
+    matTooltipPosition="above"
+    i18n-matTooltip
     (click)="goBack()"
   >
-    <mat-icon
-      class="projectButtonTooltip"
-      matTooltip="Back to Unit Plan"
-      matTooltipPosition="above"
-      i18n-matTooltip
-      >arrow_back</mat-icon
-    >
+    <mat-icon>arrow_back</mat-icon>
+  </button>
+  <button
+    mat-raised-button
+    color="primary"
+    class="topButton"
+    matTooltip="Show Rubric"
+    matTooltipPosition="above"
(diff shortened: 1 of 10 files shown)


geoffreykwan (changes requested):
- Go to the unit advanced view
- Click the "Show Rubric" button
- Click the "Show JSON" button

The JSON textarea will show up below the rubric textarea. When one of the buttons is clicked, it should hide the other textarea.

Also, the script file name and unit url fields are always visible even when you show rubric or show JSON which is a little weird. It seems to have always been like this though.

hirokiterashima:
The existing "Show JSON" button is also toggle. If you click on it, you can still see the script name and unit URL elements below it. 

I think we have two options:
1. change the button tooltip text to "Show/hide JSON" and "Show/hide Rubric", to better reflect that they are toggle buttons.
2. put those elements be in its own button/tab. What button icon should we use in this case?

Option 1 is easier. Option 2 will take a bit more work. I'm not sure it's worth the work at this time, since the unit advanced view is rarely used, and we might redesign it to something better in the future anyway. What do you think?

geoffreykwan:
I don't think it should be a toggle anymore because there are now more views (rubric and JSON). A simple solution could be

- When the rubric button is clicked, the JSON view becomes hidden
- When the JSON button is clicked, the rubric view becomes hidden

The better solution would be your option 2, to put the script filename and unit url elements in its own view with its own button.

hirokiterashima:
I like your suggestion and changed it so Rubric/JSON is shown, or neither is shown, but not both. Script name/URL is always shown, as before. PTAL.

geoffreykwan (approved):
Looks good.

PR #1339, opened 2023-07-12
feat(Node Authoring): Add Component Dialog

### LINKED ISSUE #1338: feat(Node Authoring): Add Component Dialog
Change the way authors add components by asking them to select where to add the component to before asking for the component to add.

Add “add” button to each component section, should add it below the component. Clicking on the button should pop up component chooser

hirokiterashima:
## Changes
- Change the way authors add components by asking them to select where to add the component to before asking for the component to add.
- Add “add” button to each component section, should add it below the component. Clicking on the button pops up component chooser, where author can choose a component type or import component(s) from another unit

## Test
- Add component from existing component types
- Import component(s)
- Add to top, middle, and bottom of the step
- Cancel out of flows takes you back to node authoring view

Closes #1338


geoffreykwan (changes requested):
1. The div that contains the add component button covers the whole width of the component and prevents clicking the bottom half of the component bar. This means you can't expand or click the check box for a component if you click anywhere on the bottom half of the component bar. The preview, copy, and delete component buttons also don't show up when you mouseover the bottom half of the component bar.

Screen Shot 2023-07-12 at 3 29 54 PM

2. When you import a component, it always adds it to the top of the step even if you clicked the add button below any of the components.

3. Should the add component popup have the Cancel button on the left side instead of the right? In other places, if Cancel is the only button, we show it on the left.

hirokiterashima:
Good points. I moved cancel buttons to left for both selecting a component and import component page, for consistency. It doesn't quite look ok though, feels like "back" belongs on the left, but cancel on the right?...let's discuss this in the future and come up with a standard. PTAL.

geoffreykwan (approved):
Looks good.

PR #1341, opened 2023-07-13
feat(Node Authoring): Dynamic Delete/Copy/Move buttons

### LINKED ISSUE #1340: feat(Node Authoring): Dynamic Delete/Copy/Move buttons
Hide global delete/copy/move buttons. Selecting component(s) with checkbox enable contextual menu above the list of components that allows authors to batch delete, copy, move.

hirokiterashima:
## Changes
- In node authoring, hide global move/copy/delete buttons. Selecting component(s) with checkbox will display the contextual menu above the list of components that allows authors to batch move/copy/delete.

## Test (Node Authoring)
- Buttons become visible when one or more components are selected, and become hidden when none are selected.
- Buttons work as before

Closes #1340


geoffreykwan (changes requested):
I noticed two usability inconveniences.

1. If you click a checkbox for a component and you aren't scrolled to the top of the screen, you can't see the move/copy/delete buttons. Previously this wasn't a problem because the buttons always hovered on the top bar.

2. If you click a checkbox for a component but forget to uncheck it, you can't add a component. In Google drive you can select files and still add a new file. Maybe we should only hide the + add component buttons only after the user has clicked the "Move Components" or "Copy Components" buttons.

hirokiterashima:
Good suggestions. I made some changes. PTAL.

geoffreykwan (approved):
Looks good.

The top hover area is now pretty tall and takes up quite a bit of vertical space when you scroll down. We might want to think about consolidating all the UI elements up there into just one or two rows in another issue. For example the top row could contain the back button, advanced button, undo button, icon button, step title input, expand all button, collapse all button, preview button. The second row could contain the Components label and add component button.

hirokiterashima:
Sounds good. Let's discuss ways to improve the UI.

PR #1345, opened 2023-07-18
feat(Node Authoring): Reorder components with drag and drop

### LINKED ISSUE #1342: feat(Node Authoring): Reorder components with drag and drop
Allow reordering components with drag and drop.

hirokiterashima:
## Changes
- Implement reorder components with drag&drop. Only one component can be moved this way at this time.
- Clean up code

## Test
- Component reorder using drag&drop works as expected

Closes #1342


geoffreykwan (changes requested):
Main drag and drop functionality seems to work well. I noticed two minor issues.

1. There are now unnecessary scroll bars around the components.

Screenshot 2023-07-20 113134

2. If there is only one component, the drag handle should not be displayed and the user should not be able to drag the component.

hirokiterashima:
I removed the drag handle if there is only one component.

@breity can you please take a look at the scrolling issue that @geoffreykwan found? I wasn't sure about the best way to handle it.

geoffreykwan (approved):
Looks good. I noticed a minor styling issue.

When there is only one component, there used to be some padding to the left of the component checkbox but now there is none. The component checkbox is a little too close to the border now. I think we should add the padding back when there is only one component.

Before

Screenshot 2023-07-24 155711

Now

Screenshot 2023-07-24 155735

PR #1375, opened 2023-07-31
fix(Node Authoring): Single component is draggable

### LINKED ISSUE #1374: bug(Node Authoring): Single component is draggable
## Description
When a step only contains one component, that component is still draggable in the step authoring view. This also results in the component being dragged when a user attempts to highlight text in an input field in the component content.

### Expected Behavior
Components should only be draggable for reordering when the step contains more than one component.

breity:
## Changes
Disables dragging on components in node authoring view when there is only one component in the step.

Resolves #1374.

## Test
- Make sure dragging components is disabled when there is only one component in the step.
- Make sure dragging to reorder components still works when there is more than one component.

### DIFF
--- a/src/assets/wise5/authoringTool/node/node-authoring/node-authoring.component.html
+++ b/src/assets/wise5/authoringTool/node/node-authoring/node-authoring.component.html
@@ -158,6 +158,7 @@
         [id]="component.id"
         class="component"
         cdkDrag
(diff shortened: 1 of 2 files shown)


hirokiterashima (approved):
LGTM

hirokiterashima:
@breity looks like the unit tests started failing due to changes in this PR. Can you take a look?

PR #1389, opened 2023-08-08
fix(Classroom Monitor): Update workgroup name on previous/next click

### LINKED ISSUE #1388: bug(Classroom Monitor): Workgroup previous and next buttons do not update displayed workgroup name
1. Load a run in the Classroom Monitor that has multiple workgroups
2. Go to the "Grade By Team" view
3. Choose a workgroup
4. Click the "Next Team" button, the student data will change but the name displayed in the drop down will remain the same
5. Click the "Previous Team" button, the student data will change but the name displayed in the drop down will remain the same

geoffreykwan:
## Changes

Update workgroup name when the workgroup changes. It used to always display the same workgroup name even when the workgroup changed.

## Test

1. Load a run in the Classroom Monitor that has multiple workgroups
2. Go to the "Grade By Team" view
3. Choose a workgroup
4. Click the "Next Team" button, the student data will change but the name displayed in the drop down will remain the same
5. Click the "Previous Team" button, the student data will change but the name displayed in the drop down will remain the same

Closes #1388

### DIFF
--- a/src/app/classroom-monitor/workgroup-select/workgroup-select-autocomplete/workgroup-select-autocomplete.component.ts
+++ b/src/app/classroom-monitor/workgroup-select/workgroup-select-autocomplete/workgroup-select-autocomplete.component.ts
@@ -16,20 +16,20 @@ import { copy } from '../../../../assets/wise5/common/object/object';
   encapsulation: ViewEncapsulation.None
 })
 export class WorkgroupSelectAutocompleteComponent extends WorkgroupSelectComponent {
-  myControl = new FormControl();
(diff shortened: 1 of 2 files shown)


hirokiterashima (approved):
LGTM.

On a side note, feel free to clean up surrounding code if you think it won't increase the review scope and won't get in the way of the review. So in this case, you can lower-case variable names (ConfigService, TeacherDataService), put ```@Input``` and customClass on the same line, add private/protected modifiers, etc. We should try to make the code cleaner than we found it if possible.

PR #1390, opened 2023-08-15
build(AT, CM, VLE): Remove UpgradeModule

### LINKED ISSUE #1391: Remove UpgradeModule
Remove UpgradeModule dependency from AT and CM, and from anywhere else it's used

hirokiterashima:
## Changes
- Replace AngularJS.UIRouter with Angular.Router in AT and CM
- Remove AngularJS NPM dependencies

## Test
- Restart docker-client-dev which runs npm install to remove old AngularJS dependencies
- Test AT, CM, Student VLE and Preview and make sure that everything works as before

Closes #1391


geoffreykwan (changes requested):
Here are some problems I found.

### Classroom Monitor
- Grade By Step shows the step select when on an activity and the previous and next buttons are disabled. On production the step select does not show up in this scenario.
- In Grade By Step when you click lock activity, you will be sent to the home page
- Classroom Monitor left side bar has fixed height but it should expand the whole height of the page
- See student revisions throws error and does not load
- Teacher report is not in correct location
- Classroom Monitor Milestone view does not load
- Some exports do not work
  - One workgroup per row
  - Latest work
  - All work
  - Component data
  - Raw data

### Authoring Tool
- Authoring Tool has fixed width and height and shows scrollbars
- Error is thrown when you go to step advanced constraint authoring view and you can't add a constraint to the step

hirokiterashima:
@geoffreykwan thanks for catching the issues. I've addressed all the non-UI related ones. Can you take another look?

@breity I've updated the "Unresolved Issues" section in the main description with UI-related issues. Can you take a stab at fixing them?

breity (approved):

geoffreykwan (approved):
Looks good.

PR #1394, opened 2023-08-24
refactor(Project Authoring): Extract move node location selection feature to component

### LINKED ISSUE #1393: refactor(Project Authoring): Extract move node location selection to new component
In the project authoring view, asking the author to select a new location for a moved node(s) is done in the same component (ProjectAuthoringComponent). This should be done in a separate component to reduce responsibility of this component.

To limit scope, handle extracting copied component in another task.

Look at ChooseComponentLocationComponent for reference.

hirokiterashima:
## Notes
- Please style and clean code/template as you see fit

## Changes
- Extract mode node location selection feature from ProjectAuthoring to ChooseMoveNodeLocationComponent
- Clean up related code

## Test
- Move node works as before:
   - one node at a time
   - multiple nodes at a time
   - steps
   - groups
   - to/from active and inactive sections 

Closes #1393

### DIFF
--- a/src/assets/wise5/authoringTool/choose-move-node-location/choose-move-node-location.component.html
+++ b/src/assets/wise5/authoringTool/choose-move-node-location/choose-move-node-location.component.html
@@ -45,19 +45,13 @@
       *ngTemplateOutlet="moveInsideButton; context: { nodeId: 'group0' }"
     ></ng-container>
   </ng-container>
-  <div
-    *ngFor="let nodeId of nodeIds"
-    [ngClass]="{
-      'node-select--group': isGroupNode(nodeId),
-      'node-select--node': !isGroupNode(nodeId)
-    }"
-  >
+  <div *ngFor="let nodeId of nodeIds">
     <div
       id="{{ nodeId }}"
       [ngClass]="{
-        groupHeader: isGroupNode(nodeId),
-        stepHeader: !isGroupNode(nodeId),
-        branchPathStepHeader: isNodeInAnyBranchPath(nodeId) && !isGroupNode(nodeId)
+        'lesson-header': isGroupNode(nodeId),
+        'step-header': !isGroupNode(nodeId),
+        'branch-path-step-header': isNodeInAnyBranchPath(nodeId) && !isGroupNode(nodeId)
       }"
       [ngStyle]="{ 'background-color': getBackgroundColor(nodeId) }"
     >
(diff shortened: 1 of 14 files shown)


src/assets/wise5/authoringTool/choose-move-node-location/choose-move-node-location.component.html
```diff
@@ -0,0 +1,153 @@
...
+            *ngTemplateOutlet="moveAfterButton; context: { nodeId: nodeId }"
+          ></ng-container>
+        </ng-container>
+      </div>
+    </div>
+  </div>
+</div>
+<div>
+  <div fxLayout="row" fxLayoutAlign="start center" fxLayoutGap="10px">
+    <h6 class="underline">Unused Lessons</h6>
```
geoffreykwan: Add i18n to the "Unused Lessons" element.

geoffreykwan (changes requested):
Functionality works well.

I'm wondering if instead of using ng-templates it would be better to create components. It would simplify the html and the components could be reused in other templates.

This looks like code duplication. It seems like Code Climate doesn't check html files?

https://github.com/WISE-Community/WISE-Client/blob/0497dd7d76ebd7ac9e2ee68391cc024e8e8d312f/src/assets/wise5/authoringTool/choose-move-node-location/choose-move-node-location.component.html#L118-L125

https://github.com/WISE-Community/WISE-Client/blob/0497dd7d76ebd7ac9e2ee68391cc024e8e8d312f/src/assets/wise5/authoringTool/choose-move-node-location/choose-move-node-location.component.html#L142-L149

src/assets/wise5/authoringTool/choose-move-node-location/choose-move-node-location.component.html
hirokiterashima: Done. We'll hold off on making components until we extract ChooseCopyNodeLocationComponent from ProjectAuthoring. And yes, it seems like CodeClimate does not check html atm. Maybe we told it to ignore html files?

breity (approved):
Looks good. I cleaned up some styles and removed unused classes.

I also updated margin styles to be friendly to RTL languages.

hirokiterashima:
Thanks @breity. Good to remove unused styles. Template files look cleaner now.

@geoffreykwan can you take a final look?

geoffreykwan (changes requested):
When you move a step or a lesson and are in the view to choose the location, the unused lessons and unused steps are not indented correctly.

hirokiterashima:
Fixed. PTAL.

geoffreykwan (approved):
Looks good.

PR #1398, opened 2023-08-30
feat(accessibility): Ensure links have sufficient contrast

### LINKED ISSUE #1392: feat(accessibility): Ensure links have sufficient contrast
## Description
Our `<a>` links [fail WCAG AA standards](https://webaim.org/resources/linkcontrastchecker/?fcolor=000000&bcolor=FFFFFF&lcolor=1C8DA8) due to insufficient contrast between link color and background. We should change link color and consider adding `text-decoration: underline` style to links to make them accessible.

breity:
## Changes
- Darkened primary and accent colors so they have sufficient contrast for readability on light backgrounds.
- Added underline to link text to distinguish links from surrounding text.
- Changed color and added keyboard accessibility to some links in the site.

A nice resource for checking contrast for accessibility: [Contrast Checker](https://webaim.org/resources/contrastchecker/).

Resolves #1392.

## Test
- Make sure regular links have underlines across apps.
- Make sure you can use tab and enter key to interact with links on Teacher class schedule and unit libraries.


hirokiterashima (changes requested):
New links and navigation with tab/enter keys look good.

It looks like the changes will introduce code duplication with the showInfo() functions. Can you try moving the showInfo() function to the parent LibraryComponent class?

Something like this?
```
(LibraryComponent)
...
  protected showInfo(event: Event): void {
    event.preventDefault();
    this.dialog.open(getDetailsComponent(), {
      panelClass: 'dialog-sm'
    });
  }

  protected abstract getDetailsComponent(): any;
```

```
(PersonalLibraryComponent, and other child components)
...
  protected getDetailsComponent(): any {
    return PersonalLibraryDetailsComponent;
  }
```

hirokiterashima (approved):
LGTM

PR #1410, opened 2023-09-08
feat(Feedback Rule): New myChoiceChosen token

### LINKED ISSUE #1409: Feedback Rule new "myChoiceChosen" token
Add support for ```myChoiceChosen("choiceId")``` token that returns true when the student chose the specified choice as their last response for the reference component. This should only work in activities that support a reference component, namely the Dynamic Prompt and Question Bank.

hirokiterashima:
## Changes
- Add support for ```myChoiceChosen("choiceId")``` token that returns true when the student chose the specified choice as their last response for the reference component.

## Test
- Test with https://github.com/WISE-Community/WISE-API/pull/235
- Add ```myChoiceChosen("choiceId")``` in the FeedbackRule expression for Dynamic Prompt, Question Bank and make sure that the rule is matched if the student chose the specified choice for the reference component. If should not match otherwise.
- Existing Peer Chat steps with Dynamic Prompt, Question Bank work as before

Closes #1409


geoffreykwan (changes requested):
1. The choseChoice rules do not work in Peer Chat Dynamic Prompt and Question Bank if the Reference Component is set to the Multiple Choice component. It only works if the Reference Component is set to an Open Response CRater component.

2. Ideally you should be able to set the Reference Component to the Multiple Choice component and then only need to specify the choice id in the expression like ```choseChoice("mdfmfybcuw")```.

3. You can't choose Multiple Choice components in the Reference Component. In the Component drop down, Multiple Choice components are greyed out and can't be selected.

hirokiterashima:
@geoffreykwan I addressed those issues. ```myChoiceChosen("choiceId")``` token now only works in QuestionBank and DynamicPrompt, and it looks at the ReferenceComponent's data to evaluate. PTAL

src/assets/wise5/components/common/cRater/CRaterResponse.ts
```diff
@@ -1,18 +1,15 @@
...
-  submitCounter: number;
 
   constructor(jsonObject: any = {}) {
-    for (const key of Object.keys(jsonObject)) {
-      if (jsonObject[key] != null) {
-        this[key] = jsonObject[key];
-      }
-    }
+    super(jsonObject);
+    this.populateFields(jsonObject);
```
geoffreykwan, line 12: Is this call to populateFields() necessary if the parent also calls it?

src/assets/wise5/components/common/feedbackRule/TermEvaluator/test-utils.ts
```diff
@@ -5,7 +5,7 @@ import { AbstractIdeaCountTermEvaluator } from './AbstractIdeaCountTermEvaluator
 const idea1 = new CRaterIdea('1', true);
 const idea2 = new CRaterIdea('2', true);
 const idea3 = new CRaterIdea('3', true);
-const response_with_idea_1 = new CRaterResponse({ ideas: [idea1] });
+const response_with_idea_1 = new CRaterResponse({ ideas: [idea1], ba: 1 });
```
geoffreykwan: Is the ba: 1 needed here? I can't tell what it's for.

src/assets/wise5/components/common/feedbackRule/FeedbackRuleExpression.ts
```diff
@@ -28,7 +28,7 @@ export class FeedbackRuleExpression {
     return this.text
       .replace(/ /g, '')
       .split(
-        /(hasKIScore\(\d\)|accumulatedIdeaCountEquals\(\d\)|accumulatedIdeaCountLessThan\(\d\)|accumulatedIdeaCountMoreThan\(\d\)|ideaCountEquals\(\S+\)|ideaCountLessThan\(\S+\)|ideaCountMoreThan\(\S+\)|isSubmitNumber\(\d+\)|&&|\|\||!|\(|\))/g
+        /(hasKIScore\(\d\)|accumulatedIdeaCountEquals\(\d\)|accumulatedIdeaCountLessThan\(\d\)|accumulatedIdeaCountMoreThan\(\d\)|ideaCountEquals\(\S+\)|ideaCountLessThan\(\S+\)|ideaCountMoreThan\(\S+\)|myChoiceChosen\(\S+\)|isSubmitNumber\(\d+\)|&&|\|\||!|\(|\))/g
```
geoffreykwan, line 31: Can you split this up into multiple lines to make it more readable?

src/assets/wise5/components/peerChat/peer-chat-question-bank/peer-chat-question-bank.component.ts
```diff
@@ -70,8 +70,8 @@ export class PeerChatQuestionBankComponent implements OnInit {
   ): QuestionBankRule[] {
     const cRaterResponses = peerGroupStudentData.map((peerMemberData: PeerGroupStudentData) => {
```
geoffreykwan: Should this variable be renamed to `responses` since it can now be Multiple Choice responses too?

src/assets/wise5/components/peerChat/peer-chat-question-bank/peer-chat-question-bank.component.ts
```diff
@@ -26,14 +26,20 @@ export class PeerChatQuestionBankComponent implements OnInit {
...
   ngOnInit(): void {
     if (this.displayedQuestionBankRules == null) {
-      const referenceComponent = this.getReferenceComponent(this.content.questionBank);
+      const referenceComponent = this.projectService.getReferenceComponent(
+        this.content.questionBank
+      );
       if (
         this.content.questionBank.isPeerGroupingTagSpecified() &&
-        referenceComponent.content.type === 'OpenResponse'
+        ['OpenResponse', 'MultipleChoice'].includes(referenceComponent.content.type)
```
geoffreykwan: Should the order of the component types be alphabetical like in the other locations? Ideally all these locations would just reference a single array somewhere that contains ['MultipleChoice', 'OpenResponse'].

src/assets/wise5/directives/dynamic-prompt/DynamicPromptEvaluator.ts
```diff
@@ -0,0 +1,78 @@
...
+          this.component.constraintService
+        )
+      : new FeedbackRuleEvaluator(
+          new FeedbackRuleComponent(
+            this.component.dynamicPrompt.getRules(),
+            referenceComponent.content.maxSubmitCount,
+            false
+          ),
+          this.component.constraintService
+        );
```
geoffreykwan: Is this new FeedbackRuleComponent() the same in both cases? If so, can it be done once and set into a single variable.

```
new FeedbackRuleComponent(
  this.component.dynamicPrompt.getRules(),
  referenceComponent.content.maxSubmitCount,
  false
)
```

geoffreykwan (changes requested):
When I test the Peer Chat in preview the questions in the question bank show up properly but when I test with an actual run, the questions don't show up and this error is thrown in the browser console.

```
ERROR TypeError: Cannot read properties of undefined (reading 'length')
    at PeerChatQuestionBankComponent.filterQuestions (peer-chat-question-bank.component.ts:105:38)
    at PeerChatQuestionBankComponent.chooseQuestionBankRulesToDisplay (peer-chat-question-bank.component.ts:87:17)
    at Object.next (peer-chat-question-bank.component.ts:57:38)
```

It looks like in peer-chat-question-bank.component.ts line 71 the peerGroupStudentData is an empty array even though both of my students have submitted the Multiple Choice. In preview, it retrieves the student data locally but in a real run it retrieves the work from the API and it looks like the API is returning an empty array.

src/assets/wise5/components/common/feedbackRule/TermEvaluator/test-utils.ts
hirokiterashima: I was using this for testing something. Removed.

src/assets/wise5/components/peerChat/peer-chat-question-bank/peer-chat-question-bank.component.ts
hirokiterashima: Hmm. Maybe in a ReferenceableComponents class or isReferenceableComponent() function?

src/assets/wise5/components/peerChat/peer-chat-question-bank/peer-chat-question-bank.component.ts
hirokiterashima: Changed.

src/assets/wise5/directives/dynamic-prompt/DynamicPromptEvaluator.ts
hirokiterashima: Good catch. Changed.

src/assets/wise5/components/common/feedbackRule/FeedbackRuleExpression.ts
hirokiterashima: I spent 30 minutes on this but couldn't figure it out. Can you? Or, in the interest of moving forward, let's save this for another time?

src/assets/wise5/components/common/cRater/CRaterResponse.ts
hirokiterashima: When we run the unit tests, because we initialize the ideas variable to an empty array ```ideas: CRaterIdea[] = [];``` in the class declaration, the webpack compiler (I think) adds this line below the line super(jsonObject) ```this.ideas = [];```. If we don't call ```this.populateFields(jsonObject)``` again, the ideas fields is always an empty array, and causes the tests to fail. So in effect, this line is there just to make the tests pass, which is not ideal.

This problem goes away if we don't initialize the ideas variable to an empty array, but this may cause side-effects with the actual behavior of the program.

hirokiterashima:
I forgot to note that you need to test this PR with https://github.com/WISE-Community/WISE-API/pull/235. My bad.

breity (approved):
Functionality is working well. 👍 

I did find one minor grading issue, but it exists on production and applies to all Peer Chat Question Banks. So we should make a separate issue. This is the problem:
The grading view can show different Question Bank questions than what the students see. This happens when a workgroup adds something to the chat and then goes back to the reference component and changes their answer. When the workgroup revisits the Peer Chat component, the Question Bank will update based on the students' latest response in the reference component. But the grading view will show the old Question Bank questions because the student hasn't submitted any more work to the Peer Chat component.

src/assets/wise5/components/common/cRater/CRaterResponse.ts
geoffreykwan: Seems like somewhat unexpected behavior but is good to know for the future.

src/assets/wise5/components/common/feedbackRule/FeedbackRuleExpression.ts
geoffreykwan: You could try something like this but we can do it later

https://stackoverflow.com/questions/12317049/how-to-split-a-long-regular-expression-into-multiple-lines-in-javascript

src/assets/wise5/components/peerChat/peer-chat-question-bank/peer-chat-question-bank.component.ts
geoffreykwan: Yea, somewhere to put the single source of truth to avoid duplication.

geoffreykwan (approved):
Looks good.

hirokiterashima:
Thanks @breity I wrote up the issue here: https://github.com/WISE-Community/WISE-Client/issues/1419

PR #1411, opened 2023-09-10
refactor(ProjectAuthoringComponent): deleteNodes() and selectNode()

hirokiterashima:
## Changes
- Call DeleteNodeService functions instead of handling it in ProjectAuthoringComponent
- Rename projectItemClicked() to selectNode() and clean up code using HOF
- Rename activityNodeSelected to groupNodeSelected

## Test
- Delete nodes work as before
   - Delete step(s)
   - Delete group(s)
   - Delete first step in the unit should update start node
- Selecting node(s) work as before
   - Shouldn't be able to select mix of groups and steps


geoffreykwan (approved):
Looks good.

PR #1424, opened 2023-09-20
refactor(ProjectAuthoringComponent): Move scrollToBottom() to dom.ts

hirokiterashima:
## Changes
- Move scrollToBottom request handling to scrollToElement(elementId) function in dom.ts.
- Remove need to emit request via TeacherProjectService and instead call scrollToElement() directly.

## Test
- In node constraint authoring, adding a new constraint
   - highlights the new constraint
   - scrolls to the new constraint 

(Note: neither of these are working on the develop branch atm, so we must have broken it at some point)


geoffreykwan (approved):
Looks good.

PR #1425, opened 2023-09-20
refactor(LibraryComponent): Clean up isSearchMatch()

hirokiterashima:
## Changes
- Use filter to filter metadata fields we're interested in
- Use includes() instead of multiple &&'s
- Combine return statements
- Change function to private

## Test
- Searching for units (on home page and in your teacher's Unit Library) works as before


breity (approved):
Looks good.

PR #1430, opened 2023-09-22
feat(Import Step): Add dedicated choose unit page before choosing step

### LINKED ISSUE #1429: Import Step: Add dedicated choose unit page before choosing step
Right now, when importing a step in the AT, the author is asked to choose a unit and a step in that unit, both in the same page. This clutters the interface and makes the code more complex. Instead, this should be broken up into two phases: 1. Choose unit to import from, and 2. Choose step(s) to import.

In step 1, change the UI from a drop-down to a tabbed list, so you can see more units at once. During our user studies, one researcher noted how difficult it was for them to find the unit to import.

hirokiterashima:
## Notes
- @breity Please style the pages as you see fit

## Changes
- In AT > Import Step feature, break up the choose unit&step page into two pages: 
   - Choose unit to import from, and 
   - Choose step(s) to import.
- Change the Choose unit UI from a drop-down to a tabbed list (My Units and Library Units), so you can see more units at once.

## Test
- Import steps feature 
   - lets you to choose a unit and steps in different pages
   - works as before 

Closes #1429

### DIFF
--- a/src/app/authoring-tool/import-step/choose-import-step-location/choose-import-step-location.component.html
+++ b/src/app/authoring-tool/import-step/choose-import-step-location/choose-import-step-location.component.html
@@ -1,46 +1,45 @@
-<h5 i18n>Choose a location for the imported steps</h5>
-<div style="margin-top: 20px; margin-left: 20px">
-  <div *ngFor="let nodeId of nodeIds" class="project-item">
-    <div
-      id="{{ nodeId }}"
-      [ngClass]="{
-        groupHeader: isGroupNode(nodeId),
-        stepHeader: !isGroupNode(nodeId),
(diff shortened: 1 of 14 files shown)


breity (approved):
Looks good. I updated the styles and layout, making buttons for selecting units to import from, updating spacing, and making the row of action buttons at the bottom sticky when scrolling.

I found a few issues that exist on production that we should address in separate issues:
1. Branch path step numbers are not calculated and displayed properly in the step selection stage of the import process.
2. Steps in branches do not have a background color in the step selection and location selection stages of the import process.
3. It would useful to add a "Back" button to the last stage of the import process, to match the other stages. Right now there's only "Cancel", so if the user changes their mind and wants to choose different steps to import, theyneed to start the whole process over.

PR #1434, opened 2023-09-26
fix(Peer Group): Non admins can not create new peer groupings

### LINKED ISSUE #1433: bug(Peer Group): Can not create new peer grouping
1. Sign in as a teacher that does not have admin privilege
2. Open the Authoring Tool for a run
3. Create a Peer Chat step
4. For Grouping Logic click Select
5. Click "New Grouping"
6. Set the "Grouping Name" and "Logic"
7. Click Create
8. An error occurs even though you should be able to create a new peer grouping

geoffreykwan:
## Changes

Show the appropriate error message when an error is returned from the server.

## Test

- Test with and use instructions in https://github.com/WISE-Community/WISE-API/pull/237

Closes #1433


breity (approved):
Looks good. 👍 

I updated `messages.xlf` with the error message string changes.

PR #1435, opened 2023-09-26
test(Multiple Choice): Add tests to Multiple Choice authoring

geoffreykwan:
## Changes

Added tests to Multiple Choice authoring.

## Test

Only test files were changed/added so functionality should not have changed.


breity (commented):
Test are working. 👍 

A couple thoughts:
- Could maybe reorganize to better follow the Arrange-Act-Assert pattern by moving the actions out of the `it` blocks and into the `beforeEach` blocks, leaving only assertions in the `it` blocks.
- Wondering if we need all the click action function wrappers that are only used once. Could you call the click actions directly?

These are probably issues we should discuss, agree on conventions for test structures. So maybe okay to move forward here either way.

breity (approved):

PR #1436, opened 2023-09-27
test(Export): Add test for export Discussion component

geoffreykwan:
## Changes

Added test for Discussion component export.

## Test

Make sure the Discussion component export still works.


breity (approved):
Looks good.

PR #1443, opened 2023-10-03
refactor(ProjectLibraryService): getLibraryProjects() returns Observable

hirokiterashima:
## Changes
- Instead of returning a promise (which is being deprecated), return an Observable
- Make sortAndFilterUniqueProjects(), filterUniqueProjects() and sortByProjectIdDescending() private
- Lower-case ConfigService variable and change dependency order in constructor to be alphabetical
- Update tests

## Test
- Import step and import component work as before in AT

### DIFF
--- a/src/app/services/projectLibraryService.spec.ts
+++ b/src/app/services/projectLibraryService.spec.ts
@@ -22,22 +22,24 @@ describe('ProjectLibraryService', () => {
 
 function getLibraryProjects() {
   describe('getLibraryProjects()', () => {
-    it('gets the library projects, sorted and filtered', () => {
-      const getLibraryProjectsURL = '/api/project/library';
-      spyOn(configService, 'getConfigParam').and.callFake((param) => {
+    const getLibraryProjectsURL = '/api/project/library';
+    const unit1 = { id: 1, name: 'one' };
+    const unit2 = { id: 2, name: 'two' };
+    const unit3 = { id: 3, name: 'three' };
+    const libraryProjects = [
+      {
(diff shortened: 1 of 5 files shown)


src/app/services/projectLibraryService.spec.ts
```diff
@@ -33,56 +18,30 @@ describe('ProjectLibraryService', () => {
...
       ];
-      const uniqueProjects = service.filterUniqueProjects(nonUniqueProjects);
-      expect(uniqueProjects.length).toEqual(3);
-      expect(uniqueProjects[0].id).toEqual(3);
-      expect(uniqueProjects[1].id).toEqual(1);
-      expect(uniqueProjects[2].id).toEqual(2);
+      service.getLibraryProjects().subscribe((projects) => {
+        expect(projects).toEqual([unit3, unit2, unit1]);
+      });
+      http.expectOne(getLibraryProjectsURL).flush(libraryProjects);
```
breity: Could move the variable declarations and actions outside the `it` block so that only assertions are included there (AAA pattern). But we should probably discuss and agree on test conventions for the future, so okay to leave as is until then if you like.

breity (approved):
Looks good. 👍 

I made a comment for reorganizing the test structure to consider, but good to merge.

src/app/services/projectLibraryService.spec.ts
hirokiterashima: I changed it for now. Yes, let's discuss our convention going forward at some point.

PR #1445, opened 2023-10-04
feat(Import Step): Move to Create New Step Menu

### LINKED ISSUE #1444: Move Import Step option to Create New Step Menu
Like how the Import Component option is available via Create New Component menu, move the Import Step option to Create New Step menu. This will reduce the button that authors see in the project editing view by 1.

hirokiterashima:
## Notes
- @breity please feel free to style and re-word elements as you see fit
   - For example, should the "Create New Step" button be "Add New Step" in the project-authoring view? 

## Changes
- Move "Import Step" option to "Create New Step" menu (before: "Import Step" was a top-level option in the project-authoring view). This will remove the top-level button in project-authoring view by 1, and make the import step workflow similar to import component flow.

## Test
- "Import Step" option is available through "Add New Step" menu
- "Import Step" workflow works as before, including back/cancel buttons

Closes #1444

### DIFF
--- a/src/app/authoring-tool/import-step/choose-import-step-location/choose-import-step-location.component.html
+++ b/src/app/authoring-tool/import-step/choose-import-step-location/choose-import-step-location.component.html
@@ -40,8 +40,6 @@
 <div class="nav-controls">
   <mat-divider></mat-divider>
   <div fxLayout="row" fxLayoutAlign="end center">
-    <button mat-button color="primary" routerLink="../../.." aria-label="Cancel" i18n>
-      Cancel
-    </button>
+    <button mat-button color="primary" routerLink="../../.." i18n>Cancel</button>
   </div>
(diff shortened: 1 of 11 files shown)


breity (commented):
Functionality works well. 👍 

I made a couple changes:
- Removed unused `description` field from `NewNodeTemplate`.
- Removed unnecessary `aria-label` attributes from buttons that have label text.

One request: For consistency, could we add a back button to the "Choose a unit from which to import step(s)" stage in the import sequence?

I think changing "Create New Step" to "Add New Step" could make sense. If we do, we should also change "Create New Lesson" to "Add New Lesson" to be consistent.

hirokiterashima:
Good suggestions. I made the changes. PTAL.

breity (approved):
LGTM 👍

PR #1450, opened 2023-10-04
fix(Manage Students): Remove student does not update the UI

### LINKED ISSUE #1449: bug(Manage Students): Remove student does not update the UI
1. Put a student into multiple workgroups in a run. You may need to roll back your code to do this.

If you checkout these commits you should be able to place a student into multiple workgroups.
WISE-API ```3707067faa0413878dfebffeb325298d3a6fec4d```
WISE-Client ```10f3da5b4972a75bf9ecba473d7e4d09ad5db33c```

To put a student into multiple workgroups
- Open the Manage Students view for a run in two tabs
- In tab 1, move the student to a workgroup
- In tab 2, move the student to a different workgroup
- Refresh one of the tabs and the student should now be in both workgroups

2. Click "Remove student" and the student should disappear from the workgroup in the UI but nothing actually changes. On the server, the student is removed from the period and that workgroup but it is not removed from the other workgroup. If you refresh the page, the student will still show up in the other workgroup.

geoffreykwan:
## Changes

- Remove student now removes the student from the UI. The UI used to not change at all after removing a student.
- If the student is in multiple workgroups, it only removes them from the workgroup you clicked on. If we want to remove them from all workgroups, we could send a message to all workgroups. Theoretically this would not be needed if students can no longer be in multiple workgroups.

## Test

- Test with https://github.com/WISE-Community/WISE-API/pull/243
- Use testing instructions in #1449

Closes #1449


hirokiterashima (approved):
LGTM

PR #1453, opened 2023-10-04
refactor(NodeService): Clean up getNextNode() for AT/CM modes

hirokiterashima:
## Changes
- Call ```DataService.getCurrentNodeId()``` to simplify code at the beginning of the function
- Move check for mode closer to where it's actually used
- Use ternary check for application node/group node check
- Change ```let``` to ```const``` where appropriate

## Test
- Going to the next node using the next arrow in the step tools in CM, AT, Student VLE and Preview works as before.


geoffreykwan (approved):
Looks good.

PR #1458, opened 2023-10-09
refactor(NodeService): Move authoring-specific observables to TeacherNodeService

hirokiterashima:
## Changes
- Move these authoring-specific observables from NodeService to TeacherNodeService
   - componentShowSubmitButtonValueChanged$
   - deleteStarterState$
   - starterStateResponse$
- Update references

## Test
- Authoring works as before, especially:
   - saving/deleting starter state (draw, label, concept map, etc)
   - submit button automatically gets set/unset when cRater is enabled/disabled


geoffreykwan (approved):
Looks good.

PR #1459, opened 2023-10-09
refactor(NodeService): Remove scrollToComponentAndHighlight()

hirokiterashima:
## Changes
- Use existing functions in dom.ts instead of NodeService.scrollToComponentAndHighlight(). This only affects the WISE link feature when we need to scroll and highlight a specific component in the same step or in a different step.

## Test
- WISE links work as before, when link is 
   - in the same step
   - in a different step -> should go to new step
   - (regardless of same/different step): to a specific component -> should scroll to component and highlight it


geoffreykwan (approved):
Looks good.

PR #1461, opened 2023-10-10
refactor(Annotation): Create Annotation class

hirokiterashima:
## Changes
- Create Annotation class and type variables as Annotation instead of ```any``` on some (not all yet) variables where appropriate
- Remove extra wrapper { annotation: annotation } that is sent around in observables and pass in Annotation directly
- Add private modifiers and function return types
- Fix yoda expressions and remove unnecessary code

## Test
- These annotations are sent, retrieved and displayed correctly as before:
   - Teacher -> Student annotations (score, comments)
   - Student -> Teacher annotations (C-rater auto-score values in node/student grading views)
   - C-rater auto-feedback text for students


geoffreykwan (approved):
Looks good.

PR #1464, opened 2023-10-11
refactor(TeacherDataService): retrieveRunStatus() returns Observable

hirokiterashima:
## Changes
- TeacherDataService.retrieveRunStatus() now returns Observable (before it used a Promise, which is being deprecated)
- Clean up related code, add private modifier and return types
   - favor declarative code over imperative code where appropriate to simplify code
- Remove unused function retrieveLatestStudentDataByNodeIdAndComponentIdAndPeriodId()

## Test
- Load CM for any run
   - RunStatus should be retrieved from the server
   - Period dropdown should display the periods that have been set up for that run


geoffreykwan (approved):
Looks good.

PR #1466, opened 2023-10-12
feat(Export): Peer Chat component export

### LINKED ISSUE #1465: feat(Export): Peer Chat component export
Implement with these requirements
- Each row is a student response
- Each row has a Peer Group ID
- Group rows by Peer Group ID and timestamp
- If there is a Dynamic Prompt, show a Dynamic Prompt column
- If there are Question Bank Questions, show a column for each question

geoffreykwan:
## Changes

Implemented Peer Chat component export with these requirements
- Each row is a student response
- Each row has a Peer Group ID
- Group rows by Peer Group ID and timestamp
- If there is a Dynamic Prompt, show a Dynamic Prompt column
- If there are Question Bank Questions, show a column for each question

## Test

- Make sure Peer Group IDs show up
- Make sure rows are grouped by Peer Group ID and timestamp
- Test with a Dynamic Prompt and make sure the correct Dynamic Prompt shows up
- Test with a Question Bank and make sure questions show up
- Test with no Dynamic Prompt and no Question Bank
- Make sure the student response shows up

Closes #1465

### DIFF
--- a/src/assets/wise5/classroomMonitor/dataExport/data-export/data-export.component.ts
+++ b/src/assets/wise5/classroomMonitor/dataExport/data-export/data-export.component.ts
@@ -25,6 +25,7 @@ import { millisecondsToDateTime } from '../../../common/datetime/datetime';
 import { Choice } from '../../../components/match/choice';
 import { Bucket } from '../../../components/match/bucket';
 import { ActivatedRoute, Router } from '@angular/router';
+import { PeerChatComponentDataExportStrategy } from '../strategies/PeerChatComponentDataExportStrategy';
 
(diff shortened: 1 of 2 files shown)


breity (commented):
Looks good.

One minor issue I see is that the question bank questions aren't included in the export for v1 Peer Chat components. Do you see that too? (Maybe not critical, though, since there are very few v1 components in actual runs.)

I'm also wondering if the researchers would want to see the static before and after prompt content included in the export. But maybe we can add if they ask for it.

breity (approved):
LGTM. 👍

PR #1469, opened 2023-10-13
refactor(AnnotationService): getLatestScoreAnnotation()

hirokiterashima:
## Changes
- Simplify function using ```Array.filter()``` and ```Array.at()```
- Function now returns ```undefined``` instead of ```null``` if there is no latest score or auto-score
- Add private modifiers, param types and function return types
- Update references
- Add basic tests

## Test
- Teacher score and auto-score annotations appear correctly as before in 
   - Grading by step
   - Grade by team
   - Milestone report -> Student Work
   - Student VLE component scores from teacher and auto-scores from CRater

### DIFF
--- a/src/assets/wise5/services/annotationService.ts
+++ b/src/assets/wise5/services/annotationService.ts
@@ -536,15 +536,17 @@ export class AnnotationService {
     workgroupId: number,
     scoreType: 'score' | 'autoScore' | 'any' = 'any'
   ): Annotation {
-    return this.getAnnotations()
-      .filter(
-        (annotation) =>
-          annotation.nodeId == nodeId &&
-          annotation.componentId == componentId &&
-          annotation.toWorkgroupId == workgroupId &&
-          this.matchesScoreType(annotation, scoreType)
-      )
-      .at(-1);
+    return (
+      this.getAnnotations()
+        .filter(
+          (annotation) =>
+            annotation.nodeId == nodeId &&
+            annotation.componentId == componentId &&
+            annotation.toWorkgroupId == workgroupId &&
+            this.matchesScoreType(annotation, scoreType)
+        )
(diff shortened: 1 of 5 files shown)


src/assets/wise5/services/annotationService.ts
```diff
@@ -530,59 +530,31 @@ export class AnnotationService {
...
+  }
+
+  private matchesScoreType(
+    annotation: Annotation,
+    scoreType: 'score' | 'autoScore' | 'any'
+  ): boolean {
+    return (
+      ['autoScore', 'score'].includes(annotation.type) &&
+      (scoreType === 'any' || annotation.type === scoreType)
+    );
```
geoffreykwan: Should this be something like this to be more concise?

```
    return (
      (scoreType === 'any' && ['autoScore', 'score'].includes(annotation.type)) ||
      annotation.type === scoreType
    );
```

geoffreykwan (changes requested):
Works well.

I'm wondering if getLatestScoreAnnotation() should return null for consistency. There's also a slight difference in behavior between null and undefined.

Calling this
```JSON.stringify({a: null, b: undefined})```

Returns this
```'{"a":null}'```

I'm wondering if that will come into play in places like this or similar places.

https://github.com/WISE-Community/WISE-Client/blob/5191b411a485f8dd60e85933b5c3a201f2444a78/src/assets/wise5/components/embedded/embedded-show-work/embedded-show-work.component.ts#L117-L134

and

https://github.com/WISE-Community/WISE-Client/blob/5191b411a485f8dd60e85933b5c3a201f2444a78/src/assets/wise5/components/embedded/embedded-student/embedded-student.component.ts#L236-L253

geoffreykwan (approved):
Looks good.

PR #1471, opened 2023-10-16
refactor(Project Authoring): Clean up template using node-icon-and-title

hirokiterashima:
## Changes
- Where we used to show node icon and title separately:
   - Use ```<node-icon-and-title ...>``` template to reduce duplication
   - Remove surrounding div and move flex attribute(s) and pointer css class to parent element

## Test
- Project editing page
   - Shows node icon and title as before for active and inactive lessons/steps 
   - Shows pointer for mouse when hovering over node item


geoffreykwan (approved):
Looks good.

PR #1472, opened 2023-10-17
refactor(Component Grading): Introduce ComponentGradingComponent

hirokiterashima:
## Changes
- Create new ComponentGradingComponent, which dynamically creates the appropriate component grading view
   - Instead of using *ngIf to determine which component grading view to display in "grade component" and "view revision" views, use this class 

## Test
- Component grading view for all component types (you can just select a few components to sample) appears correctly as before in
   - Grade-by-step
   - Grade-by-team
   - Milestone -> Student Work tab
   - View revisions dialog


geoffreykwan (approved):
Looks good.

PR #1473, opened 2023-10-17
refactor(Project Authoring): Clean up template

hirokiterashima:
## Changes (mostly in project-authoring.component.html)
- Aria-label now says "Select step/lesson" instead of node title
- Remove the 'group0' master root node from consideration in code instead of template
- Remove unnecessary divs and styles

## Test
- Project editing page looks and works as before, specifically:
   - Shows node icon and title as before for active and inactive lessons/steps
   - Shows pointer for mouse when hovering over node item
   - Scrolls to and highlights newly added step


geoffreykwan (changes requested):
Mostly works except now you have to click on the step title text in order to go into the step. Previously you could click anywhere on the colored bar to go into the step.

hirokiterashima:
Good catch. PTAL.

geoffreykwan (changes requested):
Now when you click on the icon or the step title, it checks the checkbox. Before if you clicked on the icon or step title it would open the step.

geoffreykwan (changes requested):
Clicking now works like it used to but the steps in "Unused Lessons" and "Unused Steps" now have a newline after the checkbox.

Screenshot 2023-10-20 115353

geoffreykwan (approved):
Looks good.

PR #1475, opened 2023-10-18
feat(Authoring): Show component info button when adding a component

### LINKED ISSUE #1457: feat(Authoring): Show component info button when adding a component
When the info button is clicked a popup should display with information about the component type along with a preview.

Use this component display
- When creating a new step and choosing which components to add to it
- When adding a component to a step that already exists

geoffreykwan:
## Changes

Wherever you can add a new component, show an info button. The info button opens a dialog which contains a description of the component type as well as a preview.

## Test

Make sure the component info works when
- creating a new step
- adding a component to a step that already exists

Closes #1457


src/app/authoring-tool/add-component/choose-new-component/choose-new-component.component.html
```diff
@@ -16,18 +16,12 @@ <h5 i18n>
...
-      <mat-card-content>
-        <mat-card-title>
-          {{ componentType.name }}
-        </mat-card-title>
-      </mat-card-content>
-    </button>
+    <ng-container *ngFor="let componentType of componentTypes">
+      <component-selector
+        [componentType]="componentType.type"
+        (componentSelectedEvent)="selectComponent($event.type)"
```
hirokiterashima: We already know the component type at this point, so no need to get it from $event?
```suggestion
        (componentSelectedEvent)="selectComponent(componentType.type)"
```

src/assets/wise5/authoringTool/addNode/add-your-own-node/add-your-own-node.component.html
```diff
@@ -39,11 +39,12 @@ <h5 i18n>Select components to add to your new step (optional):</h5>
...
     <div fxFlex="100" fxFlex.gt-sm="67">
-      <div fxLayout="row wrap" fxLayoutAlign="start start">
+      <div class="component-types" fxLayout="row wrap" fxLayoutAlign="start start">
         <div class="component-type" *ngFor="let componentType of componentTypes">
-          <button mat-stroked-button (click)="addComponent(componentType)">
-            {{ componentType.name }}
-          </button>
+          <component-selector
+            [componentType]="componentType.type"
+            (componentSelectedEvent)="addComponent($event)"
```
hirokiterashima: We already know the component type at this point, so no need to send $event?
```suggestion
            (componentSelectedEvent)="addComponent(componentType)"
```

src/assets/wise5/authoringTool/components/component-selector/component-selector.component.ts
```diff
@@ -0,0 +1,43 @@
...
+import { ComponentInfoService } from '../../../services/componentInfoService';
+import { ComponentInfoDialogComponent } from '../component-info-dialog/component-info-dialog.component';
+
+@Component({
+  selector: 'component-selector',
+  templateUrl: './component-selector.component.html',
+  styleUrls: ['./component-selector.component.scss']
+})
+export class ComponentSelectorComponent {
+  @Output() componentSelectedEvent: EventEmitter<any> = new EventEmitter<any>();
```
hirokiterashima: No need to emit any value since this selector is already for a specific component type?
```suggestion
  @Output() componentSelectedEvent: EventEmitter<void> = new EventEmitter<void>();
```

src/assets/wise5/authoringTool/components/component-selector/component-selector.component.ts
```diff
@@ -0,0 +1,43 @@
...
+        component: new ComponentFactory().getComponent(this.previewContent, 'node1'),
+        description: this.description,
+        label: this.label
+      },
+      panelClass: 'dialog-lg'
+    });
+  }
+
+  protected select(): void {
+    this.componentSelectedEvent.emit(this.componentTypeObject);
```
hirokiterashima: No need to emit any value since this selector is already for a specific component type?
```suggestion
    this.componentSelectedEvent.emit();
```

src/assets/wise5/authoringTool/components/component-selector/component-selector.component.ts
```diff
@@ -0,0 +1,43 @@
...
+  private componentTypeObject: any;
+  private description: string;
+  protected label: string;
+  private previewContent: any;
+
+  constructor(private componentInfoService: ComponentInfoService, private dialog: MatDialog) {}
+
+  ngOnInit(): void {
+    this.label = this.componentInfoService.getLabel(this.componentType);
+    this.componentTypeObject = { type: this.componentType, name: this.label };
```
hirokiterashima: Not used, remove?
```suggestion
```

src/assets/wise5/authoringTool/components/component-selector/component-selector.component.ts
```diff
@@ -0,0 +1,43 @@
...
+
+@Component({
+  selector: 'component-selector',
+  templateUrl: './component-selector.component.html',
+  styleUrls: ['./component-selector.component.scss']
+})
+export class ComponentSelectorComponent {
+  @Output() componentSelectedEvent: EventEmitter<any> = new EventEmitter<any>();
+  @Input() componentType: string;
+  private componentTypeObject: any;
```
hirokiterashima: Not used, remove?
```suggestion
```

src/assets/wise5/authoringTool/components/component-selector/component-selector.component.ts
```diff
@@ -0,0 +1,43 @@
...
+  private description: string;
+  protected label: string;
+  private previewContent: any;
```
hirokiterashima: Does this need to be a class variable, or should it be declared only where it's used, in preview() function?
```suggestion
```

src/assets/wise5/authoringTool/components/component-selector/component-selector.component.ts
```diff
@@ -0,0 +1,43 @@
...
+@Component({
+  selector: 'component-selector',
+  templateUrl: './component-selector.component.html',
+  styleUrls: ['./component-selector.component.scss']
+})
+export class ComponentSelectorComponent {
+  @Output() componentSelectedEvent: EventEmitter<any> = new EventEmitter<any>();
+  @Input() componentType: string;
+  private componentTypeObject: any;
+  private description: string;
```
hirokiterashima: Does this need to be a class variable, or should it be declared only where it's used, in preview() function?
```suggestion
```

src/assets/wise5/authoringTool/components/component-selector/component-selector.component.ts
```diff
@@ -0,0 +1,43 @@
...
+  protected label: string;
+  private previewContent: any;
+
+  constructor(private componentInfoService: ComponentInfoService, private dialog: MatDialog) {}
+
+  ngOnInit(): void {
+    this.label = this.componentInfoService.getLabel(this.componentType);
+    this.componentTypeObject = { type: this.componentType, name: this.label };
+    this.description = this.componentInfoService.getDescription(this.componentType);
+    this.previewContent = this.componentInfoService.getPreviewContent(this.componentType);
```
hirokiterashima: Move these variable initializations to preview() function as local variables?
```suggestion
```

src/assets/wise5/components/ComponentInfo.ts
```diff
@@ -0,0 +1,12 @@
+export abstract class ComponentInfo {
+  abstract description: string;
+  abstract previewContent: any;
```
hirokiterashima: Can these be declared as protected fields? Also in all the child classes of ComponentInfo?
```suggestion
  protected abstract description: string;
  protected abstract previewContent: any;
```

hirokiterashima (dismissed):
Good start. I think the general layout and behavior that we want are all there. Let's discuss the actual text and example(s) for each component together as a group.

I added some code change suggestions inline. Let me know what you think. 

I'm also wondering if the code would be cleaner if the ComponentInfo class has description, label, and previewContent fields, and add a ```getInfo(componentType)``` function to ComponentInfoService so you query the ComponentInfo object for those fields, like this:
```
const info = ComponentInfoService.getInfo(componentType)
const label = info.getLabel();
const description = info.getDescription();
const previewContent = info.getPreviewContent();
```

Then we won't need ```ComponentInfoService.get[Label/Description/PreviewContent]()``` functions.

hirokiterashima:
Changs look good. 

I'm wondering whose job it is to get ComponentInfo and Component instances for the specific componentType: is it ComponentSelectorComponent, or ComponentInfoDialog? 

Right now, ComponentSelectorComponent does those things, but I wonder if it should just be passing the componentType  as the sole data to the ComponentInfoDialogComponent, and let the ComponentInfoDialogComponent handle the the instantiation of those objects.
```
(modified)
this.dialog.open(ComponentInfoDialogComponent, {
   data: this.componentType
   panelClass: 'dialog-lg'
});
```

The benefit of this approach is that it prevents the ComponentSelectorComponent (and other future components that will need to launch ComponentInfoDialogComponent) from haven't to do extra computation, making it a light-weight, pass-through component (input is componentType and output is also componentType)

What do you think?

breity (approved):
Looks good. 👍 

I updated the component type descriptions and modified the examples to use content from our library units.

I notice there are several component types that don't work properly or don't show any content in preview mode (Discussion, Peer Chat, Show Group Work, Show My Work, Summary). We should create issues to make these show preview content that users can interact with.

hirokiterashima (approved):
LGTM.

It might be good to merge develop and test locally first, just to make sure that this change set will work with Angular 16.

hirokiterashima (changes requested):
Actually, we found that copying multiple components in the step view tries to make a request to
```
http://localhost/api/author/config/NaN 400 (Bad Request)
```

And this error is thrown in the js console:
```
ERROR Error: Uncaught (in promise): HttpErrorResponse: {"headers":{"normalizedNames":{},"lazyUpdate":null},"status":400,"statusText":"OK","url":"http://localhost/api/author/config/NaN","ok":false,"name":"HttpErrorResponse","message":"Http failure response for http://localhost/api/author/config/NaN: 400 OK","error":null}
    at resolvePromise (zone.js:1211:31)
    at resolvePromise (zone.js:1165:17)
    at zone.js:1278:17
    at _ZoneDelegate.invokeTask (zone.js:406:31)
    at core.mjs:23896:55
    at AsyncStackTaggingZoneSpec.onInvokeTask (core.mjs:23896:36)
    at _ZoneDelegate.invokeTask (zone.js:405:60)
    at Object.onInvokeTask (core.mjs:24197:33)
    at _ZoneDelegate.invokeTask (zone.js:405:60)
    at Zone.runTask (zone.js:178:47)
```

Can you take a look?

hirokiterashima (approved):
LGTM

PR #1483, opened 2023-10-20
feat(Export): Include Question Used column in Peer Chat export

### LINKED ISSUE #1482: feat(Export): Include Question Used column in Peer Chat export
If a Peer Chat has "Student can select questions to use in Peer Chat" enabled, in the export, include a column that shows the question text the student used for their response (if any). If "Student can select questions to use in Peer Chat" is not enabled, do not include the column.

geoffreykwan:
## Changes

- If "Student can select questions to use in Peer Chat" is enabled, include "Question Used" column in the export
- Now only show either "Prompt" or "Dynamic Prompt" column, not both. Previously we would show both.

## Test

- If "Student can select questions to use in Peer Chat" is enabled, show the "Question Used" column
- If a student used a question for their response, make sure it shows the question they used in the "Question Used" column
- If dynamic prompt is enabled, make sure the "Dynamic Prompt" column shows up and the "Prompt" column does not
- If dynamic prompt is not enabled, make sure the "Prompt" column shows up and the "Dynamic Prompt" column does not

Closes #1482


breity (commented):
Functionality looks good.

There is some duplication added. One way you could simplify might be to combine each of insert column functions that are very similar into two functions, `insertPromptColumns` and `insertQuestionColumns`. So, for example, the logic for each of the 4 separate insert prompt column functions would just go in one (`insertPrompColumns`). What do you think?

breity (approved):
Changes look good.

PR #1484, opened 2023-10-23
refactor(AnnotationService): Move isThereAnyScoreAnnotation() to NodeInfo

hirokiterashima:
## Changes
- Move ```AnnotationService.isThereAnyScoreAnnotation()``` to ```NodeInfoComponent.hasScoreAnnotation()``` (only place where it's used) and rewrite the body using declarative syntax 
- Extract common code to ```AnnotationService.getAnnotationsByNodeIdComponentId()```
- Add private modifier, function param types and return types

## Test
- In Grading Tool > Node Info dialog, Teacher Summary Display appears as before if a component has summary view and student work


geoffreykwan (changes requested):
If a component has a score or auto score, the Step Info dialog should show a bar graph of the score distribution for the students in the period. Right now it does not show any graph.

I think in NodeInfoComponent.hasScoreAnnotation() you either need to return the value in the some() function or remove the braces.

geoffreykwan (approved):
Looks good.

PR #1486, opened 2023-10-24
feat(AuthoringTool): Move preview unit button to top bar

hirokiterashima:
## Notes
- Please edit as you see fit. For example, Are both i18n-aria-label and i18n-matTooltip needed on the button?
```
      <button
        mat-icon-button
        aria-label="Preview Unit"
        i18n-aria-label
        matTooltip="Preview Unit"
        i18n-matTooltip
        (click)="previewProject()"
      >
        <mat-icon>visibility</mat-icon>
      </button>
```

## Changes
- Move preview unit button to the top bar (before, it was embedded in the unit editing view and was not accessible from info, assets, notebook, etc)

## Test
- Clicking on the preview unit button launches the preview in a separate window

### DIFF
--- a/src/assets/wise5/authoringTool/components/top-bar/top-bar.component.html
+++ b/src/assets/wise5/authoringTool/components/top-bar/top-bar.component.html
@@ -20,6 +20,16 @@
       </button>
       <button
         mat-icon-button
+        aria-label="Preview Unit"
+        i18n-aria-label
+        matTooltip="Preview Unit"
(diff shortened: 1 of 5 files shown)


breity (approved):
Looks good.

I removed redundant `aria-label` attributes for buttons with tooltips. Material tooltips already add accessibility for screen readers.

hirokiterashima:
Thanks. That makes sense.

PR #1489, opened 2023-10-25
fix(Authoring): Recovery view does not load when project is broken

### LINKED ISSUE #1488: bug(Authoring): Recovery view does not load when project is broken
Try to load a broken project in the recovery view. The path to the recovery view looks like

```
http://localhost:81/teacher/edit/unit/<insert project id here>/recovery
```

One way to break a project is to have a group node reference a non-existing step node id in its ids array.

geoffreykwan:
## Changes

- Added specific route for the recovery view with its own resolver
- Removed null check for projectURL because I couldn't find where this would occur

## Test

- Make sure recovery view works when the project is broken. You can break a project by adding a non-existing step node id to the ids array of a group node.
- Make sure projects load normally in the VLE, Authoring, Classroom Monitor

Closes #1488

### DIFF
--- a/src/app/services/projectService.spec.ts
+++ b/src/app/services/projectService.spec.ts
@@ -43,7 +43,6 @@ describe('ProjectService', () => {
   shouldReplaceAssetPathsInHtmlComponentContent();
   shouldNotReplaceAssetPathsInHtmlComponentContent();
   shouldRetrieveProjectWhenConfigProjectURLIsValid();
-  shouldNotRetrieveProjectWhenConfigProjectURLIsUndefined();
   shouldGetDefaultThemePath();
   shouldReturnTheStartNodeOfTheProject();
   shouldReturnTheNodeByNodeId();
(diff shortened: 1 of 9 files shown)


hirokiterashima:
Part of the original problem seems to be due to the fact that we're trying to load the unit editing view, which requires a resolver to retrieve the unit (and NOT parse it in the case of recovery). 

Should the recovery view even be under the ```/teacher/edit/unit/:unitId``` path anymore, or should it be under another path, e.g., ```/teacher/edit/recovery/:unitId```? Would this allow us to avoid the issues? We don't need all the other tools (milestone, info, assets, etc) when we're in the recovery mode, so would it make sense to separate it out?

hirokiterashima (approved):
LGTM.

I think we can clean up the projectService retrieve/parse project sequence a bit in the future. Something feels a bit off about the ```setProject()``` function that sets the project and also parses it.

PR #1490, opened 2023-10-25
refactor(Classroom Monitor): Clean up getStudentProjectCompletion()

geoffreykwan:
## Changes

- Cleaned up getStudentProjectCompletion()
- When initializing students, only look at workgroups that have a workgroup id
- Removed getWorkgroupIdsOnNode() function which was no longer used

## Test

- Make sure the Grade By Team view (where it shows all the workgroups) works and shows the unit completion for the workgroups
- Make sure the workgroup grading view (where it shows just one workgroup) works and shows the completion percentage

### DIFF
--- a/src/assets/wise5/classroomMonitor/student-progress/student-progress.component.ts
+++ b/src/assets/wise5/classroomMonitor/student-progress/student-progress.component.ts
@@ -73,7 +73,7 @@ export class StudentProgressComponent implements OnInit {
 
   private updateTeam(workgroupId: number): void {
     const location = this.getCurrentNodeForWorkgroupId(workgroupId);
-    const completion = this.getStudentProjectCompletion(workgroupId);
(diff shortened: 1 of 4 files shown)


src/assets/wise5/classroomMonitor/student-progress/student-progress.component.ts
```diff
@@ -99,7 +101,7 @@ export class StudentProgressComponent implements OnInit {
    * @return object with completed, total, and percent completed (integer between 0 and 100)
    */
   private getStudentProjectCompletion(workgroupId: number): any {
```
hirokiterashima: Remove this function definition and call ```this.classroomStatusService.getStudentProjectCompletion(workgroupId)``` directly in ```updateTeam()```?

hirokiterashima (approved):
LGTM. See inline comment before merging.

PR #1497, opened 2023-10-27
refactor(AnnotationService): Move getAllLatestScoreAnnotations() to MilestoneReportService

hirokiterashima:
## Changes
- Move AnnotationService.getAllLatestScoreAnnotations() to MilestoneReportService, the only place it's used
- Rewrite the function using filter() and reduceRight()
- Add function param and return types and make function private

## Test
- Milestone TAP report chooses and displays the correct report template as before

### DIFF
--- a/src/assets/wise5/services/annotationService.ts
+++ b/src/assets/wise5/services/annotationService.ts
@@ -640,26 +640,6 @@ export class AnnotationService {
     return this.annotations.filter((annotation) => annotation.studentWorkId === studentWorkId);
   }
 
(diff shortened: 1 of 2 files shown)


geoffreykwan (changes requested):
Could you reduce the complexity of the function some more?

You can also probably call AnnotationService.isScoreOrAutoScore(annotation) or AnnotationService.matchesScoreType(annotation, 'any') to check if it's a score annotation.

geoffreykwan (approved):
Looks good.

PR #1498, opened 2023-10-27
refactor(NodeService): Remove currentNodeHasTransitionLogic()

hirokiterashima:
## Changes
- Remove calls to NodeService.currentNodeHasTransitionLogic() and the function itself. This is because every node should have a transition logic field. If it doesn't, it's a problem with the unit and the behavior is unknown.

## Test
- Units load and work as before in student VLE and preview


geoffreykwan (approved):
Looks good.

PR #1502, opened 2023-10-31
refactor(NodeService): Move evaluateTransitionLogicOn() to Node

hirokiterashima:
## Changes
- Move NodeService.evaluateTransitionLogicOn() to Node.isEvaluateTransitionLogicOn()
- Clean up isEvaluateTransitionLogicOn()
   - Remove null-check (this function was always called when there was a currentNode)
   - Reduce return statement from 2 to 1
   - Add function param and return types

## Test


geoffreykwan (approved):
Looks good.

PR #1503, opened 2023-11-01
refactor(Classroom Status Service): Clean up getNodeCompletion()

geoffreykwan:
## Changes

- Reduced complexity of getNodeCompletion() function
- Added tests

## Test

Make sure the completion is calculated correctly in the "Grade by Step" and "Grade by Team" views.


hirokiterashima (approved):
LGTM. 

```getNodeCompletion()``` is easier to follow than before and the completion calculation results look correct in my local testing.

We can address new CC issues created by this change set in a future task.

PR #1504, opened 2023-11-01
refactor(NodeService): Clean up evaluateTransitionLogic()

hirokiterashima:
## Changes
- Remove null-check for currentNode, since it will always be set to something when this function is called.
- Remove unused variables (transition, fromNodeId, toNodeId) and empty else case
- Remove redundant comments
- Move variable declaration closer to use
- Add function return type

## Test


geoffreykwan (approved):
Looks good.

PR #1505, opened 2023-11-01
fix(Site): Search shows two x buttons

### LINKED ISSUE #1405: bug(Site): Search unit input shows two x buttons
In these locations
- Teacher Run Search
- Teacher Unit Library Search
- Student Run Search

Enter some text in the search input and place your mouse over the input and you will see two x buttons appear. Both x buttons will work properly to clear the text. We only need the white x button to appear.

geoffreykwan:
## Changes

Removed type="search" from the search input to get rid of the extra x button.

## Test

Make sure the search still works in these locations
- Teacher Run Search
- Teacher Unit Library Search
- Student Run Search

Enter some text in the search input and only the white x button should appear. Previously two x buttons would appear like in the screenshot below.

Screenshot 2023-11-01 152039

Closes #1405


breity (approved):
Looks good. 👍

PR #1508, opened 2023-11-05
fix(Authoring): Concurrent authors message not updating on log out

### LINKED ISSUE #1423: Concurrent authors message do not update when other logs out
I noticed that if two teachers are authoring the same project at the same time and one of them signs out, the other teacher will still see the message. If they refresh the Authoring Tool, the message goes away.

geoffreykwan:
## Changes
- When the user logs out, send the author project end message to remove their name from the concurrent authors

## Test
1. Sign in as two teachers that are authoring the same project
2. Make sure the concurrent authors message updates when one teacher
- Refreshes the page
- Clicks the "Go Home" button and then the browser back button
- Clicks the "Sign Out" button

Closes #1423


hirokiterashima (approved):
LGTM

PR #1511, opened 2023-11-07
refactor(NodeService): evaluateTransitionLogic()

hirokiterashima:
## Changes
- Change the way current node is retrieved, so we can expect a non-null transitionLogic
- Consolidate cases when we need to call ```chooseTransition()```

## Test


geoffreykwan (approved):
Looks good.

PR #1514, opened 2023-11-13
refactor(Export): Create open response component export strategy

geoffreykwan:
## Changes
- Created OpenResponseComponentExportStrategy
- Removed Open Response export code from data-export.component.ts
- Refactored some common code in the Discussion and Peer Chat component exports

## Test
Make sure these exports still work properly
- Open Response
  - with single auto score
  - with multiple auto sub scores and ideas
- Discussion
- Peer Chat


hirokiterashima:
Seems to be working, but this is a lot of changes, which makes it a bit hard for me to follow/review. Was the extraction of common code in the Discussion and Peer Chat component exports a necessary part of creating OpenResponseComponentDataExportStrategy, or can it be done in a separate refactoring/review task?

Also, should ComponentDataExportStrategy's filename be ComponentDataExportStrategy.ts?

geoffreykwan:
Yea I misnamed the file but now it should be fixed.

I saw duplication in Code Climate so I refactored the common code. I could have just done that in a different issue.

hirokiterashima (approved):
Ok new changes look good.

Yah, this would have been better if it was broken into two PR's. Going forward, let's try to limit the changes so we can perform better reviews.

Issue #1513, opened 2023-11-08
Multiple languages per unit


hirokiterashima:
### Description

Add ability to let users view a unit in multiple languages. Right now, if you want to use a unit in your classroom in 2 languages, you would need to create two runs (one for each language), and remember to assign the right run to the right students to match their language preferences. With this new feature, you would just create one run and WISE would automatically show the right language based on the students’ preferred language.

Mockups:
<img width="829" alt="image" src="https://github.com/WISE-Community/WISE-Client/assets/119416/edd00528-73cc-418c-8040-c2087f54d468">

PR #1520, opened 2023-11-14
feat(Student,Preview): Switch between unit languages

hirokiterashima:
## Notes
- This will merge to the #1513 feature branch
- Known issue: notebook content does not get translated

## Changes
- Implement switching between languages of the unit content in in student and preview
- Add language select drop-down if project has locale settings

## Test Prep
1. In a unit that is used for a run, add the following block to the project's metadata:
```
                "locale": {
                        "default": "en_US",
                        "supported": [
                                "en_US",
                                "es",
                                "ja"
                        ]
                }
```

2. Add ```"[PropertyName].i18nId": "MD5 Hash of [PropertyValue]"``` for each text that you want to make available in other languages. For example:
```
   "type": "OpenResponse",
   "prompt": "What do you think about Global Warming?",
   "prompt.i18nId": "905d0b11f1653624e48d5cd058873463",
```
In this case, "905d0b11f1653624e48d5cd058873463" is the MD5 hash of the text "What do you think about Global Warming?". You can use any MD5 Hash generator tool online.

3. For each supported locale, create ```project.[locale].json``` file with ```MD5 Hash of [PropertyValue]: Translated text of [PropertyValue]``` mappings. For example, in project.es.json, we'd add this mapping:
```
  "905d0b11f1653624e48d5cd058873463": "¿Qué opinas sobre el calentamiento global?"
```
4. Repeat steps 2-3 and try translating many places in the unit (e.g. unit title, step title, choice text, etc)

## Test
While logged in as a student or previewing a unit, test that:
- Languages drop-down appears only if the unit has at least 2 supported locales


src/assets/wise5/services/translateProjectService.ts
```diff
@@ -0,0 +1,59 @@
...
+          .get(this.getTranslationMappingURL(locale), {
+            headers: new HttpHeaders().set('cache-control', 'no-cache')
+          })
+          .pipe(
+            tap((translations: any) => {
+              this.applyTranslations(project, translations);
+              this.projectService.setProject(project);
+            })
+          )
+      : of({});
```
geoffreykwan: Can the pipe/tap be replaced with subscribe() and then the caller doesn't need to subscribe()? Then you also don't need the of({}) case since the caller doesn't need to subscribe().

src/assets/wise5/services/translateProjectService.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  private applyTranslations(projectElement: any, translations: any): void {
+    Object.keys(projectElement)
+      .filter((key) => key.endsWith('.i18nId'))
+      .forEach((key) => {
+        const keyWithoutI18NId = key.substring(0, key.lastIndexOf('.i18nId'));
+        const translationKey = projectElement[key];
+        projectElement[keyWithoutI18NId] =
+          translations[translationKey] ?? projectElement[keyWithoutI18NId];
```
geoffreykwan: Isn't this setting projectElement[keyWithoutI18NId] to itself projectElement[keyWithoutI18NId] in the else case?

You can just do
```
if (translations[translationKey]) {
  projectElement[keyWithoutI18NId] = translations[translationKey];
}
```

geoffreykwan (changes requested):
Changing languages mostly works.

The first time I go to a new step, none of the components load. If I leave the step and come back the components will load.

src/assets/wise5/services/translateProjectService.ts
hirokiterashima: Yes, we can, but an issue may arise when we add code in the future that calls translate() from ```InitializeVLEService.initialize[Student|Preview]()``` while the Student/Preview views are bootstrapping. In this case, I was imagining that we'd want to do these in order:
1. Retrieve project (subscribe to this call to make sure project is retrieved before moving on)
2. Translate project (subscribe to this call to make sure project is translated before moving on)
3. Initialize everything else

To make the above possible, this translate() function would need to return an Observable (so we can subscribe to it in ```InitializeVLEService.initialize[Student|Preview]()```). 

The other option is to say that the translation in step 2 does not have to happen before step 3, that it can run as part of the "Initialize everything else" sequence. I'm not sure what the ramifications of that would be on the application (e.g. will it cause the page to flicker?), but we can try it?

src/assets/wise5/services/translateProjectService.ts
hirokiterashima: Good suggestion. I changed the code.

src/assets/wise5/services/translateProjectService.ts
hirokiterashima: As we discussed as a group, I changed the translate() func to return a Promise instead of an Observable. This removed the need to ```subscribe()```. This should also let us ```await``` for it in InitializeVLEService in the future.

hirokiterashima:
I fixed the issue where Notebook report starter content was not updating. Here's the new behavior:
- if you have NOT saved any changes to the notebook report, it should translate.
- if you have saved any change to the notebook report, it should NOT translate.

This should work in student VLE and preview.

@geoffreykwan I think I addressed all of the issues you found? PTAL.

geoffreykwan (changes requested):
The notebook report translation works now. The notes don't seem to be translated though. This includes things like
- Tooltip for Notes button at the bottom right
- Tooltip for the Add note button when the notes sidebar is open
- The label at the top of the notes sidebar

hirokiterashima:
@geoffreykwan good catch. We now translate these places:
   - Tooltip for Notes button at the bottom right
   - Tooltip for the Add note button when the notes sidebar is open
   - The label at the top of the notes sidebar
   - Text in edit/delete note item dialog

PTAL

geoffreykwan (approved):
Looks good.

breity (changes requested):
Looks great for the most part!

I found a case where certain content in the project.json file can break switching to a different language - when there is a "null" value in an array in some part of the unit content. For example, this series data for a graph component:
```
"series":[{"data":[0,null],"color":"green","marker":{"symbol":"circle"},"canEdit":false,"name":"Future"}]
```

When I try to switch to a different language, the `[0,null]` value for the `data` field in that series causes the following error:
```
ERROR Error: Uncaught (in promise): TypeError: Cannot convert undefined or null to object
TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at TranslateProjectService.applyTranslations (translateProjectService.ts:45:12)
...
```

hirokiterashima:
@breity in this case, how did you get ```{"data":[0,null]...``` in your graph content? When I author the graph component, the ```data``` field is a 2D array, like ```{"data":[[0,0],[]]...```. When I remove a data point, it removes the inner-array, so removing the first data point changes it to ```{"data":[[]]```. 

Also, can you think of other cases when there is a "null" value in an array in some part of the unit content? I'm not sure if it's by design (if so, we should handle it in ```applyTranslations()```), or if it's an indication of a application/user error (if so, we should prevent it from occuring).

breity:
Yeah, as we discussed, looks like this happens when you author a categorical graph component and add a data point but don't fill in the value for one of the categories. So we should probably handle this case in the graph authoring (default to 0, perhaps) to prevent having a null value in the series data array. @hirokiterashima will create a separate issue and I'll approve this PR.

breity (approved):

PR #1521, opened 2023-11-14
fix(Table): Don't truncate overflow Tabulator cell content

breity:
## Changes
Tabulator cells with content longer than the width of the cell no longer truncate the cell text.

Previous behavior:
Screenshot 2023-11-13 at 8 52 04 PM

New behavior:
Screenshot 2023-11-13 at 8 52 42 PM

## Test
- Author a Table component with some long text in the cells.
- Ensure that all the content in the cells with long text is visible and the row height adjusts to accommodate (text wraps to new lines when necessary).


hirokiterashima (approved):
LGTM

PR #1526, opened 2023-11-15
build(Angular): Remove custom-webpack build dependency

### LINKED ISSUE #1525: build(Angular): Remove custom-webpack build dependency
### Description

Change
```@angular-builders/custom-webpack:[browser,dev-server,karma,etc]```
to
```@angular-devkit/build-angular:[browser,dev-server,karma,etc]```

Move away from using webpack to build our applications so we can take advantage of build performance optimizations in recent releases of Angular, which uses esbuild and vite.

hirokiterashima:
## Notes
It looks like we no longer need to use the custom-webpack builder. I can't recall exactly why we needed it in the first place, but it may have something to do with when we were using the SVGEditor for our drawing tool.

This change will allow us to move away from using Webpack to build our applications and we'll be able to take advantage of build performance optimizations in recent releases of Angular (which uses esbuild and vite).

## Changes
- Remove @angular-builders/custom-webpack dependency from package.json
- Change build from 
```@angular-builders/custom-webpack:[browser,dev-server,karma,etc]``` to 
```@angular-devkit/build-angular:[browser,dev-server,karma,etc]```

## Test Prep
- Remove node_modules dir (you might want to save a local copy, just in case)
- Run ```npm i``` to install dependencies

## Test
1. These workflows work as before
- start dev server
- extract i18n (i.e., generate messages.xlf)
- run unit tests
- build production locally (via ```npm run build-prod```) => should create artifacts in dist directory

2. WISE works locally as before, in particular:
- components that use svg, like concept map, label, animation

Closes #1525


geoffreykwan (approved):
Looks good.

breity (approved):
Looks good.

PR #1529, opened 2023-11-16
feat(Authoring): Allow switching between components in component info

### LINKED ISSUE #1516: Show component info: Switch between multiple components
### Description

In the show component info dialog, add controls (drop-down, arrows) to let user switch to other components. This will let them quickly learn about other components without having to close and open each component. 

First create mockup to discuss design options.

geoffreykwan:
## Changes
- Added drop down to switch between component types
- Added previous and next buttons to switch between component types
- Added component info label
- Added description label
- Added some indentation to differentiate labels from content

## Test
- Make sure the drop down switches components
- Make sure previous and next buttons switch components

Closes #1516


hirokiterashima (commented):
Functionality works well and the style changes look good. New navigation controls make it easy to switch between components.

While reading the code, I thought that the ComponentInfoDialog is getting bigger and more complex (e.g. has new responsibilities, which may be misplaced). I'm wondering if it'd be better if we created ComponentTypeSelectorComponent that has the prev/next button and the drop-down, and emitted the newly-selected component type to ComponentInfoDialogComponent? 

With this change,
- it'd be up to the ComponentTypeSelectorComponent to calculate the prev/next component, firstComponent, lastComponent, etc. It'll emit  newly-selected component type back to ComponentInfoDialogComponent
- the ComponentInfoDialogComponent can stay mostly the same as before. It just needs to embed the <component-type-selector> in its template and respond when ComponentTypeSelectorComponent emits a newly-selected component type.

What do you think?

geoffreykwan:
I created the ComponentTypeSelectorComponent and renamed an existing component from ComponentSelector to ComponentTypeButton.

hirokiterashima (approved):
LGTM.

It looks like you can remove changeComponentType() and call displayComponent() directly from the template? Also, you can remove the ```ComponentInfoDialogComponent.label``` field since we don't use it anymore.

PR #1531, opened 2023-11-20
refactor(NodeComponent): createComponentStatesResponseHandler()

hirokiterashima:
## Changes
- Remove getDataArraysToSaveFromComponentStates() and set variables directly
- Clean up logic when translation logic should be evaluated using ```some()```

## Test
- Branching based on auto-score c-rater item (e.g. score 1 => branch 1, score 2 => branch 2, etc) works as before.
- Saving component state works as before


geoffreykwan (approved):
Looks good.

PR #1532, opened 2023-11-21
refactor(StudentStatusService): Clean up code

hirokiterashima:
## Changes
- Don't call retrieveStudentStatus() if preview mode. Start with an empty StudentStatus instead.
- Remove setStudentStatus() and set field directly
- Make studentStatus field private.
- Only call saveStudentStatus() in student mode. Do mode check before function is called.
- Change saveStudentStatus() return type to void

## Test
- Student status retrieval/saving works as before in student mode
   - project completion saves 
- Student status works as before in preview mode
   - does not retrieve or save student status (as before)


geoffreykwan (approved):
Looks good.

PR #1535, opened 2023-11-21
refactor(NodeAuthoring): Componentize EditComponentAdvancedButton

hirokiterashima:
## Changes
- Extract EditComponentAdvancedButton from NodeAuthoring into a new component
   - With this change, launching the EditComponentAdvanced dialog will no longer be NodeAuthoringComponent's responsibility 

## Test
- Edit Component Advanced button works as before
   - It should appear only when the component authoring view is expanded
   - Clicking on the button should launch the edit component advanced dialog
   - Clicking on the button should not collapse the component authoring view


geoffreykwan (approved):
Looks good.

PR #1536, opened 2023-11-22
refactor(NodeAuthoring): Convert componentsToChecked to signal

hirokiterashima:
## Changes
- Convert componentsToChecked and isAnyComponentSelected fields to signals and make them protected
- Update tests

## Test (in node authoring)
- checking/unchecking a component checkbox shows/hides the contextual buttons (move, copy, delete)
   - as long as one component is checked, the contextual buttons should show 
- move/copy/delete single/multiple components work as before using contextual buttons and component-specific buttons


geoffreykwan (approved):
Looks good.

PR #1539, opened 2023-11-29
fix(Graph Authoring): Set categorical graph data point to default to 0

### LINKED ISSUE #1538: Categorical graph component saves null data point
### Description

Categorical graph component saves null data point, when it should instead save 0 or not save the data point altogether

### Steps to Reproduce

1. Author a categorical graph component and add a data point but don't fill in the value for one of the categories. 

**Expected behavior:**
Empty data point should be saved as "0"? (or don't save data point altogether)?

**Actual behavior:**
Empty data point is saved as null

### Additional Information
This issue was found during the review of this PR: https://github.com/WISE-Community/WISE-Client/pull/1520#pullrequestreview-1751676466

geoffreykwan:
## Changes
When a categorical graph data point is added, it now defaults to 0 instead of null.

## Test
2. Create a Graph step
3. Set the Plot Type to Column Plot
4. Scroll down to the Data Points and add a data point. The value should default to 0 instead of null.

Closes #1538

### DIFF
--- a/src/assets/wise5/components/graph/graph-authoring/graph-authoring.component.ts
+++ b/src/assets/wise5/components/graph/graph-authoring/graph-authoring.component.ts
@@ -247,7 +247,7 @@ export class GraphAuthoring extends AbstractComponentAuthoring {
       ) {
         series.data.push([]);
       } else if (this.componentContent.xAxis.type === 'categories') {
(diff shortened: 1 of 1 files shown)


hirokiterashima (approved):
LGTM

PR #1542, opened 2023-11-29
refactor(NodeAuthoring): Extract AddComponentButtonComponent

hirokiterashima:
## Changes
- Extract AddComponentButtonComponent from NodeAuthoring component and template. This reduces NodeAuthoringComponent's responsibility.
   - AddComponent dialog is now launched from AddComponentButtonComponent instead of NodeAuthoringComponent
   - Result from AddComponent dialog (chosen component type or chosen import component(s)) is now saved by AddComponentButtonComponent and sent back to NodeAuthoring via newComponentsEvent output
- Add tests for AddComponentButtonComponent

## Test (in Node Authoring)
- Add new component(s) and import component(s) work as before
   - insert as the first component of the node
   - insert after an existing component


geoffreykwan (changes requested):
Functionality works well.

The add-component-button test is misleading because it makes it sound like clicking on the button will open the dialog, create the component, and save the project. In actuality, clicking the button will only open the dialog. It might be better to separate it into two tests. One tests opening the dialog and another tests when the dialog is closed, it creates the component and saves.

geoffreykwan (approved):
Looks good.

PR #1543, opened 2023-11-29
fix(Table): Properly calculate row heights after build

breity:
## Changes
Properly calculates Tabulator row heights for certain table components when the table is built.

[Production example](https://wise.berkeley.edu/preview/unit/41025/node223) where row heights are not properly sized:
Screenshot 2023-11-29 at 11 24 19 AM

With this fix, example table renders like this:
Screenshot 2023-11-29 at 11 26 38 AM

## Test
Make sure Table component row heights are properly calculated and Table components work as before.


hirokiterashima (approved):
LGTM

PR #1546, opened 2023-12-01
feat(Edit Project Info): Multiple languages setting

hirokiterashima:
## Notes
- Don't worry about styling for this PR- we can do that in a separate branch. 

## Changes
- Create EditProjectLanguageSettingComponent to allow authors to specify default language and add/delete supported languages for the project.

## Test (in Unit Authoring > Info view)
- default language 
   - displays previously-chosen value
      - for units without a default language (e.g., older units), it should show "English"
   - can be changed, and saves immediately upon change
   - must always have 1 default language (i.e., can't be empty) 
- supported languages 
   - displays previously-chosen value
      - for units without a default language (e.g., older units), it should show empty
   - can be changed (add/delete languages), and saves immediately upon change  

## Test (in unit preview and student VLE views)
- Supported languages options appear in the "Language" dropdown at the top-right corner of the page

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,6 +21,7 @@
         "@angular/platform-browser": "^16.2.9",
         "@angular/platform-browser-dynamic": "^16.2.9",
         "@angular/router": "^16.2.9",
+        "@ng-select/ng-select": "^11.2.0",
         "@ng-web-apis/common": "^2.0.1",
         "@ng-web-apis/intersection-observer": "^3.0.0",
         "@stomp/rx-stomp": "^1.1.4",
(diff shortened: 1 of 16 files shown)


geoffreykwan (changes requested):
Here are some things I noticed.

1. Language settings do not get saved if the project does not have a metadata.locale field. All old and new projects do not have a metadata.locale field.

2. If the default language is not added to "additional languages", the default language will not show up in the student upper right language select so they won't be able to switch back to the default language.

3. Should the languages in the drop downs be ordered alphabetically?

4. Do we need to specify "Dutch (Netherlands)" or is "Dutch" by itself ok?

5. The "additional languages" select adding/deleting feels a little inconsistent. When you add a language, it closes the drop down. When you have the drop down open and delete a language by clicking it in the drop down, it keeps the drop down open.

Here are potential other configurations to consider.

[Here is an example of the drop down staying open when adding and deleting.](https://ng-select.github.io/ng-select#/multiselect) (use the first demo and open the drop down and delete and then add "Karyn Wright")

[Here is an example of options being removed from the drop down and only being able to delete using the x button.](https://ng-select.github.io/ng-select#/multiselect) (use the second demo and add and then delete "Karyn Wright")

hirokiterashima:
@geoffreykwan good catch. I think I've resolved all the issues you found. PTAL. Remember this branch still uses the old  project.json format using ```[field].i18nId``` keys.

geoffreykwan (approved):
Looks good.

breity (approved):
Looks good.

PR #1547, opened 2023-12-01
feature(VLE): Add support for custom unit styles

breity:
## Changes
- Adds support for custom unit styles via a `project_styles.css` file in the project assets directory.

Previously, we supported custom css via a `style` value in the project.json, but that was broken after converting from AngularJS -> Angular. This implementation uses a separate CSS file, which is easier to edit/manage. For now, only trusted authors can upload CSS files to a project's assets. This is good for security, since we need to bypass the Angular sanitizer to load the file. In the future, we'll add a custom CSS option to the authoring tool that will dynamically create and update the file so that authors don't need to upload the file, but can create rules using the authoring interface.

## Test
- Add a `project_styles.css` file with some rules to the project assets folder for a unit.
- Verify that the custom styles are applied to when previewing the unit and viewing as a student.

### DIFF
--- a/src/assets/wise5/vle/vle.component.html
+++ b/src/assets/wise5/vle/vle.component.html
@@ -1,6 +1,6 @@
 <link rel="stylesheet" href="/assets/wise5/themes/default/style/angular-material.css" />
 <link rel="stylesheet" href="/assets/wise5/themes/default/style/vle.css" />
-<style ng-bind-html="projectStyle"></style>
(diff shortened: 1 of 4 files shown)


hirokiterashima (approved):
LGTM

hirokiterashima (approved):
LGTM. Before you merge, can you do these:
1. Change fdescribe -> describe in vle.component.spec.ts
2. Generate messages.xlf again to reflect the new changes

PR #1548, opened 2023-12-04
Multiple languages: Change i18nId string to i18n object

hirokiterashima:
## Notes
- This will merge to the https://github.com/WISE-Community/WISE-Client/issues/1513 feature branch (not to the develop branch)
- The only file that changed is TranslateProjectService. Others are merge commits that for some reason still show up as being changed, even though the feature branch already contain those commits. I'm not quite sure how this happened.

## Changes
- Change ```i18nId: string``` to ```i18n: { id: string, modified: number }``` in project.json. This will allow us to compare modified times of the original and translation strings and offer better translation support in the future.
- Rename translation files from ```project.[locale].json``` to ```translations.[locale].json```

## Test Prep
0. (Optional) make a backup of the project that has translations, in case you need to review other "Multiple Language..." PR's using the old i18n format and want to easily switch project versions
1. In ```project.json```, change instances of
```
   "some_key.i18nId": "abc123",
```
to
```
   "some_key.i18n": { "id": "abc123", "modified": 123 },
```
2. Rename ```project.[locale].json``` to ```translations.[locale].json``` for all supported locales for the unit
3. In ```translations.[locale].json```, change instances of
```
   "abc123": "[translated text]",
```
to
```
   "abc123": {"value": "[translated text]", "modified": 345 },
```

## Test
- Switching between translations work as before. While logged in as a student or previewing a unit, test that:
   - Languages drop-down appears only if the unit has at least 1 supported locale

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -70,6 +70,7 @@
         "@babel/core": "^7.17.9",
         "@babel/preset-env": "^7.16.11",
         "@ngtools/webpack": "^16.2.6",
+        "@semantic-release/exec": "^6.0.3",
         "@semantic-release/github": "^8.0.4",
         "@types/dom-mediacapture-record": "^1.0.11",
         "@types/jasmine": "^4.3.1",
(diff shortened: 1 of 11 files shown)


geoffreykwan (approved):
The new structure works well.

I noticed that if a translation value is an empty string, then an empty string will show up for the student. This can either be handled by checking for empty string or just making sure a translation entry is removed if a translator clears out a translation value. This is something we'll want to handle eventually.

hirokiterashima:
@geoffreykwan good point. I think we should ensure that we don't save empty translation values in the AT. I'll keep that in mind when working on the AT translation feature.

breity (approved):
Looks good.

PR #1550, opened 2023-12-05
feat(Multiple Languages): Show ProjectLanguageChooser in AT

hirokiterashima:
## Note
- Remember to use the ```[field].i18n = { id: "abc123", "modified": 123 }``` project format for this PR and all future PR's.

## Changes
- Change ProjectLanguageChooserComponent so we can use it in the VLE and the AT
   - Take in ```ProjectLocale``` as ```@Input``` 
   - Emit selected ```Language``` as ```@Output```
   - Move to common folder
- Display ProjectLanguageChooserComponent in the AT's TopBarComponent if unit has translations
   - Update available language options when author changes them in the Project Info view

## Test AT
- ProjectLanguageChooserComponent displays supported languages iff unit has translations
- Editing language options in the Project Info view updates the ProjectLanguageChooserComponent's available options
   - If there are no available languages, the ProjectLanguageChooserComponent should not appear 

*Note: nothing should happen we switch the language option in the ProjectLanguageChooserComponent atm. We'll handle that in a future PR.

## Test Student VLE
- Switching between translations works as before. While logged in as a student or previewing a unit, test that:
   - Languages drop-down appears only if the unit has at least 1 supported locale


geoffreykwan (changes requested):
Mostly works but I found one issue.

In the Authoring Tool, if you have set the language in the top bar to one of the additional languages and then go to the Unit Info view and change the default language or add/delete an additional language, the language in the top bar changes back to the default language.

I think how we should handle this is if the author is in the Unit Info view and changes the default language or additional languages, the top bar language should not change. The top bar language should only change to the default language if the language that is currently selected in the top bar is removed from the additional languages.

hirokiterashima:
@geoffreykwan Good catch. We now set selected language to the default language only when the previously-selected option is no longer available (i.e., removed by author), or this is the first time that the component is displayed. PTAL.

src/app/common/project-language-chooser/project-language-chooser.component.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  ngOnChanges(): void {
+    this.availableLanguages = this.projectLocale.getAvailableLanguages();
+    if (
+      this.selectedLanguage == null ||
+      !this.availableLanguages.some((lang) => lang.locale === this.selectedLanguage.locale)
+    ) {
+      this.selectedLanguage = this.availableLanguages.find((lang) =>
+        this.projectLocale.isDefaultLocale(lang.locale)
+      );
```
geoffreykwan: Can this be changed to 
```
this.selectedLanguage = this.projectLocale.getDefaultLanguage();
```
now that there is a isSameLanguage() function that is used in [compareWith]?

src/app/common/project-language-chooser/project-language-chooser.component.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+  });
+
+  it('keeps selected language option when language option changes', async () => {
+    const options = await getOptions();
+    await options[1].click();
+    setProjectLocale(new ProjectLocale({ default: 'it', supported: ['de', 'fr', 'ja', 'es'] }));
+    const newOptions = await getOptions();
+    expect(await newOptions[3].isSelected()).toBeTrue();
+  });
+});
```
geoffreykwan: Instead of 
```
await options[1].click();
```
you could do
```
const selectHarness = await loader.getHarness(MatSelectHarness);
await selectHarness.clickOptions({ text: 'Japanese' });
```
and then you can get the selected value by doing
```
expect(await selectHarness.getValueText()).toEqual('Japanese');
```

src/app/common/project-language-chooser/project-language-chooser.component.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+  component.ngOnChanges();
+  fixture.detectChanges();
+}
+
+async function getOptions(): Promise<MatOptionHarness[]> {
+  const selects = await loader.getAllHarnesses(MatSelectHarness);
+  const selectComponent = selects[0];
+  await selectComponent.open();
+  return await selectComponent.getOptions();
+}
```
geoffreykwan: You can do this to get the first MatSelectHarness without having to get all of them and then getting the first one from the array.
```
const selectHarness = await loader.getHarness(MatSelectHarness);
```

src/app/common/project-language-chooser/project-language-chooser.component.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+  });
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(ProjectLanguageChooserComponent);
+    loader = TestbedHarnessEnvironment.loader(fixture);
+    component = fixture.componentInstance;
+    setProjectLocale(new ProjectLocale({ default: 'en_US', supported: ['ja', 'es'] }));
+  });
+
+  it('shows available languages and selects the default language', async () => {
```
geoffreykwan, line 28: Should this be wrapped in a describe that describes the initial state like ```describe('when the language chooser is first loaded')``` or something.

src/app/common/project-language-chooser/project-language-chooser.component.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+    const options = await getOptions();
+    expect(options.length).toEqual(3);
+    expect(await options[0].getText()).toMatch('English');
+    expect(await options[1].getText()).toMatch('Japanese');
+    expect(await options[2].getText()).toMatch('Spanish');
+    expect(await options[0].isSelected()).toBeTrue();
+  });
+
+  it('keeps selected language option when language option changes', async () => {
```
geoffreykwan, line 37: Should this be wrapped in a describe that describes the action like ```describe('when the language options change')``` or something.

geoffreykwan (approved):
Looks good. I have some minor suggestions.

src/app/common/project-language-chooser/project-language-chooser.component.ts
hirokiterashima: Yes, this works. I changed it.

src/app/common/project-language-chooser/project-language-chooser.component.spec.ts
hirokiterashima: Good suggestion. Made the code easier to understand.

src/app/common/project-language-chooser/project-language-chooser.component.spec.ts
hirokiterashima: Good suggestion. Made the code easier to understand.

PR #1551, opened 2023-12-06
refactor(NodeAuthoring): Extract CopyComponentButtonComponent

hirokiterashima:
## Changes
- Extract CopyComponentButtonComponent from NodeAuthoring component and template. This reduces NodeAuthoringComponent's responsibility.
   - Changes are saved by CopyComponentButtonComponent and the new component is sent back to NodeAuthoring via newComponentEvent output so that it can be expanded and highlighted in the view

## Test (in Node Authoring)
- Copy component button (button on the component block, not the contextual copy button that appears when you check a component) works as before. It should copy the component, place it below the original, expand and highlight it


geoffreykwan (approved):
Looks good.

PR #1552, opened 2023-12-07
refactor(TeacherProjectService): Extract getUniqueAuthors to Array.ts

hirokiterashima:
## Notes
- Looks like CC's analysis is hanging atm. But it's probably safe to proceed with review.
 
## Changes
- Move ```TeacherProjectService.getUniqueAuthors()``` to ```Array.reduceByUniqueId()```. This function didn't have to be in ```TeacherProjectService```, because it didn't touch any of the class's other fields/functions.
- Clean up accompanying test. We only care about unique id's in the resulting array, so I removed checks for other fields in the object like firstname and lastname.
- Add private modifier to ```getAuthors()``` and ```broadcastSavingProject()```

## Test
- Upon saving changes in the AT, the project's ```medata.authors``` field only contains unique authors

### DIFF
--- a/src/app/services/teacherProjectService.spec.ts
+++ b/src/app/services/teacherProjectService.spec.ts
@@ -65,7 +65,6 @@ describe('TeacherProjectService', () => {
   shouldBeAbleToInsertAGroupNodeInsideAGroupNode();
   shouldNotBeAbleToInsertAStepNodeInsideAStepNode();
   getComponentPosition();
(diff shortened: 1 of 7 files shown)


geoffreykwan (approved):
Looks good.

The tests can be written in the given-when-then form and have the 'should' word removed from the it blocks.

hirokiterashima:
Thanks. Let's discuss and improve test readability in the future.

PR #1555, opened 2023-12-08
fix(Discussion): Limit student post length

### LINKED ISSUE #1554: fix(Discussion): Limit student post length
If a student posts a really long response (like 1 million characters), the Discussion component will fail to load and just hang. This is because for the Discussion component, we run the student posts through a complex regex matcher to search and replace link text to turn them into clickable anchor links. Other component types do not have this problem because we do not run the student response from them through a regex. We should limit the number of characters a student can post to the Discussion.

geoffreykwan:
## Changes
- Limit the Discussion student post length to 10000 characters
- Set overflow wrap to break word. Previously if the student posted a long string of characters without a space, the post would only display on one line and run off to the right and off the page. This means you could only see about 200 characters of the text before it ran off the page. Now the text should wrap even if it does not have any spaces.

## Test
- As a student, go to a Discussion component and try to post a string that contains a lot of characters (like 1 million) and does not have any spaces. If you do this, the text should automatically be truncated to 10000 characters and you should be able to submit it and everything should work in regards to viewing the post and reloading the post when you leave and come back to the step.

Closes #1554

### DIFF
--- a/src/assets/wise5/components/discussion/class-response/class-response.component.html
+++ b/src/assets/wise5/components/discussion/class-response/class-response.component.html
@@ -169,6 +169,7 @@
       aria-label="Add Comment"
       i18n-aria-label
       cdkTextareaAutosize
+      maxlength="10000"
     >
     </textarea>
   </div>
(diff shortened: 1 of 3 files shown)


breity (approved):
Looks good. 👍

PR #1556, opened 2023-12-11
refactor(Export): Label component export extends new abstract component

geoffreykwan:
## Changes
- Refactored LabelComponentDataExportStrategy.ts to extend the new abstract component data export class
- Deleted the old AbstractComponentDataExportStrategy.ts
- Renamed ComponentDataExportStrategy.ts to AbstractComponentDataExportStrategy.ts

## Test
- Make sure Label export all and latest work
- Make sure Open Response export works
- Make sure Discussion export works
- Make sure Peer Chat export works


hirokiterashima (approved):
LGTM.

Future refactoring considerations:
1. ```generateComponentHeaderRow()```'s second parameter, ```columnNameToNumber``` doesn't seem to be used- can be removed?
2. It would be nice if ```AbstractComponentDataExportStrategy```'s constructor took in a ```Component``` instead of a ```nodeId: string``` and ```component: any``` because ```Component``` contains both.

PR #1557, opened 2023-12-11
refactor(NodeAuthoring): highlightAndExpandComponents()

hirokiterashima:
## Changes
- Clean up ```highlightNewComponentsAndThenShowComponentAuthoring()``` by choosing a shorter name, and removing unused ```expandComponents``` second parameter (which was always ```true```)
- Remove unused ```nodePosition``` variable
- Move duplicate ```scrollToTopOfPage()``` to dom.ts

## Test
- In node authoring, adding new component(s) by copying/importing/adding new component highlights and expands the new component(s) as before

## Note
- ```scrollToTopOfPage()``` is used in ProjectListComponent, but doesn't seem to be working in develop branch atm. The newly-copied unit should appear, but right now you don't see any changes reflected on the page. I'll write up a separate issue to fix this issue

### DIFF
--- a/src/assets/wise5/authoringTool/node/node-authoring/node-authoring.component.html
+++ b/src/assets/wise5/authoringTool/node/node-authoring/node-authoring.component.html
@@ -47,7 +47,7 @@
     <h5 i18n>Components</h5>
     <add-component-button
       [node]="node"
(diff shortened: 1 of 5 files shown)


geoffreykwan (approved):
Looks good.

PR #1560, opened 2023-12-12
feat(Classroom Monitor): Add save indicator for comments and scores

### LINKED ISSUE #218: Teacher comment save indicator
### Description

Show an indication that the comment has been saved in the grading tool.

geoffreykwan:
@breity Can you fix the styling when the save message is displayed and the browser window is really narrow. Right now when that happens you can't see the Period drop down.

## Changes
- Created SaveIndicatorComponent
- Added save indicator to the Classroom Monitor tool bar at the top
- Show save message when comment is saved
- Show save message when score is saved
- Reduced comment debounce time from 5 seconds to 1 second to be more responsive and prevent potential comment loss

## Test
- Save a comment and make sure the save message appears
- Save a score and make sure the save message appears

Closes #218

### DIFF
--- a/src/assets/wise5/common/save-indicator/save-indicator.component.html
+++ b/src/assets/wise5/common/save-indicator/save-indicator.component.html
@@ -1,7 +1,6 @@
 <div fxLayout="row wrap" fxLayoutAlign="start center">
   <mat-spinner
     *ngIf="globalMessage.isProgressIndicatorVisible"
-    class="spinner"
     [mode]="'indeterminate'"
     [diameter]="24"
   >
--- a/src/assets/wise5/common/save-indicator/save-indicator.component.html
+++ b/src/assets/wise5/common/save-indicator/save-indicator.component.html
@@ -1,0 +1,12 @@
+<div fxLayout="row wrap" fxLayoutAlign="start center">
+  <mat-spinner
+    *ngIf="globalMessage.isProgressIndicatorVisible"
+    class="spinner"
+    [mode]="'indeterminate'"
+    [diameter]="24"
+  >
+  </mat-spinner>
+  <span *ngIf="globalMessage.text" class="component__actions__info mat-small global-message">
+    {{ globalMessage.text }} {{ globalMessage.time | date: 'medium' }}
+  </span>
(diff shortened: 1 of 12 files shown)


src/assets/wise5/services/notificationService.ts
```diff
@@ -474,4 +474,34 @@ export class NotificationService {
...
+  }
+
+  /*
+   * @param message The message to show.
+   * @param timeout The amount of time to wait in milliseconds before changing the message from
+   * 'Saving...' to 'Saved' assuming showSavingMessage() was previously called. We do this to make
+   * sure the 'Saving...' message is visible long enough for the user to notice otherwise it might
+   * switch too fast from 'Saving...' to 'Saved'. If we don't perform the wait, it will always say
+   * 'Saved' and the author might wonder whether anything was ever saved.
+   */
```
breity: I think we can remove the rationale in the explanation for the `timeout` param here, since it references a specific use case not related to the component itself. I'm actually wondering if we need the `@param` documentations for this function at all, as the parameter names seem self-explanatory to me.

src/assets/wise5/common/save-indicator/save-indicator.component.html
```diff
@@ -0,0 +1,12 @@
+<div fxLayout="row wrap" fxLayoutAlign="start center">
+  <mat-spinner
+    *ngIf="globalMessage.isProgressIndicatorVisible"
+    class="spinner"
+    [mode]="'indeterminate'"
+    [diameter]="24"
```
breity: I don't think the 'spinner' class is ever used, is it? If not, can remove from here.

breity (approved):
Functionality looks good. 👍 

I made a few minor updates and a couple comments that could be addressed before merging.

As discussed, we'll move forward with this PR, then refactor to use the save indicator component in the authoring tool. When that happens, you can move the styles related to the save indicator from `authoring-tool-bar.component.scss` to the save indicator component styles. We'll then remove the use of the global save indicator in the Classroom Monitor and switch to showing the timestamp for the latest score/comment annotations below the annotation fields.

PR #1561, opened 2023-12-13
refactor(TeacherProjectService): Create updateParentGroupStartId()

geoffreykwan:
## Changes
- Moved some code from removeNodeIdFromTransitions() to a new function named updateParentGroupStartId()

## Test
- Delete a step that is the first step in a group and make sure it properly updates the start id of the group


hirokiterashima (approved):
LGTM

PR #1563, opened 2023-12-14
fix(Authoring Tool): Project does not load if there are no lessons

### LINKED ISSUE #1562: bug(Authoring Tool): Project does not load if there are no lessons
Delete all the lessons in a project and the Authoring Tool will crash and the project will no longer load again.

geoffreykwan:
## Changes
Added null check for node.transitionLogic because group0 may not have transitionLogic.

## Test
2. Delete all the lessons in the project
3. Make sure the Authoring Tool does not freeze and that you can reload the project in the Authoring Tool

Closes #1562


hirokiterashima (approved):
LGTM

PR #1565, opened 2023-12-15
fix(Authoring Tool): Component preview save and submit buttons

geoffreykwan:
## Changes
- Fixed save and submit buttons in the component preview popup
- Fixed problem with choices not being selected when you click on them when previewing Multiple Choice

## Test
- Make sure the save and submit buttons work in the component preview popup for all component types. The buttons should now grey out after you click them.
- Make sure previewing Multiple Choice with multiple answer works properly

Closes #1564

PR #1566, opened 2023-12-15
feat(Component Info): Allow multiple examples per component type

### LINKED ISSUE #1519: Show component info: Multiple examples per component
### Description

In the show component info dialog, add controls (tabs? drop-down?) to let the user see multiple examples of a component

geoffreykwan:
## Changes
- Allow multiple examples per component type in the component info popup
- If a component type has multiple examples, it will display tabs
- If a component type only has one example, it will not display tabs
- Multiple Choice now has two examples but all others still only have one
- Updated all component type info classes to contain an array of preview examples even if they only have one example

## Test
- Make sure Multiple Choice shows two examples
- Make sure all other component types only show one example

Closes #1519

### DIFF
--- a/src/assets/wise5/authoringTool/components/component-info-dialog/component-info-dialog.component.html
+++ b/src/assets/wise5/authoringTool/components/component-info-dialog/component-info-dialog.component.html
@@ -11,7 +11,7 @@
   <p class="description">{{ description }}</p>
   <mat-divider></mat-divider>
   <h3 i18n>Example:</h3>
-  <div class="example app-background">
+  <mat-card appearance="outlined">
     <preview-component
       *ngIf="previewComponents.length === 1"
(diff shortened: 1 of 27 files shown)


breity (changes requested):
Looks good. 👍 

I updated styles some to better distinguish between description and example content.

One minor issue: The multiple choice example that allows multiple selection doesn't seem to preview properly. Clicking on any one of the options doesn't toggle the checkbox. If you switch to another component type and then come back to the multiple choice component, the multiple choice preview is broken. Switching to another component and going back a second time renders the multiple choice preview again and will show the multiple selected options.

geoffreykwan:
The problems with Multiple Choice preview should be fixed in #1565.

breity (approved):

PR #1567, opened 2023-12-18
refactor(Archive Units): use string union keys

hirokiterashima:
## Changes
- Use SelectAllItemsStatus string literal union keys ('none' | 'some' | 'all') to indicate the current status of the
SelectAllItemsCheckboxComponent. The status makes it easy to see the current state of the checkbox, compared to using two variables (selectedAllItems and selectedSomeItems). Since there can only be one state (none, some, all) for the checkbox, I think this change would make it simpler and easier to understand.

- Move status calculation from PersonalLibraryComponent to the SelectAllItemsCheckboxComponent, where it's actually being used. This requires passing in numSelectedItems and numAllItems. This reduces PersonalLibraryComponent's responsibility of calculating and remembering the status values.

- Call PersonalLibraryComponent.(un)selectAllProjects() directly as a result of the output from the SelectAllItemsCheckboxComponent. This makes PersonalLibraryComponent's API a bit more clear for outside classes: call unselectAllProjects() to unselect all projects, and selectAllProjects() to select all projects, making the body of SelectAllItemsCheckboxComponent.select() easier to follow. Also remove the need to pass in a boolean flag.

### DIFF
--- a/src/app/modules/library/select-all-items-checkbox/select-all-items-checkbox.component.html
+++ b/src/app/modules/library/select-all-items-checkbox/select-all-items-checkbox.component.html
@@ -1,7 +1,7 @@
 <mat-checkbox
   color="primary"
-  [checked]="status == 'all'"
-  [indeterminate]="status == 'some'"
+  [checked]="status === 'all'"
+  [indeterminate]="status === 'some'"
   (click)="select()"
   [matTooltip]="'Select all ' + label"
   matTooltipPosition="above"
--- a/src/app/modules/library/select-all-items-checkbox/select-all-items-checkbox.component.html
+++ b/src/app/modules/library/select-all-items-checkbox/select-all-items-checkbox.component.html
@@ -1,7 +1,7 @@
 <mat-checkbox
   color="primary"
-  [checked]="selectedAllItems"
-  [indeterminate]="selectedSomeItems"
+  [checked]="status == 'all'"
+  [indeterminate]="status == 'some'"
   (click)="select()"
   [matTooltip]="'Select all ' + label"
   matTooltipPosition="above"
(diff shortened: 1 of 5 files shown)


src/app/modules/library/select-all-items-checkbox/select-all-items-checkbox.component.html
```diff
@@ -1,7 +1,7 @@
 <mat-checkbox
   color="primary"
-  [checked]="selectedAllItems"
-  [indeterminate]="selectedSomeItems"
+  [checked]="status == 'all'"
+  [indeterminate]="status == 'some'"
```
geoffreykwan: These equality checks can be changed to ===

src/app/modules/library/select-all-items-checkbox/select-all-items-checkbox.component.ts
```diff
@@ -1,18 +1,36 @@
...
 })
 export class SelectAllItemsCheckboxComponent {
   @Input() label: string = $localize`items`;
-  @Input() selectedAllItems: boolean = false;
-  @Input() selectedSomeItems: boolean = false;
-  @Output() selectAllItemsEvent: EventEmitter<boolean> = new EventEmitter<boolean>();
+  @Input() numAllItems: number;
+  @Input() numSelectedItems: number;
+  @Output() selectAllItemsEvent: EventEmitter<void> = new EventEmitter<void>();
+  @Output() selectNoItemsEvent: EventEmitter<void> = new EventEmitter<void>();
```
geoffreykwan: Should selectNoItemsEvent be named unselectAllItemsEvent to be consistent and a little more accurate?

hirokiterashima: As we discussed, I renamed these to "allSelectedEvent" and "noneSelectedEvent" to match class's status values.

src/app/modules/library/select-all-items-checkbox/select-all-items-checkbox.component.html
hirokiterashima: Good catch. Made the change.

hirokiterashima:
Made changes as we discussed today. PTAL.

geoffreykwan (approved):
Looks good.

PR #1569, opened 2023-12-19
fix(Table Authoring): Only allow positive number of rows and columns

### LINKED ISSUE #724: Table component authoring code and ui improvements
- Remove firstRow null checks from table-authoring.component.ts. First row should never be null.
- Do not allow case where author can create 0 rows and 0 columns. It should be at least 1 row, 1 column.

geoffreykwan:
## Changes
- Restricted rows and columns to only be positive integers
- Disabled row delete button if there is only one row
- Disabled column delete button if there is only one column
- Removed firstRow variable null checks

## Test
In the Table authoring, make sure the rows and columns inputs only allows positive integers and does not allow
- Negative sign
- Decimal
- The letter e
- Anything that is not a number 

Closes #724


hirokiterashima (commented):
Have you looked into performing validation with Angular's Validator? https://stackoverflow.com/a/63169367/4970939

Is it not applicable in this scenario?

geoffreykwan:
I thought validators only prevented the user from submitting the form and don't actually restrict what can actually be entered into the input.

hirokiterashima:
It looks like validators will show you an error if validation fails, but will not prevent the model from updating with the invalid value.


> If the user tries to change the value to 0 or empty, it will just change it to 1 behind the scenes

I wonder if it'd be better if we don't automatically change the model value to 1 (and update the table in the view) when the user tries to set an invalid value, and just show an error indication on the page instead?

In this example below, I set the cols and rows to 0 (invalid value), but the view updates and the table shows 1 column, 1 row, so there's a disconnect here. Wouldn't it be better if the page told me that 0 is not allowed (show error message below the input box?), and keep the table unchanged? I think that would be more in line with what I would expect to happen when I enter an invalid value.
Screenshot 2023-12-21 at 9 20 53 AM

geoffreykwan:
Inputs no longer automatically change the value to 1. Now it shows an error message if the value is invalid.

hirokiterashima (approved):
Looks good. Can you add appropriate private/protected modifiers to dimensionsForm, numColumnsFormControl and numRowsFormControl variables before merge?

PR #1570, opened 2023-12-21
refactor(Authoring Tool): Use save indicator for save messages

geoffreykwan:
## Changes
- Authoring Tool now uses save indicator to show save message

## Test
Make sure the save message works in
- Authoring Tool
- Classroom Monitor (score and comments)


hirokiterashima (approved):
Works well. Looks like SaveIndicatorComponent can be a standalone component, and remove the need for SaveIndicatorModule?

geoffreykwan:
I've converted the SaveIndicatorComponent to be standalone.

PR #1578, opened 2024-01-19
feat(Multiple Languages): TranslatableInputComponent

hirokiterashima:
## Notes

## Changes
- Create TranslatableInputComponent to show translated text for the current language in the input box and allow author to make changes. (Saving those changes is not yet implement)
- Convert some inputs in MC and Match authoring to use TranslatableInputComponent
   - MC: choice text and feedback
   - Match: source bucket and choices
## Test AT (with unit available in multiple languages)
- Switching to a supported language using the language chooser (top-right corner of AT) shows...
   - the correct text in the specified language, if it has been translated before
   - empty box, if it has not been translated before
   - for the MC and Match inputs that have been modified to use the TranslatableInputComponent, check that the placeholder and label appear correctly.
- Making changes
   - in the default language => should save changes as before
   - in a supported language => does nothing at the moment (saving will be implemented in the future)

## Test AT (with single-language unit)
- Authoring works as before

### DIFF
--- a/src/assets/wise5/authoringTool/components/translatable-input/translatable-input.component.spec.ts
+++ b/src/assets/wise5/authoringTool/components/translatable-input/translatable-input.component.spec.ts
@@ -1,0 +1,31 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+
+import { TranslatableInputComponent } from './translatable-input.component';
+import { StudentTeacherCommonServicesModule } from '../../../../../app/student-teacher-common-services.module';
+import { TeacherProjectService } from '../../../services/teacherProjectService';
+import { HttpClientTestingModule } from '@angular/common/http/testing';
+
+describe('TranslatableInputComponent', () => {
+  let component: TranslatableInputComponent;
+  let fixture: ComponentFixture<TranslatableInputComponent>;
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      imports: [
+        HttpClientTestingModule,
+        StudentTeacherCommonServicesModule,
+        TranslatableInputComponent
+      ],
+      providers: [TeacherProjectService]
+    });
+    spyOn(TestBed.inject(TeacherProjectService), 'isDefaultLocale').and.returnValue(true);
+    fixture = TestBed.createComponent(TranslatableInputComponent);
+    component = fixture.componentInstance;
+    component.content = {};
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+});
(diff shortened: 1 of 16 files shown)


src/assets/wise5/authoringTool/components/translatable-input/translatable-input.component.spec.ts
```diff
@@ -0,0 +1,31 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+
```
geoffreykwan: This empty line can be deleted.

src/assets/wise5/authoringTool/components/translatable-input/translatable-input.component.ts
```diff
@@ -0,0 +1,45 @@
...
+  standalone: true,
+  selector: 'translatable-input',
+  imports: [CommonModule, FormsModule, MatInputModule],
+  styleUrls: ['./translatable-input.component.scss'],
+  templateUrl: './translatable-input.component.html'
+})
+export class TranslatableInputComponent {
+  @Input() content: object;
+  @Input() key: string;
+  @ContentChild(MatInput) input: MatInput;
```
geoffreykwan: Maybe the input variable can be renamed to something like contentInput to be a little more clear particularly when reading the template file.

src/assets/wise5/authoringTool/components/translatable-input/translatable-input.component.ts
```diff
@@ -0,0 +1,45 @@
...
+  @ContentChild(MatLabel, { read: ElementRef }) labelRef: ElementRef;
```
geoffreykwan: Maybe the labelRef variable can be renamed to something like contentLabelRef to be a little more clear particularly when reading the template file.

src/assets/wise5/authoringTool/components/translatable-input/translatable-input.component.ts
```diff
@@ -0,0 +1,45 @@
...
+    this.showTranslationInput = computed(() => !this.projectService.isDefaultLocale());
+  }
+
+  ngOnInit(): void {
+    const i18nId = this.content[`${this.key}.i18n`]?.id;
+    this.translatedText = computed(() => {
+      if (this.showTranslationInput()) {
+        if (this.translateProjectService.currentTranslations()[i18nId]?.value) {
+          return this.translateProjectService.currentTranslations()[i18nId]?.value;
+        }
```
geoffreykwan: ```this.translateProjectService.currentTranslations()[i18nId]?.value``` can be retrieved just once and then used in the if and return.

geoffreykwan (approved):
Works well.

PR #1581, opened 2024-01-22
fix(Grade by Team): Changing workgroup using select updates status/score

### LINKED ISSUE #1580: bug(Grade by Team): Changing workgroup using select drop down does not update status and score
### Description

2. Go to the Grade by Team view
3. Click on a team
4. Use the team select drop down to switch to a different team
5. It will show the Status as "No Work" and Score as "-/0" for all the steps even if the workgroup worked on the step and has scores

geoffreykwan:
## Changes
- Changing workgroup using select drop down now also changes the route. Previously only the next and previous buttons would change the route but now using the select drop down also changes the route.

## Test
2. Go to the Grade by Team view
3. Click on a team
4. Use the team select drop down to switch to a different team
5. The Status and Score should update. Previously it would show the Status as "No Work" and Score as "-/0" for all the steps even if the workgroup worked on the step and had scores

Closes #1580


breity (approved):
Looks good. 👍

PR #1583, opened 2024-01-23
feat(Multiple Languages): Show default language text hint below the translatable input box

hirokiterashima:
## Notes

## Changes
- Show default language text hint below the translatable input box. This only appears when the current language is not the default language
- Cleaned up code to compute translatedText using ternary statement

## Test AT (with unit available in multiple languages)
- Switching to a supported language using the language chooser (top-right corner of AT) shows the default language text hint below the translatable input box
- Switching to the default language using the language chooser hides the default language text hint


geoffreykwan (approved):
Looks good.

The variable name defaultLanguageTextHint might be little more clear if it was just called defaultLanguageText. That way we have translatedText and defaultLanguageText.

PR #1585, opened 2024-01-24
feat(Multiple Languages): Convert EditComponentPrompt input to TranslatableInput

hirokiterashima:
## Notes
 
## Changes
- Pass in ComponentContent object to EditComponentPrompt instead of just the prompt string. This lets us convert the input to TranslatableInput.
- Convert EditComponentPrompt's input to TranslatableInput

## Test in AT
These should apply to all components, but you can just test with handful of components to save time (since they all use the same code in EditComponentPrompt):
- In default language, editing prompt should save as before
- Switching to a supported language using the language chooser (top-right corner of AT)...
   - shows the correct prompt text in the specified language, if it has been translated before
   - shows empty box, if it has not been translated before
   - shows the placeholder and label appear correctly
   - making changes to the prompt does nothing at the moment (saving will be implemented in the future)

### DIFF
--- a/src/app/authoring-tool/edit-component-prompt/edit-component-prompt.component.html
+++ b/src/app/authoring-tool/edit-component-prompt/edit-component-prompt.component.html
@@ -2,7 +2,7 @@
   <mat-label i18n>Prompt</mat-label>
   <textarea
     matInput
-    [(ngModel)]="prompt"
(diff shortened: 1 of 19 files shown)


geoffreykwan (approved):
Looks good.

PR #1586, opened 2024-01-24
refactor(Archive Run): Reference run.project.archived field

geoffreykwan:
## Changes
- Archive run now uses the run.project.archived field instead of run.archived. This is so we have a common field to reference whether we are archiving a run in the class schedule or archiving a unit in the library.

## Test
- Make sure archiving/restoring runs using the global archive/restore button works
- Make sure archiving/restoring a run from the run menu works
- Make sure archive/restore undo works

### DIFF
--- a/src/app/teacher/run-menu/run-menu.component.html
+++ b/src/app/teacher/run-menu/run-menu.component.html
@@ -29,11 +29,11 @@
       <mat-icon>report_problem</mat-icon>
       <span i18n>Report Problem</span>
     </a>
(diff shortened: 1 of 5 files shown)


hirokiterashima (approved):
LGTM.

In TeacherRunListComponent.performFilter(), you can remove the ```return``` statement and the surrounding brackets to simplify the code block.

```
 this.filteredRuns = this.filteredRuns.filter(
      (run: TeacherRun) =>
        (!this.showArchived && !run.project.archived) || (this.showArchived && run.project.archived)
    );
```

PR #1588, opened 2024-01-25
feat(Multiple Languages): Show current language in TranslatableInput's label

hirokiterashima:
## Changes
- Display language word directly in the TranslatableInput label  “[label] (Language)”, e.g. "Prompt (Japanese)", "Choice Text (Spanish)". This helps the translator know to type the translation for the currently-selected language.

## Test in AT with supported languages
- Switch between languages
   - When you choose a supported language, the TranslatableInput shows the selected language in the label.
   - When you choose the default language, TranslatableInput does not show the language.

image

### DIFF
--- a/src/assets/wise5/authoringTool/components/translatable-input/translatable-input.component.html
+++ b/src/assets/wise5/authoringTool/components/translatable-input/translatable-input.component.html
@@ -1,6 +1,6 @@
 <ng-content *ngIf="!showTranslationInput()"></ng-content>
 <mat-form-field *ngIf="showTranslationInput()">
-  <mat-label i18n
+  <mat-label
     >{{ defaultLanguageLabelRef.nativeElement.innerHTML }} ({{
       currentLanguage().language
     }})</mat-label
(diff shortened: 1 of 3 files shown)


src/assets/wise5/authoringTool/components/translatable-input/translatable-input.component.html
```diff
@@ -1,6 +1,10 @@
 <ng-content *ngIf="!showTranslationInput()"></ng-content>
 <mat-form-field *ngIf="showTranslationInput()">
-  <mat-label i18n>{{ defaultLanguageLabelRef.nativeElement.innerHTML }}</mat-label>
+  <mat-label i18n
```
breity: I think you can remove the i18n attribute here because I don't think this mat-label needs to be translatable.

breity (approved):
Looks good. 👍 

Ready to merge after removing the i18n attribute I mentioned.

src/assets/wise5/authoringTool/components/translatable-input/translatable-input.component.html
hirokiterashima: Thanks for the catch.

PR #1589, opened 2024-01-25
feat(ProjectLanguageChooser): Make button tooltip into an input

hirokiterashima:
## Changes
- The tooltip when hovering over the ProjectLanguageChooser can be specified as an input to the component
- In the AT, the tooltip now says "Choose Language to Translate". This is a more helpful tooltip for translators.

## Test (with a unit that has multiple supported languages)
- In the AT, the tooltip says "Choose Language to Translate"
- In the VLE, the tooltip says "Select Language"


breity (approved):
Looks good. 👍

PR #1590, opened 2024-01-25
feat(TranslatableInput): Move default language text to hint

breity:
## Changes
Move the default language text display to a hint for the form field.

## Test
Make sure default language text appears directly below input (in a `mat-hint` element) field when translating to a supported language.


hirokiterashima (approved):
LGTM

PR #1592, opened 2024-01-26
refactor(ArchiveProjectsButton): Input selected projects only

hirokiterashima:
## Changes
- Only send selected projects as input to ArchiveProjectsButtonComponent. This component did not need all the projects to do its work.
- Don't call extractProjects() in ngOnInit() because it's already called in ngOnChanges(). Move extractProjects()'s body to ngOnChanges().

## Test
- Archiving and restoring units work as before in Project Library and Class Schedule pages


geoffreykwan (approved):
Looks good.

PR #1597, opened 2024-01-29
fix(Project Library): Search unselects projects that do not match search

### LINKED ISSUE #1593: bug(Project Library): Search should apply to selected projects
### Description

Using the search textbox to filter projects should be applied to the selected projects also.

### Steps to Reproduce

1. In Unit Library, click on "Select All". Say 12 units are selected, and only 1 unit is "Photosynthesis"
2. In the search textbox, type in "Photosynthesis"

**Expected behavior:** 
At this point, only 1 unit should be selected

**Actual behavior:**
12 units are still selected

geoffreykwan:
## Changes
- Selected projects should be unselected if a search is performed and the project does not match the search criteria

On a related note, I think we might be able to get rid of the projectIdToIsSelected variable due to recent changes where we use the project.selected field to store whether a project is selected. I can look into this in another issue.

## Test
1. Go to the personal unit library
2. Select all the units using the select all checkbox
3. Go to the search input and type in a string that only matches some of the units
4. The number of selected units should update accordingly

Closes #1593

### DIFF
--- a/src/app/modules/library/personal-library/personal-library.component.spec.ts
+++ b/src/app/modules/library/personal-library/personal-library.component.spec.ts
@@ -81,15 +81,36 @@ describe('PersonalLibraryComponent', () => {
   allProjectsSelected_clickSelectAllProjects_unselectsAllProjects();
   projectsAreSelected_goToArchivedView_projectsAreUnselected();
   projectsAreSelected_goToNextPage_projectsAreUnselected();
+  projectsAreSelected_performSearch_projectsNotMatchingSearchAreUnselected();
 });
 
 function setUpFiveProjects() {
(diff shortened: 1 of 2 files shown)


hirokiterashima:
The new functionality works as you described.

When I do select all -> search filter some units -> clear search filter, what should be the expected resulting view? 

Should it be:
1. Only select some units (current behavior)
2. Revert to selecting all units
3. Clear all selections

The current behavior feels a bit off to me atm. I think I was expecting behavior 2 or 3?

hirokiterashima (approved):
We discussed this together and agreed that behavior 1 (current behavior) is good for now. In the future, we'll consider aligning the behavior similar to how gmail does it, where select and filter are separate actions

PR #1602, opened 2024-01-30
feat(Multiple languages): Save changes to existing translations

hirokiterashima:
## Changes
- Save changes to an existing translatable field (one that already has a value in translations.[locale].json)

## Test Prep
- Test with https://github.com/WISE-Community/WISE-API/pull/253

## Test in AT with project that support additional languages
- Change the text of an existing translation (one that already has a value in translations.[locale].json).
   - There should be a debounce before the saving sequence starts 
   - The saving...saved indicator should appear in the top-right corner of the AT
   - New translation text should be saved. You can verify this by reloading the AT or previewing the unit.

### DIFF
--- a/src/assets/wise5/authoringTool/components/translatable-input/translatable-input.component.ts
+++ b/src/assets/wise5/authoringTool/components/translatable-input/translatable-input.component.ts
@@ -23,12 +23,12 @@ export class TranslatableInputComponent {
   @ContentChild(MatInput) defaultLanguageInput: MatInput;
   @ContentChild(MatLabel, { read: ElementRef }) defaultLanguageLabelRef: ElementRef;
   protected defaultLanguageText: Signal<string>;
-  @Input() key: string;
   private i18nId: string;
+  @Input() key: string;
   protected showTranslationInput: Signal<boolean>;
-  protected translatedText: Signal<string>;
-  protected translatedTextChanged: Subject<string> = new Subject<string>();
-  protected translatedTextChangedSubscription: Subscription;
+  protected translationText: Signal<string>;
+  protected translationTextChanged: Subject<string> = new Subject<string>();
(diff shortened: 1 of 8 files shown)


src/assets/wise5/authoringTool/components/translatable-input/translatable-input.component.ts
```diff
@@ -21,10 +24,14 @@ export class TranslatableInputComponent {
   @ContentChild(MatLabel, { read: ElementRef }) defaultLanguageLabelRef: ElementRef;
   protected defaultLanguageText: Signal<string>;
   @Input() key: string;
+  private i18nId: string;
```
geoffreykwan, line 26: The variable ```i18nId``` should be above the ```key``` variable to be alphabetical.

src/assets/wise5/authoringTool/components/translatable-input/translatable-input.component.ts
```diff
@@ -21,10 +24,14 @@ export class TranslatableInputComponent {
...
   protected showTranslationInput: Signal<boolean>;
   protected translatedText: Signal<string>;
+  protected translatedTextChanged: Subject<string> = new Subject<string>();
+  protected translatedTextChangedSubscription: Subscription;
```
geoffreykwan: I noticed that some variable names use the word translation and others use translated. It might be good to be consistent. Perhaps just use the word translation and change the variables names to be translationText, translationTextChanged, translationTextChangedSubscription.

geoffreykwan (approved):
Works well.

Should we be sending all the translations every time one translation is changed?

hirokiterashima:
> Should we be sending all the translations every time one translation is changed?

We agreed this would be good address in the future. Keep as is for now.

PR #1605, opened 2024-01-30
fix(Authoring): Move two lessons into unused section bug

### LINKED ISSUE #1604: bug(Authoring): Move two lessons into unused section throws error
2. Select two lessons
3. Click the "Move" button
4. Click "Insert Inside" the Unused Lessons section
5. An error will occur and the lessons will not be moved. If you refresh the unit, you will see the lessons were never moved.

geoffreykwan:
## Changes
- Fixed a bug where node id strings were being added to the inactiveGroupNodes and inactiveStepNodes arrays which actually expect node objects. This caused an error when moving two lessons into the Unused Lessons section.

## Test
2. Select two lessons
3. Click the "Move" button
4. Click "Insert Inside" the Unused Lessons section
5. The lessons should be moved to the Unused Lessons section. Previously, an error would occur and the lessons would not be moved.

Closes #1604

### DIFF
--- a/src/assets/wise5/services/teacherProjectService.ts
+++ b/src/assets/wise5/services/teacherProjectService.ts
@@ -721,10 +721,10 @@ export class TeacherProjectService extends ProjectService {
       this.insertNodeAfterInactiveNode(node, nodeIdToInsertAfter);
     }
     if (node.type === 'group') {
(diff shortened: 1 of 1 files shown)


hirokiterashima (approved):
Verified bug and fix.

PR #1606, opened 2024-02-01
refactor(Authoring): Create project authoring step component

geoffreykwan:
## Changes
- Created project authoring step component

## Test
- In the project authoring view make sure
  - Steps show up properly
  - Step numbers show up correctly
  - Steps are highlighted on mouse hover
  - Steps can be clicked to open them
  - Steps can be selected using their checkbox and then moved, copied, and deleted
  - Steps in a branch path display their branch color
  - If a step has a constraint, it should show a constraint icon on the step that can be clicked which brings you to the constraint authoring view
  - If a step is a branch point, it should show a branch icon on the step that can be clicked which brings you to the branch authoring view
  - If a step has a rubric, it should show a rubric icon (clicking on this does not bring you to the rubric authoring view, this is the current behavior on production)
  - Steps show up properly in the unused lessons section
  - Steps show up properly in the unused steps section


hirokiterashima (approved):
LGTM.

While reviewing, I had a thought about renaming css classes to be shorter- now that we have lesson and step components, do we need to say projectItemTitleDiv anymore, or can we just say titleDiv (or even just "title"?) in the respective files, because the projectItem part is already implied in the context? Similar thing with "groupHeader" => "header". We can do this in a future task.

PR #1607, opened 2024-02-01
fix(Teacher Library): Ensure checkbox appears at top of library unit card

breity:
## Changes
- Ensures that the checkboxes on teacher library units appears at the top of the unit card.
- Fixes a bug where the checkbox would appear at the bottom sometimes.
- Also added a border-radius to the checkbox to make it's appearance less harsh.

## Test
- Make sure checkboxes always appear the top of the unit card.


hirokiterashima (approved):
LGTM

geoffreykwan (approved):
Looks good.

PR #1609, opened 2024-02-01
refactor(Export): Remove Embedded component export

geoffreykwan:
## Changes
- Removed Embedded component export since it only worked for a specific model and required the component to have the tag 'devices-kwh-co2-table' or 'transportation-co2-table'
- Removed functions that were no longer used since Embedded component export was the last component type to be cleaned up

## Test
- Make sure the other exports still work


hirokiterashima (approved):
LGTM

PR #1611, opened 2024-02-02
refactor(Student): Create chat input component

geoffreykwan:
## Changes
- Created chat input component which contains a text input and a send button
- Use chat input component in AI Chat and Dialog Guidance

## Test
- In AI Chat and Dialog Guidance student views
  - Make sure the student can send chat messages
  - Make sure the student can not send another chat message until after they receive a response from the computer

### DIFF
--- a/src/assets/wise5/common/chat-input/chat-input.component.html
+++ b/src/assets/wise5/common/chat-input/chat-input.component.html
@@ -1,0 +1,21 @@
+<div fxLayout="row" fxLayoutAlign="start center" fxLayoutGap="8px">
+  <mat-form-field fxFlex class="form-field-no-label form-field-no-hint" appearance="fill">
+    <textarea
+      matInput
+      cdkTextareaAutosize
+      placeholder="Add response..."
+      i18n-placeholder
+      [(ngModel)]="response"
+      (keypress)="keyPressed($event)"
+    ></textarea>
+  </mat-form-field>
+  <button
+    mat-flat-button
+    color="primary"
+    (click)="submit()"
+    [disabled]="response.length === 0 || submitDisabled"
+    i18n
+  >
+    Send
+  </button>
+</div>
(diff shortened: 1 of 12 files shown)


hirokiterashima (commented):
The new ChatInputComponent makes the code in AIChat and DG components much simpler. 👍 

This part looks like it can be improved in ChatInputComponent.html:
```
[disabled]="response.length === 0 || submitDisabled"
```
In ChatInputComponent, we have ```submitDisabled``` field, which, by the name of the variable, I would expect to be the only place we need to look at to disable/enable the submit button. But instead, we need to evaluate a boolean expression involving response length.

Can we change the code so that this part looks like this instead?
```
[disabled]="submitDisabled"
```

hirokiterashima (approved):
Approving this PR, since it's already an improvement to the codebase. You can address my comment in this PR or in a future PR- up to you.

geoffreykwan:
The reason why it's like that is because submitDisabled is an input variable that allows the parent to disable the submit button. We use this in scenarios where the student has submitted and we are waiting for the response from CRater or Chat GPT. We do not want the student to be able to submit when we are waiting for the response.

We also want to disable the submit button if the student has not typed anything into the input. That's why we need the or expression.

I guess we could name the input variable something like submitDisabledByParent and then create a computed signal variable named submitDisabled that computes response.length === 0 || submitDisabledByParent. This would make the code more complex and be doing the same thing.

hirokiterashima:
Yah, I understood why it's like that, but the naming of the submitDisabled variable still threw me off when I was reading the component code and template. Renaming the input from ```submitDisabled``` to ```submitDisabledByParent``` would make it clearer. We can still have ```submitDisabled``` variable that gets computed when ```submitDisabledByParent``` or response length changes, and continue to use that to enable/disable the submit button.

PR #1614, opened 2024-02-05
refactor(Student): Move computer avatar code to common location

geoffreykwan:
## Changes
- Created mixin for
  - computer avatar
  - computer avatar component
  - computer avatar component content
- AI Chat and Dialog Guidance now apply the computer avatar mixin

## Test
- Make sure computer avatar works for AI Chat and Dialog Guidance


hirokiterashima (approved):
LGTM. Functionality seems to work as before.

Some minor things to consider before merge:
- Make a new computerAvatar folder inside the common folder for storing computer avatar files?
- Rename ```ComputerAvatarComponent``` to ```ComputerAvatarComponentMixin```? I noticed that you created a class called ```ComputerAvatarMixin```, so adding Mixin suffix will make it consistent?
- Generate messages.xlf?

geoffreykwan:
I'm beginning to think that we might not want to name things mixin since they're actually just classes and don't necessarily need to be used as a mixin. If we rename ComputerAvatarMixin to not have the word Mixin, what would it be?

We already have things named
- ComputerAvatar
- ComputerAvatarComponent
- ComputerAvatarComponentContent
- ComputerAvatarSettings

PR #1618, opened 2024-02-06
refactor(TranslatableInput): Use two inputs instead of content projection

hirokiterashima:
## Changes
- Pass in required values (label, placeholder) and output (defaultLanguageTextChanged) into the TranslatableInputComponent. This will let us use two inputs and switch between them based on whether we're translating or not. This will allow us to style the input more easily.
- Note that prompt authoring has not been converted yet, because it uses a textarea instead of an input

## Test
- Inputs that have been converted to TranslatableInputs (MC choice and feedback authoring, Match choice and source bucket name) work as before, including
   - showing text, label, placeholder in default and translated languages
   - saving text in default and translated languages. Note that at the moment, saving only works in translated language if translation already exists in the translation file.


breity (changes requested):
Looks good, functionality works well.

breity (approved):

PR #1619, opened 2024-02-06
refactor(Export): Move createStudentWorkExportRow function

geoffreykwan:
## Changes
- Moved createStudentWorkExportRow function from data-export.component.ts to StudentWorkDataExportStrategy.ts because it is only used in that class

## Test
Make sure export works, particularly
- One workgroup per row
- Latest Work
- All Work
- Component data Discussion


hirokiterashima (approved):
LGTM. Remember to generate messages before merge?

PR #1620, opened 2024-02-07
feat(Classroom Monitor): Sort by first and last names in Grade by Team

breity:
## Changes
- Add "First Name" and "Last Name" columns to the Grade by Team view in the Classroom Monitor.
- Display one row per student instead of one row per workgroup.
- When user doesn't have view student names permission, include a "Student" column with student id # instead of first name/last name columns.
- Rename "Grade by Team" to "Grade by Student".

## Test
- Ensure that first name and last name columns are displayed in Grade by Student view and there is one row per student.
- Ensure that sorting students by each column header works as expected.

The StudentProgress component  could use some refactoring to reduce complexity and the template could be refactored to reduce duplication. We can address this in a separate PR.

### DIFF
--- a/src/app/help/faq/teacher-faq/teacher-faq.component.html
+++ b/src/app/help/faq/teacher-faq/teacher-faq.component.html
@@ -256,7 +256,7 @@
     <li i18n>Here you will be able to look at your students' work.</li>
     <li i18n>
       On the left sidebar you can choose to look at the work by step by clicking "Grade By Step" or
-      you can choose to look at the work by team by clicking "Grade By Team".
+      you can choose to look at the work by student by clicking "Grade By Student".
     </li>
     <li i18n>Next to each student's work, you will be able to give them a score and comment.</li>
   </ul>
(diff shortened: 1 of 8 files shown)


hirokiterashima (changes requested):
Everything looks good except one new bug: sorting multiple-times (asc -> desc -> asc) by current location doesn't work if there is a student who has not worked on the unit. It throws a js error (see below). This appears to be working on production, so it's likely introduced in this PR?

```
ERROR TypeError: Cannot read properties of null (reading 'localeCompare')
    at student-progress.component.ts:232:31
    at Array.sort (<anonymous>)
    at StudentProgressComponent.sortWorkgroups (student-progress.component.ts:154:29)
    at StudentProgressComponent.setSort (student-progress.component.ts:258:10)
    at StudentProgressComponent_Template_button_click_17_listener (student-progress.component.html:99:26)
    at executeListenerWithErrorHandling (core.mjs:16795:16)
    at wrapListenerIn_markDirtyAndPreventDefault (core.mjs:16828:22)
    at HTMLButtonElement.<anonymous> (platform-browser.mjs:665:17)
    at _ZoneDelegate.invokeTask (zone.js:402:31)
    at core.mjs:10757:55
```

breity:
@hirokiterashima Ah, good catch. Sorting by completion also didn't sort properly when there was a student that hadn't started the unit. Just pushed a fix for both.

hirokiterashima (approved):
LGTM.

PR #1624, opened 2024-02-09
feat(Speech to Text): Add SpeechToTextComponent and add to Open Response

hirokiterashima:
## Changes
- Add support to connect to AWS Transcribe streaming service (mostly handled in TranscribeService.ts)
- Add support for specifying ```speechToText``` block in the unit JSON
- Add SpeechToTextComponent (language select and record button) that appears in OR components if the unit is set up to use speech to text.

## Test Prep
- Test with https://github.com/WISE-Community/WISE-API/pull/254
   - configure aws region id and identity pool id in application properties file 
- Add ```speechToText``` block to unit JSON (top-level, along with metadata, nodes, and notebook blocks):
```
...
    "speechToText": {
        "enabled": true,
        "defaultLanguage": "en-US",
        "supportedLanguages": [
            "en-US",
            "ja-JP",
            "it-IT",
            "es-US"
        ]
    },
...
```

## Test
- With the above setting in the unit, every OR component in the unit should display the select language drop-down and a record button below the textarea
- If you set enabled=false, or remove the speechToText block altogether from the unit, the controls should not appear anywhere in the unit.
- Hit the record button and try talking in different languages. The transcribed text should appear in the OR textarea, and it should mark the component as dirty (enable the save button for the component or the step)
- Make sure that transcribing works on OR components on the same step and on different steps.


src/assets/wise5/services/transcribeService.ts
```diff
@@ -0,0 +1,118 @@
...
+      MediaSampleRateHertz: this.SAMPLE_RATE,
+      AudioStream: this.getAudioStream()
+    });
+    const data = await this.transcribeClient.send(command);
+    for await (const event of data.TranscriptResultStream) {
+      for (const result of event.TranscriptEvent.Transcript.Results || []) {
+        if (result.IsPartial === false) {
+          const noOfResults = result.Alternatives[0].Items.length;
+          for (let i = 0; i < noOfResults; i++) {
+            callback(result.Alternatives[0].Items[i].Content + ' ');
```
geoffreykwan: I think the punctuation can be improved if we change this line to
```
            let text = result.Alternatives[0].Items[i].Content;
            if (![',', '.', '?', '!', ':', ';'].includes(text)) {
              text = ' ' + text;
            }
            callback(text);
```

src/assets/wise5/components/openResponse/speech-to-text/speech-to-text.component.ts
```diff
@@ -0,0 +1,81 @@
...
+    { languageCode: 'en-US', language: $localize`English` },
+    { languageCode: 'fr-FR', language: $localize`French` },
+    { languageCode: 'de-DE', language: $localize`German` },
+    { languageCode: 'it-IT', language: $localize`Italian` },
+    { languageCode: 'ja-JP', language: $localize`Japanese` },
+    { languageCode: 'ko-KR', language: $localize`Korean` },
+    { languageCode: 'pt-BR', language: $localize`Portuguese (Brazilian)` },
+    { languageCode: 'es-US', language: $localize`Spanish` }
+  ];
+  protected languages;
```
geoffreykwan: This variable can be typed.

src/assets/wise5/components/openResponse/speech-to-text/speech-to-text.component.ts
```diff
@@ -0,0 +1,81 @@
...
+
+@Component({
+  standalone: true,
+  selector: 'speech-to-text',
+  imports: [CommonModule, FormsModule, MatFormFieldModule, MatOptionModule, MatSelectModule],
+  templateUrl: './speech-to-text.component.html',
+  styleUrls: ['./speech-to-text.component.scss']
+})
+export class SpeechToTextComponent {
+  @ViewChild('inputLanguageList') inputLanguageList: MatSelect;
```
geoffreykwan: This variable name can be ordered alphabetically.

geoffreykwan (approved):
Looks good.

breity (approved):
LGTM. 👍 

I agree with @geoffreykwan's suggestion for improving the punctuation. Seems to work well.

The delay that's present when transcribing speech can result in some of the student's spoken text not being added to the text input. This occurs when the student finishes speaking and presses the button to stop recording before all the speech has been added to the input. It would be good to try to address this in a future issue: try to ensure that the service continues processing all the student speech that was recorded before the recording was stopped.

PR #1626, opened 2024-02-12
refactor(Export): Create component for step and component checkboxes

geoffreykwan:
## Changes
- Created a component to display lessons, steps, and components and their checkboxes

## Test
2. Go to the Data Export view
3. Go to the export types below
   - One Workgroup Per Row
   - Latest Work
   - All Work
   - Raw Data
4. Choose "Select Steps To Export"
5. Choose a specific lesson, step, or component
6. Generate the export and make sure the export only contains data for the items you selected


hirokiterashima (approved):
LGTM.

For future refactoring consideration: I wonder if the input to SelectStepAndComponentCheckboxesComponent should be a ```project``` instead of ```nodes``` and ```projectIdToOrder```, and get those variables inside the ngOnInit() of the component? I think this would make it easier for DataExportComponent and any other component in the future that wants to use the SelectStepAndComponentCheckboxesComponent.

PR #1627, opened 2024-02-12
feat(Authoring Tool): Create TranslatableTextareaComponent

hirokiterashima:
## Notes
- After this PR is merged, TranslatableTextareaComponent and TranslatableInputComponent will have a lot of duplicated code. Don't worry about the duplication in this review. After this, we'll do a separate refactor PR to extract common code in this feature branch (to ensure that the duplication won't make it into the develop branch). 
- Please style as you see fit

## Changes
- Create TranslatableTextareaComponent to allow translators to translate textarea fields in the authoring tool
- Modify EditComponentPromptComponent to use this new component

## Test
- Component prompt authoring work as before and can be translated into different languages, including
   - shows text, label, placeholder in default and translated languages
   - shows default language text as hint if you're authoring in translated language
   - saves text in default and translated languages. Note that at the moment, saving only works in the translated language if translation already exists in the translation file.


breity (approved):
LGTM. 👍

PR #1632, opened 2024-02-14
feat(Multiple Languages): Save new translations

hirokiterashima:
## Changes
This change allows translators to save translations for fields that have never been translated into another language before. The code adds a new i18n field to the default language content, and uses the new i18nId (30-character random string) to save the new translation text.

## Known issue
If you have two translatable languages (e.g. Default=English, Translatable Languages=[Spanish, Italian]), and you save a new translation text in Spanish and then switch to Italian, the Spanish text appears in the input, even though it should show empty. This is not an issue that this PR introduced, so you can ignore it for now. I'll look into it later.

## Test
- Enter translation texts for fields that have never been translated before.
   - This should create a new i18n field in the default language's content (with new 30-character i18n ID and modified timestamp) and save the project.json
   - This should also save the new translation in the translations.[locale].json file using the new 30-character i18n ID.
- Updating existing translations should work as before


geoffreykwan (approved):
Looks good.

PR #1634, opened 2024-02-15
feat(Speech to Text): Ensure only one component on a step is recording

hirokiterashima:
## Changes
- Introduced a ```recording``` boolean signal in TranscribeService to globally keep track of when transcription recording is happening. Use this variable to enable/disable the controls (language selection and record button) in the SpeechToTextComponent
- When one component is recording, other component(s)' STT controls are displayed but disabled.

## Test (on a step with two or more OR components with STT controls enabled)
- Click record for one component
   - It should disable language selection and record button for the other component(s).
  - Transcription should work as before
- Click stop for the component. It should re-enable language selection and record button for the other component(s).

### DIFF
--- a/src/assets/wise5/components/openResponse/speech-to-text/speech-to-text.component.ts
+++ b/src/assets/wise5/components/openResponse/speech-to-text/speech-to-text.component.ts
@@ -47,7 +47,7 @@ export class SpeechToTextComponent {
   protected languages: Language[];
   @Output() newTextEvent: EventEmitter<string> = new EventEmitter<string>();
   protected recording: Signal<boolean> = this.transcribeService.recording;
-  protected requestedRecording = false;
+  protected recordingRequester = false;
   protected selectedLanguage: Language = { languageCode: null, language: null };
 
   constructor(
@@ -83,7 +83,7 @@ export class SpeechToTextComponent {
 
   private async startRecording(): Promise<void> {
     try {
-      this.requestedRecording = true;
+      this.recordingRequester = true;
       await this.transcribeService.startRecording(
         this.selectedLanguage.languageCode,
         this.processTranscriptionText.bind(this)
@@ -96,7 +96,7 @@ export class SpeechToTextComponent {
 
   private stopRecording(): void {
     this.transcribeService.stopRecording();
-    this.requestedRecording = false;
+    this.recordingRequester = false;
   }
 
   private processTranscriptionText(text: string): void {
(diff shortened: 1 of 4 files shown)


src/assets/wise5/components/openResponse/speech-to-text/speech-to-text.component.ts
```diff
@@ -44,10 +44,10 @@ export class SpeechToTextComponent {
     { languageCode: 'pt-BR', language: $localize`Portuguese (Brazilian)` },
     { languageCode: 'es-US', language: $localize`Spanish` }
   ];
-  @ViewChild('languageSelectButton') languageSelectButton: MatButton;
   protected languages: Language[];
   @Output() newTextEvent: EventEmitter<string> = new EventEmitter<string>();
-  protected recording: boolean;
+  protected recording: Signal<boolean> = this.transcribeService.recording;
+  protected requestedRecording = false;
```
geoffreykwan: I'm wondering if the requestedRecording variable name can be renamed to be more clear? It took me a while to figure out that this variable signifies that this component is the one that is recording as opposed to any recording currently being requested. Maybe if it's named recordingRequester or currentRecorder or something it would be better.

geoffreykwan (approved):
Functionality works.

This section can be cleaned up.

https://github.com/WISE-Community/WISE-Client/blob/35524e457e30a021e31198a09b139d5eb5b238d0/src/assets/wise5/components/openResponse/speech-to-text/speech-to-text.component.html#L24-L47

Here's a possible way to improve this code. Also moving the color from the mat-icon to the button will make the button actually look disabled when it's disabled. Currently, the button does not grey out when disabled.

```
  <button
    *ngIf="!(recording() && requestedRecording)"
    [disabled]="recording()"
    mat-icon-button
    color="primary"
    id="recordButton"
    (click)="toggleRecording()"
    matTooltip="Record voice"
    i18n-mat-tooltip
  >
    <mat-icon>mic</mat-icon>
  </button>
  <ng-container *ngIf="recording() && requestedRecording">
    <button
      mat-icon-button
      color="warn"
      id="recordButton"
      (click)="toggleRecording()"
      matTooltip="Stop recording"
      i18n-mat-tooltip
    >
      <mat-icon>stop_circle</mat-icon>
    </button>
    <span class="mat-caption text-secondary" i18n>Recording...</span>
  </ng-container>
```

src/assets/wise5/components/openResponse/speech-to-text/speech-to-text.component.ts
hirokiterashima: Yah, good suggestion. I struggled with naming when I was working on this. I ended up going with recordingRequester.

hirokiterashima:
Thx for your suggestions. All is working now.

PR #1635, opened 2024-02-15
feat(Speech to Text): Persist language selection across components

hirokiterashima:
## Changes
- Use a shared signal in TranscribeService to store the user's language selection. This way, students will no longer need to choose their preferred language each time they need to transcribe.
   - This required moving the initialization code (for ```selectedLanguage``` and ```languages```) from SpeechToTextComponent to TranscribeService

## Test
- Choose a language on one component
   - It should immediately update the selection on another component (if on the same step)
   - The selection should persist across steps too
- Transcription works as before

### DIFF
--- a/src/assets/wise5/services/transcribeService.ts
+++ b/src/assets/wise5/services/transcribeService.ts
@@ -18,8 +18,8 @@ window.process = process;
 window.Buffer = Buffer;
 
 export interface Language {
-  languageCode: LanguageCode;
-  language: string;
+  code: LanguageCode;
+  name: string;
 }
 
 @Injectable()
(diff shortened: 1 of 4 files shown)


src/assets/wise5/services/transcribeService.ts
```diff
@@ -11,21 +11,54 @@ import { StartStreamTranscriptionCommand } from '@aws-sdk/client-transcribe-stre
...
 import { ConfigService } from './configService';
+import { ProjectService } from './projectService';
 
 window.process = process;
 // @ts-ignore
 window.Buffer = Buffer;
 
+export interface Language {
+  languageCode: LanguageCode;
+  language: string;
```
geoffreykwan: These variable can be ordered alphabetically.

src/assets/wise5/services/transcribeService.ts
```diff
@@ -11,21 +11,54 @@ import { StartStreamTranscriptionCommand } from '@aws-sdk/client-transcribe-stre
...
+  constructor(private configService: ConfigService, private projectService: ProjectService) {
+    const { defaultLanguage, supportedLanguages } = this.projectService.project.speechToText;
+    this.selectedLanguageSignal.set({
+      languageCode: defaultLanguage,
+      language: this.allLanguages.find((language) => language.languageCode === defaultLanguage)
+        .language
+    });
+    this.languages = this.allLanguages.filter((language) =>
+      supportedLanguages.includes(language.languageCode)
+    );
```
geoffreykwan: The comparison between languageCode and defaultLanguage seems a little off from the standpoint of just reading the code. The objects contain a languageCode and a language field so you kind of expect the language field to be compared to the defaultLanguage. Maybe if we change the name of the language field to languageName or something it would be more clear.

geoffreykwan (approved):
Functionality works well.

src/assets/wise5/services/transcribeService.ts
geoffreykwan: Another option is to change the Language interface to be
```
export interface Language {
  code: LanguageCode;
  name: string;
}
```

hirokiterashima: Thanks for the suggestion. Renaming the Language fields made the code clearer.

PR #1636, opened 2024-02-16
refactor(Grade By Student): reduce duplication and complexity

breity:
## Changes
- Use ng-template for th elements
- Replace long switch statement with object literal

## Test
- Make sure sorting and filtering still works as before in Grade by Student view
- Ensure that First Name and Last Name columns are only visible when teacher has permission to view names
- Ensure that Student column is only visible when teacher doesn't have permission to view names

### DIFF
--- a/src/assets/wise5/classroomMonitor/student-progress/student-progress.component.html
+++ b/src/assets/wise5/classroomMonitor/student-progress/student-progress.component.html
@@ -1,3 +1,27 @@
+<ng-template #thead let-sortBy="sortBy">
+  <th class="table--list__thead__th" sticky>
+    <button
+      mat-button
+      class="table--list__thead__link"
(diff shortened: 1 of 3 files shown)


hirokiterashima (approved):
LGTM. Template is much simpler with the new changes.

Do you need to wait until ngOnInit() to initialize the sortOptions? If you initialize the sortOption class field when you declare it, it still seems to work:
```
export class StudentProgressComponent implements OnInit {
  ...
  sortOptions: any = {
    team: {
      label: $localize`Team`,
      method: this.createSortTeam
    },
    ...
```
Then, you can remove the ```initializeSorting()``` function and call ```this.sortWorkgroups()``` directly in ngOnOnit()?

breity:
@hirokiterashima Yep, makes sense. Thanks.

PR #1641, opened 2024-02-20
fix(Authoring): Move inactive lesson duplicates steps

### LINKED ISSUE #1640: bug(Authoring): Move inactive lesson after another inactive lesson will duplicate steps
2. Create or move two lessons (that have at least one step) into the Unused Lessons section
3. Move the top unused lesson after the other unused lesson
4. The steps in the unused lesson that was moved will now be duplicated in the inactiveNodes within the project.json. These nodes should only show up once in the inactiveNodes array.

geoffreykwan:
## Changes
- When moving an inactive lesson after another inactive lesson, check if its child steps are already in the inactive array and do not add them if they are already in the inactive array
- Added inactive lessons to the demo project used for testing

## Test
2. Create or move two lessons (that have at least one step) into the Unused Lessons section
3. Move the top unused lesson after the other unused lesson
4. Make sure the steps in the unused lesson that was moved were not duplicated in the inactiveNodes within the project.json. Previously these steps would show up twice in the inactiveNodes array but now they should only show up once.

Closes #1640

### DIFF
--- a/src/assets/wise5/services/teacherProjectService.ts
+++ b/src/assets/wise5/services/teacherProjectService.ts
@@ -2574,9 +2574,11 @@ export class TeacherProjectService extends ProjectService {
    */
   addGroupChildNodesToInactive(node) {
     for (const childId of node.ids) {
-      const childNode = this.getNodeById(childId);
-      this.project.inactiveNodes.push(childNode);
-      this.inactiveStepNodes.push(childNode);
+      if (!this.isInactive(childId)) {
+        const childNode = this.getNodeById(childId);
+        this.project.inactiveNodes.push(childNode);
+        this.inactiveStepNodes.push(childNode);
+      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


src/assets/wise5/services/teacherProjectService.ts
```diff
@@ -2574,9 +2574,11 @@ export class TeacherProjectService extends ProjectService {
    */
   addGroupChildNodesToInactive(node) {
     for (const childId of node.ids) {
-      const childNode = this.getNodeById(childId);
-      this.project.inactiveNodes.push(childNode);
-      this.inactiveStepNodes.push(childNode);
+      if (!this.isInactive(childId)) {
```
hirokiterashima, line 2577: Can we use isActive() to avoid double negation?
```suggestion
      if (this.isActive(childId)) {
```

hirokiterashima (approved):
LGTM. I added code comment inline.

src/assets/wise5/services/teacherProjectService.ts
geoffreykwan: Looks like using isActive() doesn't work because at that point in time the steps have been taken out of the active nodes array so isActive() would return false and then the steps would not get added to the inactive nodes array even when they should. This would specifically break the action of moving an active lesson to the unused lessons section.

hirokiterashima: I see. So this is a state where the node is NOT active, and also is NOT inactive? Like in-between? Maybe we should have a function that checks for that instead? When I read that "X is not inactive", I take it to mean that "X is active".

PR #1644, opened 2024-02-21
feat(AI Chat): Implement authoring view

geoffreykwan:
## Changes
- Implemented authoring view for the AI Chat item

Note: I will create another PR for restricting who can create an AI Chat item

## Test
- AI Chat "Add New Component", "More Info" demo
- Adding an AI Chat item
- Authoring the parameters of the AI Chat item
- Previewing the AI Chat item using "Preview Component"
- Changing the model in the AI Chat authoring advanced view

### DIFF
--- a/src/assets/wise5/components/aiChat/ai-chat-authoring/ai-chat-authoring.component.ts
+++ b/src/assets/wise5/components/aiChat/ai-chat-authoring/ai-chat-authoring.component.ts
@@ -1,7 +1,5 @@
 import { Component } from '@angular/core';
 import { AbstractComponentAuthoring } from '../../../authoringTool/components/AbstractComponentAuthoring';
-import { DialogWithCloseComponent } from '../../../directives/dialog-with-close/dialog-with-close.component';
-import { MatDialog } from '@angular/material/dialog';
 import { ProjectAssetService } from '../../../../../app/services/projectAssetService';
 import { ConfigService } from '../../../services/configService';
 import { TeacherNodeService } from '../../../services/teacherNodeService';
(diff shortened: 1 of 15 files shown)


src/assets/wise5/components/aiChat/ai-chat-authoring/ai-chat-authoring.component.ts
```diff
@@ -0,0 +1,9 @@
+import { Component } from '@angular/core';
+import { AbstractComponentAuthoring } from '../../../authoringTool/components/AbstractComponentAuthoring';
+
+@Component({
+  selector: 'ai-chat-authoring',
```
hirokiterashima: Not used, remove?
```suggestion
```

src/assets/wise5/components/aiChat/edit-ai-chat-advanced/edit-ai-chat-advanced.component.ts
```diff
@@ -0,0 +1,12 @@
+import { Component } from '@angular/core';
+import { EditAdvancedComponentComponent } from '../../../../../app/authoring-tool/edit-advanced-component/edit-advanced-component.component';
+import { AiChatContent } from '../AiChatContent';
+
+@Component({
+  selector: 'edit-ai-chat-advanced',
+  templateUrl: './edit-ai-chat-advanced.component.html',
+  styleUrls: ['./edit-ai-chat-advanced.component.scss']
```
hirokiterashima: Scss is an empty file, remove this line and the file?
```suggestion
```

hirokiterashima (commented):
Works well.

Should we make the model into a select dropdown in the advanced view?

It'd be good to work on the phrasing a bit in the hints for the System Prompt and help them thing about writing a good system prompt...maybe add some examples and tips for what makes a good system prompt? We should have hints for the prompt too, to distinguish between it and the System Prompt and the computer avatar fields.

Also, I added some code improvement suggestion comments inline.

src/assets/wise5/components/aiChat/ai-chat-authoring/ai-chat-authoring.component.ts
geoffreykwan: We don't remove the selector for any of the other item authoring components. Should we be consistent?

hirokiterashima:
New changes look good. 

Should EditAiChatAdvancedComponent.models be a protected field?

Let's work on the the hints for the inputs.

hirokiterashima:
Good to have more info about the system prompt. I'm wondering if it should be a "show/hide more info below" kind of UX instead of the popup? It might be good to be able to see the explanation and examples as the author is typing. Let's discuss.

breity:
@hirokiterashima That's a good point. I agree that show/hide for more info might be more useful here. Yes, let's discuss.

I'm also wondering how we should handle authoring input fields that have hints once multiple language units are supported. We will need to make sure both hints (the informational ones like we have in this PR and the one showing the default language text) are displayed when user is translating into a different language. But we'll probably also want to distinguish them visually somehow.

geoffreykwan:
I've moved the system prompt instructions and example to below the system prompt textarea.

geoffreykwan:
I've moved the system prompt instructions out of the mat-hint and placed it above the system prompt. I also added prompt instructions above the prompt.

hirokiterashima (approved):
LGTM

breity (approved):
LGTM. 👍 

I made some minor updates to the instructions text and styles.

PR #1653, opened 2024-02-23
refactor(Export): Move select controls into another component

geoffreykwan:
## Changes
- Moved the Export All Steps and Select Steps To Export Select All, and Deselect All controls into the SelectStepAndComponentCheckboxesComponent since those controls are only used when using that component

## Test
- Make sure these export types work. Within these export types, try exporting "Export All Steps" and "Select Steps To Export".
  - One Workgroup Per Row
  - Latest Work
  - All Work
  - Events
  - Raw Data


hirokiterashima (approved):
LGTM.

Is the ```projectIdToOrder``` null-check necessary in the ```selectAll()``` function? Seems like it should always be set to something by DataExportComponent?

``` 
  protected selectAll(doSelect: boolean = true): void {
    if (this.projectIdToOrder != null) {
```

geoffreykwan:
Yea looks like that null check wasn't needed.

PR #1655, opened 2024-02-25
refactor(Authoring): Move event stop propagation to step template

geoffreykwan:
## Changes
- In the ProjectAuthoringStepComponent, moved $event.stopPropagation() calls from the functions to the template

## Test
- Make sure that in the project authoring view, the step move, copy, and delete functionality still works properly


hirokiterashima (approved):
LGTM

PR #1657, opened 2024-02-26
feat(User): Change role to roles

### LINKED ISSUE #1656: feat(User): Change role to roles
Right now the user object has a role field which is a single string. Change this field to a roles field which is an array of strings. Users can actually have multiple roles so this will allow us to check if a user has a specific role.

geoffreykwan:
## Changes
- User object now has a roles string array field instead of a role string field
- Updated all the places that look at the role field and make them work with the roles array instead

## Test
- Test with https://github.com/WISE-Community/WISE-API/pull/258
- Make sure the header displays the correct links depending on the roles
- Make sure the header account menu displays the correct elements depending on the roles

Closes #1656


src/app/modules/mobile-menu/mobile-menu.component.ts
```diff
@@ -26,7 +25,7 @@ export class MobileMenuComponent implements OnInit {
 
   getUser() {
     this.userService.getUser().subscribe((user) => {
-      if (user && user.role) {
+      if (user.roles?.length > 0) {
         this.signedIn = true;
       } else {
         this.signedIn = false;
```
hirokiterashima: Rewrite this if/else block with in one line?
``` this.signedIn = user.roles?.length > 0 ```

hirokiterashima (approved):
LGTM. I added one code improvement suggestion inline.

src/app/modules/mobile-menu/mobile-menu.component.ts
geoffreykwan: Ok, I've made that change.

PR #1658, opened 2024-02-26
refactor(Authoring): Remove classes only used for testing

geoffreykwan:
## Changes
- Remove classes only used for testing
- Use matTooltip attribute value to select elements when testing

## Test
- Make sure the move, copy, and delete buttons still work


hirokiterashima (approved):
LGTM

PR #1662, opened 2024-02-27
feat(Authoring): Implement add step button menu next to all steps

### LINKED ISSUE #1661: feat(Authoring): Implement add step button menu next to all steps
Add an add step button next to all the steps in the project view. When the button is clicked, it should open a menu that lets the author "Add Step Before" or "Add Step After". When either of those are chosen, a new step should be added named "New Step" and should have no components.

geoffreykwan:
## Changes
- Added an add button next to all the steps in the project view
- Clicking on the add button will open a menu that has the options "Add Step Before" and "Add Step After"
- Clicking on either of the options will create a new step named "New Step"
- The new step will have no components

## Test
- Use the new add step button menus to add a step
  - before the first step in a lesson
  - before a step that is not the first step in a lesson
  - after a step

Closes #1661

### DIFF
--- a/src/assets/wise5/authoringTool/project-authoring-lesson/project-authoring-lesson.component.html
+++ b/src/assets/wise5/authoringTool/project-authoring-lesson/project-authoring-lesson.component.html
@@ -96,8 +96,18 @@
         </button>
       </mat-menu>
     </div>
-    <div *ngIf="lesson.ids.length === 0" class="no-steps-message" fxLayoutAlign="start center" i18n>
-      This lesson has no steps
+    <div *ngIf="lesson.ids.length === 0" class="no-steps-message" fxLayoutAlign="start center">
+      <div i18n>This lesson has no steps</div>
+      <button
+        mat-icon-button
+        color="primary"
+        (click)="addStepInside(lesson.id)"
+        matTooltip="Add step"
+        matTooltipPosition="above"
+        i18n-matTooltip
+      >
+        <mat-icon>add_circle</mat-icon>
+      </button>
     </div>
   </div>
 </div>
--- a/src/assets/wise5/authoringTool/project-authoring-lesson/project-authoring-lesson.component.html
+++ b/src/assets/wise5/authoringTool/project-authoring-lesson/project-authoring-lesson.component.html
@@ -76,29 +76,25 @@
         [projectId]="projectId"
         fxFlex
       ></project-authoring-step>
-      <div class="div-next-to-step">
-        <button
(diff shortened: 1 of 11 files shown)


hirokiterashima (changes requested):
Looks good.

Here are the changes that we discussed today:
1. Always show the "Add step button" instead of just on hover.
2. Add "Add step button" to empty lesson, next to the text "This lesson has no steps"

Also- do we want to allow authors to add steps without going through the "Add step wizard"? This would be a new behavior. I'm wondering if we should make this merge to a new feature branch instead of into develop. For the next task in this feature branch, we can connect this new "Add step button" to go to the "Add step wizard". What do you think?

geoffreykwan:
I've made the two changes. I think we can merge this into develop and make a new PR for the add step wizard. Authors can still use the add new step button at the top of the screen for now.

hirokiterashima:
New changes look good. After further testing, I found possible issues adding steps in a branch activity.

1. The newly added step does not have the same color as the other steps in the same branch path. Here, I added a step after step 1.3A:
Screenshot 2024-02-28 at 5 18 24 PM

The new step 1.4A should have the green color.


2. When I add a step before the first step in a branch path, the behavior is not what I expect:

Before adding a step before step 1.2A:
Screenshot 2024-02-28 at 5 13 49 PM

After:
Screenshot 2024-02-28 at 5 13 58 PM

In this scenario, I expected the new step to be the new step 1.2A. But as you can see, the new step became a branch point. If this is hard to fix, maybe we can disable the "add step before" option to the first step in a branch path?

hirokiterashima:
New changes look good. After further testing, I found possible issues adding steps in a branch activity.

1. The newly added step does not have the same color as the other steps in the same branch path. Here, I added a step after step 1.3A:
Screenshot 2024-02-28 at 5 18 24 PM

The new step 1.4A should have the green color.


2. When I add a step before the first step in a branch path, the behavior is not what I expect:

Before adding a step before step 1.2A:
Screenshot 2024-02-28 at 5 13 49 PM

After:
Screenshot 2024-02-28 at 5 13 58 PM

In this scenario, I expected the new step to be the new step 1.2A. But as you can see, the new step became a branch point. If this is hard to fix, maybe we can disable the "add step before" option to the first step in a branch path?

geoffreykwan:
I've fixed the background color issue and disabled the "add step before" button for steps that are the first step in a branch path.

hirokiterashima (approved):
LGTM

PR #1667, opened 2024-02-29
fix(Classroom Monitor): Grade by Student sort by location

### LINKED ISSUE #1666: bug(Classroom Monitor): Grade by Student sort by location errors
## Description
Sort by location in Grade by Student doesn't always order the students correctly because it is sorting numerically using the node position and title, which is this format: "2.2: Newton Scooter Concepts". Because we're sorting using `localeComapreBy`, step 2.12 will be considered as before step 2.2, for example, because the number 2.12 is less than 2.2. But in our naming structure, 2.12 means lesson 2, step 12. So it should come after 2.2 (lesson 2, step 2).

For example, students on step 1.12 should be sorted below students on 1.5 and 1.8 in this example (sort is by location ascending):

Image

breity:
## Changes
- Use node order in unit to perform sort by location in Classroom Monitor Grade by Student.

## Test
- Make sure sorting by location in Grade by Student now performs correctly in all cases (e.g. students on step 2.12 come after students on step 2.2 when sorted by location ascending).

Closes #1666.

### DIFF
--- a/src/assets/wise5/classroomMonitor/classroomMonitorComponents/shared/testing/ClassroomMonitorTestHelper.ts
+++ b/src/assets/wise5/classroomMonitor/classroomMonitorComponents/shared/testing/ClassroomMonitorTestHelper.ts
@@ -10,6 +10,7 @@ export class ClassroomMonitorTestHelper {
   students: StudentProgress[] = [
     new StudentProgress({
       location: '1.2: Open Response',
+      order: 2,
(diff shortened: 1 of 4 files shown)


hirokiterashima (approved):
LGTM.

Can you also add type to the workgroupId param before merge?
```getCurrentNodeLocationForWorkgroupId(workgroupId: number): any {```

PR #1669, opened 2024-02-29
feat(Export): Create export events route

### LINKED ISSUE #1668: feat(Export): Create export events route
Move the export events code out of data-export.component.ts and into a new component/route.

geoffreykwan:
## Changes
- Created export events route
- Moved code from data-export.component.ts to export-events.component.ts

## Test
- Make sure the export events works

Closes #1668

### DIFF
--- a/src/assets/wise5/classroomMonitor/dataExport/abstract-export.component.ts
+++ b/src/assets/wise5/classroomMonitor/dataExport/abstract-export.component.ts
@@ -86,4 +86,26 @@ export abstract class AbstractExportComponent {
     }
     return selectedNodes;
   }
+
+  /**
+   * @param users An array of user objects. Each user object contains an id and name.
+   * @returns {object} An object that contains key/value pairs. The key is userIdX
+   * or studentNameX where X is an integer. The values are the corresponding actual
+   * values of user id and student name.
+   */
+  extractUserIDsAndStudentNames(users: any[]): any {
+    const extractedUserIDsAndStudentNames = {};
+    for (let u = 0; u < users.length; u++) {
+      let user = users[u];
+      extractedUserIDsAndStudentNames['userId' + (u + 1)] = user.id;
+      if (this.canViewStudentNames) {
+        extractedUserIDsAndStudentNames['studentName' + (u + 1)] = user.name;
+      }
+    }
+    return extractedUserIDsAndStudentNames;
+  }
+
+  protected goBack(): void {
+    this.router.navigate(['..'], { relativeTo: this.route });
+  }
 }
--- a/src/assets/wise5/classroomMonitor/dataExport/abstract-export.component.ts
+++ b/src/assets/wise5/classroomMonitor/dataExport/abstract-export.component.ts
@@ -93,16 +93,16 @@ export abstract class AbstractExportComponent {
    * or studentNameX where X is an integer. The values are the corresponding actual
    * values of user id and student name.
    */
-  extractUserIDsAndStudentNames(users: any[]): any {
-    const extractedUserIDsAndStudentNames = {};
+  extractUserIDsAndStudentNames(users: any[]): Record<string, number | string> {
+    const userIDsAndStudentNames = {};
     for (let u = 0; u < users.length; u++) {
       let user = users[u];
-      extractedUserIDsAndStudentNames['userId' + (u + 1)] = user.id;
+      userIDsAndStudentNames['userId' + (u + 1)] = user.id;
       if (this.canViewStudentNames) {
-        extractedUserIDsAndStudentNames['studentName' + (u + 1)] = user.name;
+        userIDsAndStudentNames['studentName' + (u + 1)] = user.name;
       }
     }
-    return extractedUserIDsAndStudentNames;
+    return userIDsAndStudentNames;
   }
 
   protected goBack(): void {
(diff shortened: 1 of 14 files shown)


src/assets/wise5/classroomMonitor/dataExport/abstract-export.component.ts
```diff
@@ -86,4 +86,26 @@ export abstract class AbstractExportComponent {
...
   }
+
+  /**
+   * @param users An array of user objects. Each user object contains an id and name.
+   * @returns {object} An object that contains key/value pairs. The key is userIdX
+   * or studentNameX where X is an integer. The values are the corresponding actual
+   * values of user id and student name.
+   */
+  extractUserIDsAndStudentNames(users: any[]): any {
+    const extractedUserIDsAndStudentNames = {};
```
hirokiterashima: Do we need the extracted prefix?
```suggestion
    const userIDsAndStudentNames = {};
```

src/assets/wise5/classroomMonitor/dataExport/export-events/export-events.component.ts
```diff
@@ -0,0 +1,22 @@
...
+
+@Component({
+  selector: 'export-events',
+  templateUrl: './export-events.component.html',
+  styleUrls: ['./export-events.component.scss']
+})
+export class ExportEventsComponent extends AbstractExportComponent {
+  public includeNames: boolean = false;
+  public includeStudentEvents: boolean = true;
+  public includeTeacherEvents: boolean = true;
```
hirokiterashima, line 13: Do these fields need to be public?

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractDataExportStrategy.ts
```diff
@@ -162,4 +163,8 @@ export abstract class AbstractDataExportStrategy implements DataExportStrategy {
       `${componentNumber}_${componentTypeWithUnderscore}_work.csv`
     );
   }
+
+  generateCSVFile(rows: any[], fileName: string): void {
+    generateCSVFile(rows, fileName);
+  }
```
hirokiterashima, line 169: Should we call generateCSVFile() directly and not add this function?

src/assets/wise5/classroomMonitor/dataExport/abstract-export.component.ts
```diff
@@ -86,4 +86,26 @@ export abstract class AbstractExportComponent {
...
     return selectedNodes;
   }
+
+  /**
+   * @param users An array of user objects. Each user object contains an id and name.
+   * @returns {object} An object that contains key/value pairs. The key is userIdX
+   * or studentNameX where X is an integer. The values are the corresponding actual
+   * values of user id and student name.
+   */
+  extractUserIDsAndStudentNames(users: any[]): any {
```
hirokiterashima: Use [Record](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type) to specify types of key and value on the returned object?


```suggestion
  extractUserIDsAndStudentNames(users: any[]): Record<string, number | string>  {
```

hirokiterashima (commented):
Functionality works.

Should we require users to select "Include Student Events" or "Include Teacher Events" or both before they can click the export button? The export wouldn't be that useful unless they select one or both of those options.

Also, I added some comments about code inline. Lmk what you think.

geoffreykwan:
The events export now requires at least one of the student or teacher events checkboxes to be checked before you can export. I've updated the extractUserIDsAndStudentNames() function with the suggested changes.

src/assets/wise5/classroomMonitor/dataExport/export-events/export-events.component.ts
geoffreykwan: These need to be public due to the coupling of the strategy with the controller. Other export strategies have this problem too. We should refactor all of them later after all the exports have their own route/component.

https://github.com/WISE-Community/WISE-Client/blob/0c29088c2b2f30180b2a3d8b1c7275f961257482/src/assets/wise5/classroomMonitor/dataExport/strategies/EventDataExportStrategy.ts#L6-L17

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractDataExportStrategy.ts
geoffreykwan: This is required for testing since I couldn't find an easy way to monitor the parameters that are sent to generateCSVFile() if we use the function directly from csv.ts. If we could find a better solution to this in the future, that would be good.

src/assets/wise5/classroomMonitor/dataExport/export-events/export-events.component.ts
hirokiterashima: Ok. Sounds good.

src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractDataExportStrategy.ts
hirokiterashima: Ok, let's refactor this in the future. It's not ideal to have to write extra code just to make the tests pass.

hirokiterashima (approved):
LGTM.

PR #1670, opened 2024-03-01
feat(Multiple Languages): Disable default language editor

hirokiterashima:
## Changes
- Disable the default language editor to prevent authors from editing the default language HTML content when in translation mode.

## Test prep
- Test with https://github.com/WISE-Community/WISE-API/pull/253. Make sure to get the latest changes.

## Test HTMLComponent in AT with translation enabled unit
- When in another language
   - you can translate the content in the first tab (translation language)
   - you cannot edit the content in the second tab (default langauge)
- When in default language
   - you can edit the content


geoffreykwan:
Functionality works. When in the default language view, should we have some kind of note or explanation as to why it isn't editable? I imagine some authors would expect to be able to edit the default language in that view.

hirokiterashima:
Good idea. I added a message in the default language tab.

geoffreykwan (approved):
Looks good.

PR #1674, opened 2024-03-06
fix(Authoring): Fix moving lesson above first empty lesson breaks unit

### LINKED ISSUE #1673: fix(Authoring): Moving lesson above first empty lesson breaks unit
Note: Performing these steps will break the unit and it will no longer load
1. Open or create a unit in the Authoring Tool
2. Create a lesson with no steps and make it the first lesson in the unit
3. Move a lesson that contains at least one step and make it the first lesson in the unit. This will immediately break the Authoring Tool.

geoffreykwan:
## Changes
- Authors can now move a lesson that contains steps above the first empty lesson without breaking the unit
- After this action is performed, for the lesson that contains steps, the last step in the lesson will have no transitions. Everything still works because the lesson that has steps still has a transition to the empty lesson.

## Test
1. Open or create a unit in the Authoring Tool
2. Create a lesson with no steps and make it the first lesson in the unit
3. Move a lesson that contains at least one step and make it the first lesson in the unit. This used to immediately break the Authoring Tool but now it should not.

Closes #1673

### DIFF
--- a/src/assets/wise5/services/teacherProjectService.ts
+++ b/src/assets/wise5/services/teacherProjectService.ts
@@ -2267,18 +2267,9 @@ export class TeacherProjectService extends ProjectService {
                 const oldToGroup = this.getNodeById(oldToGroupId);
                 if (oldToGroup != null) {
                   const oldToGroupStartId = oldToGroup.startId;
-                  const transition = {};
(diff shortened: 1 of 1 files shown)


hirokiterashima (approved):
LGTM

PR #1677, opened 2024-03-07
fix(Authoring): Fix deleting first lesson when second lesson is empty

### LINKED ISSUE #1676: Bug(Authoring Tool): Deleting all lessons at the same time breaks unit
### Description
Deleting all lessons at the same time breaks the unit. 

### Steps to reproduce:
1. Create a new unit
2. Add a new empty lesson after the first default lesson
3. Select both lessons and click the global delete icon in the top row

You will see this:
Screenshot 2024-03-06 at 5 03 14 PM

Refreshing the browser throws this js error and takes you back to the homepage:
```
src_app_teacher_teacher-authoring_module_ts.js:2 ERROR Error: Uncaught (in promise): TypeError: Cannot read properties of null (reading 'startId')
TypeError: Cannot read properties of null (reading 'startId')
    at TeacherProjectService.calculateNodeNumberForLesson (projectService.ts:1715:14)
    at TeacherProjectService.calculateNodeNumbersHelper (projectService.ts:1475:32)
    at TeacherProjectService.calculateNodeNumbers (projectService.ts:1451:10)
    at TeacherProjectService.parseProject (projectService.ts:226:10)
    at TeacherProjectService.setProject (projectService.ts:57:10)
    at Object.next (projectService.ts:786:14)
```

geoffreykwan:
## Changes
- Fix deleting first lesson when second lesson is empty. This used to break the unit.

## Test
1. Create a new unit
2. Add a new empty lesson after the first default lesson
3. Select both lessons and click the global delete icon in the top row. This used to break the unit but now it should not.

Closes #1676

### DIFF
--- a/src/assets/wise5/services/deleteNodeService.ts
+++ b/src/assets/wise5/services/deleteNodeService.ts
@@ -83,7 +83,7 @@ export class DeleteNodeService {
       let nextNodeId = transitions[0].to;
       if (this.ProjectService.isGroupNode(nextNodeId)) {
         const nextGroupStartId = this.ProjectService.getGroupStartId(nextNodeId);
-        if (nextGroupStartId == null) {
+        if (nextGroupStartId == null || nextGroupStartId === '') {
           this.ProjectService.setStartNodeId(nextNodeId);
         } else {
           this.ProjectService.setStartNodeId(nextGroupStartId);


hirokiterashima (approved):
LGTM

PR #1678, opened 2024-03-07
fix(MC): Clicking on checkbox not properly updating checked status

hirokiterashima:
## Changes
- Fix issue where clicking on checkbox in MC was not properly updating checked status. This may be related to changes in Angular

## Test
- Make sure that MC checkbox item with correctness works properly


geoffreykwan (approved):
Looks good.

PR #1681, opened 2024-03-10
feat(Translation): Convert table item inputs to translatable

geoffreykwan:
## Changes
- Converted Table item inputs to translatable inputs

## Test
- Author a Table item and make sure the table cells are translatable

### DIFF
--- a/src/assets/wise5/components/table/table-authoring/table-authoring.component.html
+++ b/src/assets/wise5/components/table/table-authoring/table-authoring.component.html
@@ -146,15 +146,14 @@
       </td>
       <td *ngFor="let cell of row" class="inner-cell-container">
         <div fxLayout="column" fxLayoutAlign="center center">
-          <mat-form-field class="cell-text">
-            <input
-              matInput
-              [(ngModel)]="cell.text"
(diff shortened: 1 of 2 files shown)


hirokiterashima (approved):
LGTM.

Unrelated to this PR- when I preview the unit, the language text don't show in the cells. I see default language text for all the cells instead. Do you see this too?

geoffreykwan:
Yea I see that problem too.

Problem occurs when
1. Load the unit preview (you will be in the default language English)
2. Go to the Table step
3. Switch to another language like Spanish
4. The text in the cells will incorrectly remain in English

Problem does not occur when
1. Load the unit preview not on the Table step (you will be in the default language English)
2. Switch to another language like Spanish
3. Go to the Table step
4. The text in the cells are correctly translated to Spanish

I think the problem is caused by the student data saving. When you switch language, it triggers an auto save. If the student data saves in English, it will remain in English even if you switch language. This can be a problem for other item types too depending on what kind of student data is saved. If the student starts the unit in Spanish and never changes the language, they won't run into this problem.

PR #1683, opened 2024-03-11
feat(Translation): Make animation authoring inputs translatable

breity:
## Changes
- Convert authoring input for Text field in animation components to translatable input.

Note: The other animation fields that should be translatable all include image choosers, which don't yet work with translatable inputs. These will be made translatable in a separate PR.

## Test
- Author an animation component and add a Text object.
- Make sure the "Text" field is translatable.


hirokiterashima (approved):
LGTM.

PR #1684, opened 2024-03-11
feat(Translation): Make concept map authoring inputs translatable

breity:
## Changes
- Convert authoring inputs for Node Label, Links Title, and Link Label fields in concept map components to translatable inputs.

Note: The Node Image field should also be translatable, but it is connected to a file (image) upload, which doesn't yet work with translatable inputs. This will be made translatable in a separate PR.

## Test
- Author a concept map component and add some nodes, link title and links.
- Make sure the fields listed above are translatable.


hirokiterashima (approved):
LGTM.

PR #1687, opened 2024-03-12
feat(Teacher): Implement apply tags button

geoffreykwan:
## Changes
- Implemented apply tags button in the Class Schedule view
- Tags can be applied to a unit
- Tags can be removed from a unit

Note: You can't create a new tag at the moment.

## Test
- Test with https://github.com/WISE-Community/WISE-API/pull/260
- In the Class Schedule view, select some units by clicking on their checkbox
- Click the "Apply tags" button and make sure you can add and remove a tag from a unit using the apply tags menu. You should have an "archived" tag already created if you have previously archived any units. In the future we might not want to display the "archived" tag in this menu.
- When you apply a tag, the tag should immediately show up in the unit card
- When you remove a tag, the tag should immediately disappear from the unit card

Note: Adding or removing the "archived" tag on a unit will not immediately move the unit to the active or archived section.

### DIFF
--- a/src/app/teacher/apply-tags-button/apply-tags-button.component.ts
+++ b/src/app/teacher/apply-tags-button/apply-tags-button.component.ts
@@ -1,0 +1,57 @@
+import { Component, Input, OnInit } from '@angular/core';
+import { Project } from '../../domain/project';
+import { TagService } from '../../../assets/wise5/services/tagService';
+
+@Component({
+  selector: 'apply-tags-button',
+  templateUrl: './apply-tags-button.component.html',
+  styleUrls: ['./apply-tags-button.component.scss']
+})
+export class ApplyTagsButtonComponent implements OnInit {
+  protected tags: any[] = [];
+  @Input() selectedProjects: Project[] = [];
+
+  constructor(private tagService: TagService) {}
+
+  ngOnInit(): void {
+    this.tagService.retrieveUserTags().subscribe((tags) => {
+      for (const tag of tags) {
+        tag.checked = this.doesAnyProjectHaveTag(tag);
+      }
+      this.tags = tags;
+    });
+  }
+
+  private doesAnyProjectHaveTag(tag: any): boolean {
+    for (const project of this.selectedProjects) {
+      if (project.tags.includes(tag.text)) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  protected toggleTagOnProjects(tag: any, checked: boolean): void {
+    if (checked) {
+      this.addTagToProjects(tag, this.selectedProjects);
+      this.tagService.applyTagToProjects(tag.text, this.selectedProjects);
+    } else {
+      this.removeTagFromProjects(tag, this.selectedProjects);
+      this.tagService.removeTagFromProjects(tag.text, this.selectedProjects);
+    }
+  }
+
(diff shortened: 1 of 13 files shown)


src/assets/wise5/services/tagService.ts
```diff
@@ -80,4 +82,25 @@ export class TagService {
...
     return this.getExistingTagNames().includes(tagName);
   }
+
+  retrieveUserTags(): Observable<any> {
+    return this.http.get(`/api/user/tags`);
+  }
+
+  applyTagToProjects(tag: any, projects: Project[]): Subscription {
+    const projectIds = projects.map((project) => project.id);
+    return this.http.put(`/api/projects/tag/${tag}`, projectIds).subscribe(() => {});
```
hirokiterashima: No need to pass in a function?
```suggestion
    return this.http.put(`/api/projects/tag/${tag}`, projectIds).subscribe();
```

src/assets/wise5/services/tagService.ts
```diff
@@ -80,4 +82,25 @@ export class TagService {
...
+  removeTagFromProjects(tag: any, projects: Project[]): Subscription {
+    let params = new HttpParams();
+    for (const project of projects) {
+      params = params.append('projectIds', project.id);
+    }
+    return this.http
+      .delete(`/api/projects/tag/${tag}`, {
+        params: params
+      })
+      .subscribe(() => {});
```
hirokiterashima: No need to pass in a function?
```suggestion
      .subscribe();
```

src/app/teacher/apply-tags-button/apply-tags-button.component.ts
```diff
@@ -0,0 +1,57 @@
...
+  private doesAnyProjectHaveTag(tag: any): boolean {
+    for (const project of this.selectedProjects) {
+      if (project.tags.includes(tag.text)) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  protected toggleTagOnProjects(tag: any, checked: boolean): void {
```
hirokiterashima: The second parameter indicates whether we're adding or removing the tag, so rename variable to match?
```suggestion
  protected toggleTagOnProjects(tag: Tag, addTag: boolean): void {
```

hirokiterashima (commented):
Functionality works as described. I agree that we shouldn't show the archived tag- this can be removed in the future before we merge to develop.

I have a few comments/questions:
- I'm wondering if the checkbox should behave like the SelectAllItemsCheckboxComponent: none, some, and all. If some of the selected units have the "archived" tag, it should show a minus sign instead of checked? Can we reuse the SelectAllItemsCheckboxComponent in the Apply tags dropdown?
- Is the class type of tag a string? If so, we can replace occurrences of "any" with "string".
- I added some code comments inline.

geoffreykwan:
I've created a Tag interface and used it for parameter types. I've also removed the subscribe empty functions.

I'll add the minus checkbox sign in a different PR. I'll see if we should use the SelectAllItemsCheckboxComponent for that functionality.

hirokiterashima (approved):
LGTM.

Before merge, can you alphabetize the fields in ApplyTagsButtonComponent?

Another thing to consider for the future is making ApplyTagsButtonComponent into a standalone component and removing the need for ApplyTagsButtonModule.

PR #1689, opened 2024-03-13
feat(Translation): Convert peer chat authoring inputs to translatable

geoffreykwan:
## Changes
- Convert Peer Chat authoring inputs to translatable inputs
- Moved the "Delete question" button out of the mat-form-field for question bank rule version 2 questions
- Set translatable textarea to have 100% width

## Test
- Author a Peer Chat and make sure these fields are translatable
  - Dynamic Prompt
    - Pre Prompt
    - Prompt Rule Prompt
    - Post Prompt
  - Question Bank
    - Custom Label
    - Question Bank Rule Question

Note: Question Bank Rule version 1 questions are not translatable because it uses an array of strings instead of an array of objects


src/app/authoring-tool/edit-question-bank-rules/edit-question-bank-rules.component.html
```diff
@@ -94,17 +83,45 @@ <h5 fxLayout="row" fxLayoutAlign="start center" fxLayoutGap="4px">
...
-                <mat-icon>add_circle</mat-icon>
-              </button>
+              <ng-container *ngIf="version === 2">
+                <translatable-textarea
+                  [content]="rule.questions[questionIndex]"
+                  key="text"
+                  [label]="
+                    rule.questions.length === 1 ? 'Question' : 'Question #' + (questionIndex + 1)
+                  "
+                  i18n-label
```
breity: I think we should probably move this label text to a function. Then we could split the translatable text into two separate entries so that we don't put the the conditional in the translation text itself. I think this would be better for translators.

breity (approved):
Looks good. 👍 

I made a suggestion about the label text for question bank questions. See the comment above. I also made a small style change.

I noticed that in some places we use self-closing tags for <translatable-textarea> elements, but here we have <translatable-textarea></translatable-textarea>. We should probably discuss and settle on a convention.

geoffreykwan:
Ok, I've created a function for the question label.

PR #1690, opened 2024-03-13
feat(Translation): Convert outside url authoring inputs to translatable

geoffreykwan:
## Changes
- Convert Outside URL authoring inputs to translatable inputs

## Test
- Author an Outside URL item and make sure the url field is translatable


breity (approved):
The translatable input field for the resource url works.

I think we should also allow users to be able to choose one of the provided open educational resources when they are editing a non-source language. Right now, if an author selects one of the OERs when editing a non-source language, the url for the source language changes to the newly selected OER url. This should probably be done in a separate PR.

PR #1691, opened 2024-03-13
feat(Translation): Convert match authoring inputs to translatable

geoffreykwan:
## Changes
- Convert Match authoring inputs to translatable inputs

## Test
- Author a Match item and make sure these fields are translatable
  - Choice Name
  - Source Bucket Name
  - Target Bucket Name
  - Feedback
  - Incorrect Position Feedback (you need to enable "Choices need to be ordered within buckets" and then set "Is Correct" for a choice in order for this input to show up)


breity (approved):
Looks good. Just needs `messages.xlf` to be updated to resolve conflicts.

PR #1692, opened 2024-03-14
feat(Multiple Languages): TranslatableAssetChooserComponent

hirokiterashima:
## Changes
- Create TranslatableAssetChooserComponent that allows translators to choose an asset for a certain field
- Animation Authoring
   - Convert object image input field to TranslatableInputComponent
   - Convert asset selector button to TranslatableAssetChooserComponent

## Notes
- Please style as you see fit. For some reason, the chooser button is wider (54px) compared to other buttons like the move up/down buttons (50px). I couldn't figure it out.
- When you are translating to another language, the changes are saved, but the input field does not immediately update with the newly-selected asset at the moment. This is an underlying issue that should be fixed in a separate PR. For now, collapse->expand the component, or switch->switch back languages to see the new asset in the input field.
- Right now, this class assumes that the asset is an image. We can parameterize TranslatableAssetChooserComponent to work with non-image assets (e.g. filenames) in the future

## Test (in Animation authoring)
- In default language mode, you can choose image assets using the button in the "Objects" section, and the changes are saved and reflected immediately in the input field.
- In translation mode, you can choose image assets using the button in the "Objects" section, and the changes are saved and NOT reflected immediately in the input field.
   -  Collapse->expand the component, or switch->switch back languages to see the new asset in the input field.
- Preview the component and make sure that the images appear correctly in different languages

### DIFF
--- a/src/app/teacher/component-authoring.module.ts
+++ b/src/app/teacher/component-authoring.module.ts
@@ -87,6 +87,7 @@ import { EditComponentAdvancedButtonComponent } from '../../assets/wise5/authori
 import { TranslatableInputComponent } from '../../assets/wise5/authoringTool/components/translatable-input/translatable-input.component';
 import { TranslatableTextareaComponent } from '../../assets/wise5/authoringTool/components/translatable-textarea/translatable-textarea.component';
 import { TranslatableRichTextEditorComponent } from '../../assets/wise5/authoringTool/components/translatable-rich-text-editor/translatable-rich-text-editor.component';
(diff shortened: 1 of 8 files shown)


breity (approved):
LGTM 👍 

I fixed the button icon spacing to match other button sin the authoring tool. Looks like we add duplicate overriding css in many parts of the authoring tool, so we should refactor this eventually.

hirokiterashima:
Thanks. Let's add refactoring to our to-do list.

PR #1694, opened 2024-03-15
refactor(Multiple Languages): Change TranslationText from signal<string> to string

hirokiterashima:
## Changes
- Change ```AbstractTranslatableFieldComponent.translationText``` type from ```signal<string>``` to just ```string```. This will make the field easier to maintain when the languages are switched in the language chooser and in the tab in the HTML authoring. 
   - For example, we no longer need the ```TranslatableRichTextEditorComponent.setLanguage()``` function.
- Convert ```TranslateProjectService.currentTranslations``` signal to an observable to know when to update the ```translationText``` model.

## Test
- Translating the input, textarea, and RichTextarea (in HTML authoring) fields work as before

### DIFF
--- a/src/assets/wise5/authoringTool/components/abstract-translatable-field/abstract-translatable-field.component.ts
+++ b/src/assets/wise5/authoringTool/components/abstract-translatable-field/abstract-translatable-field.component.ts
@@ -5,11 +5,14 @@ import { EditProjectTranslationService } from '../../../services/editProjectTran
 import { TeacherProjectService } from '../../../services/teacherProjectService';
 import { TranslateProjectService } from '../../../services/translateProjectService';
 import { generateRandomKey } from '../../../common/string/string';
+import { toObservable } from '@angular/core/rxjs-interop';
+import { Translations } from '../../../../../app/domain/translations';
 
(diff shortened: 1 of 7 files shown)


geoffreykwan (approved):
Looks good.

Slightly unrelated so you can fix this in another issue, I still see the problem of changing the default text and then switching to another language and the default text not showing the latest English value in the hint.

I was able to fix this problem by making the change below in the template files.
```
  <mat-hint>[{{ defaultLanguage.language }}] {{ content[key] }}</mat-hint>
```

I also noticed that in the AbstractTranslatableFieldComponent file that the defaultLanguageText was only being set once and would never change even if the author changed the default text. I think this is because the computed function does not reference any signal variables. I think this line can be deleted (and the signal variable too) if we make the change above in the template.
```
    this.defaultLanguageText = computed(() => this.content[this.key]);
```

hirokiterashima:
Fyi, I fixed the issue where default language text wasn't updating in this commit: https://github.com/WISE-Community/WISE-Client/commit/35d281cc1eac0152934bc47ec985ae89a6fc491b

PR #1696, opened 2024-03-17
feat(AI Chat): Restrict authors to admin, researcher, and trusted authors

geoffreykwan:
## Changes
- Only allow admin, researcher, and trusted authors to add AI Chat items in the Authoring Tool

## Test
- Test with https://github.com/WISE-Community/WISE-API/pull/252 (make sure to pull for latest code)
1. Teacher not allowed to author AI Chat
   - Sign in as a teacher that is not an admin, researcher, or trusted author
   - Go to the Authoring Tool and go to the view where you add a step or item
   - You should not see AI Chat as an item that you can add

2. Teacher allowed to author AI Chat
   - Sign in as a teacher that is an admin, researcher, or trusted author
   - Go to the Authoring Tool and go to the view where you add a step or item
   - You should see AI Chat as an item that you can add


hirokiterashima (approved):
LGTM. 

This can be in a separate task- we should refactor the code to better encapsulate the ```User``` class. I think ```User.roles``` should be a private field in the User class and UserService shouldn't know that roles is an array. It should only access roles via a public function, like via a ```User.hasRole(role: string): boolean {...}``` function. I'll write this up in a separate issue.

PR #1703, opened 2024-03-19
refactor(Multiple Choice): remove isDisabled checks

### LINKED ISSUE #1679: refactor(MC): Remove isDisabled checks
### Description
In MultipleChoiceStudentComponent, radioChoiceSelected() and updateStudentChoices(), we have this check
```
    if (this.isDisabled) {
      return;
    }
```

We don't need this check because these functions will never be called if the component is disabled.

hirokiterashima:
## Changes
- Removed isDisabled checks in MC component because they will never be called from the template when the component is disabled (after max submits have been exhausted, or in Show Work).

## Test
- MC in student view works as before, when the choices are enabled and disabled (after max submits have been exhausted, or in Show Work).

Closes #1679


geoffreykwan (approved):
Looks good.

PR #1704, opened 2024-03-19
feat(Multiple languages): Copy default language in HTML component

hirokiterashima:
## Changes
- Allow translator to copy content from default language in HTML component authoring

## Notes
- Please style the button as you see fit

## Test Prep
- Test with ```at-units-in-multiple-languages``` branch in API

## Test (in HTML component authoring for a translated unit)
- "Copy content from [default language]" and "Copy content to [translation language]" buttons appear in the translation tab and default language tab and correctly shows the languages
- Click on the "Copy content from [default language]" button in the translation tab
   - If the translation language text is empty, it should copy the text
   - If the translation language text is not empty, it should confirm with the author before copying
- Click on the "Copy content to [translation language]" button in the default language tab
   - If the translation language text is empty, it should copy the text and switch to the translation tab
   - If the translation language text is not empty, it should confirm with the author before copying and switching to the translation tab


breity (approved):
Functionality looks good. 👍 

I updated styles and layout.

PR #1706, opened 2024-03-20
fix(Authoring): Fix max submit authoring

### LINKED ISSUE #1705: fix(Authoring): Max submit does not save
When you try to set the max submit for some items, it does not save. It does not work for these items
- Draw
- Graph
- Label
- Match
- Multiple Choice
- Table

geoffreykwan:
## Changes
- Fixed max submit authoring that was broken for some items

## Test
Make sure max submit authoring works for these items (almost all of them are in the advanced editing popup)
- Dialog Guidance (not in advanced editing popup)
- Draw
- Graph
- Label
- Match
- Multiple Choice
- Table

Closes #1705


hirokiterashima (approved):
LGTM

PR #1707, opened 2024-03-21
feat(Tag): Implement create tag dialog

geoffreykwan:
## Changes
- Implemented create tag dialog

## Test
- Test with https://github.com/WISE-Community/WISE-API/pull/262
- In the Class Schedule view, select some units by clicking on their checkbox
- Click the "Manage Tags" menu item
- Click the "+ New Tag" button
- Enter a name for the new tag
- Click "Create". The new tag should be added to the end of the tags list.

### DIFF
--- a/src/app/teacher/apply-tags-button/apply-tags-button.component.ts
+++ b/src/app/teacher/apply-tags-button/apply-tags-button.component.ts
@@ -33,6 +33,11 @@ export class ApplyTagsButtonComponent implements OnInit {
         tag.text = tagThatChanged.text;
       })
     );
+    this.subscriptions.add(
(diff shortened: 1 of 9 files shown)


hirokiterashima (changes requested):
Functionality works as described. I had a few thoughts while reviewing:

1. Can you add a snackbar save confirmation when a new tag is created?
2. Right now, you can create multiple tags with the same text. I think each tag text should be unique. The server-side should check and prevent this from happening?

geoffreykwan:
I've added the 'Tag created' message when a tag is created.

We can prevent duplicate tags on the client and server side in another PR.

hirokiterashima (approved):
LGTM

PR #1711, opened 2024-04-01
feat(Authoring): Remove add step before option on steps after the first

### LINKED ISSUE #1710: Remove "add step before" option
### Description

If the step is not the first step in the lesson, remove the "add step before" option. This option should only appear if it's the first step.

Also update the tooltip to reflect the change.

This will help reduce the number of clicks for the author.

geoffreykwan:
## Changes
- Created add step button component
- If the step is the first in a lesson, clicking the button will show a menu with the options "Add Step Before" and "Add Step After"
- If the step is not the first in a lesson, clicking the button will go directly to the add step route

## Test
- Open a unit in the Authoring Tool
- If the step is the first in a lesson, clicking the button shows a menu with the options "Add Step Before" and "Add Step After"
- If the step is not the first in a lesson, clicking the button will go directly to the add step route
- If a lesson has no steps, clicking the "Add Step" button will go directly to the add step route

Closes #1710

### DIFF
--- a/src/app/teacher/authoring-tool.module.ts
+++ b/src/app/teacher/authoring-tool.module.ts
@@ -70,7 +70,6 @@ import { AddStepButtonComponent } from '../../assets/wise5/authoringTool/add-ste
     AddLessonChooseTemplateComponent,
     AddLessonConfigureComponent,
     AddProjectComponent,
-    AddStepButtonComponent,
     AddYourOwnNode,
     AdvancedProjectAuthoringComponent,
     AuthoringToolComponent,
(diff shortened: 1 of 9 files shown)


hirokiterashima (commented):
The functions in AddStepButtonComponent don't look quite right to me. I think it'd be better if ```addStepBefore()``` and ```addStepAfter()``` both call a private function that calls ```this.router.navigate(...``` to reduce duplication.

Also, should AddStepButtonComponent be a standalone component?

geoffreykwan:
I've cleaned up the code, converted AddStepButtonComponent to be standalone, and changed the capitalization for the menu text.

hirokiterashima (approved):
LGTM. Remove the empty add-step-button.component.scss file and the styleUrls line in AddStepButtonComponent before merge?

PR #1715, opened 2024-04-04
refactor(Tag): Create project tag service

geoffreykwan:
## Changes
- Create project tag service
- Move project tag service code from tag service to project tag service

## Test
- Make sure the basic project tag functionality still works
  - apply tag
  - edit tag
  - create tag


hirokiterashima (approved):
LGTM. 

We use both "user tag" (retrieveUserTags(), /api/user/tags) and "project tag" (/api/projects/tag) in ProjectTagService. Are these referring to the same tag? We might want to rename/refactor these in a future task to clarify.

PR #1716, opened 2024-04-04
feat(Tag): Use tag objects instead of strings in project

geoffreykwan:
## Changes
- Changed project tags to be objects instead of just the string text
- Sort tags alphabetically
- Update tags on run cards when a tag is applied or edited
- Archiving or restoring a run now receives the archived tag object

## Test
- Test with https://github.com/WISE-Community/WISE-API/pull/264
- Make sure tags show up properly on run cards
- Apply a tag to a run and make sure it immediately appears on the run card
- Edit a tag that is applied to a run and make sure it immediately updates on the run card
- Make sure archiving or restoring a run properly shows or removes the archived tag

### DIFF
--- a/src/app/teacher/teacher-run-list-item/teacher-run-list-item.component.ts
+++ b/src/app/teacher/teacher-run-list-item/teacher-run-list-item.component.ts
@@ -58,6 +58,7 @@ export class TeacherRunListItemComponent implements OnInit {
         const tagOnProject = this.run.project.tags.find((tag: Tag) => tag.id === tagThatChanged.id);
         if (tagOnProject != null) {
           tagOnProject.text = tagThatChanged.text;
+          this.projectTagService.sortTags(this.run.project.tags);
         }
       })
     );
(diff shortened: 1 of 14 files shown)


src/app/teacher/teacher-run-list-item/teacher-run-list-item.component.ts
```diff
@@ -44,18 +49,35 @@ export class TeacherRunListItemComponent implements OnInit {
         this.run.highlighted = false;
       }, 7000);
     }
+    this.subscribeToTagUpdated();
   }
 
-  ngAfterViewInit() {
+  subscribeToTagUpdated(): void {
```
hirokiterashima: private?

src/app/teacher/teacher-run-list-item/teacher-run-list-item.component.ts
```diff
@@ -44,18 +49,35 @@ export class TeacherRunListItemComponent implements OnInit {
...
+    this.subscriptions.add(
+      this.projectTagService.tagUpdated$.subscribe((tagThatChanged: Tag) => {
+        const tagOnProject = this.run.project.tags.find((tag: Tag) => tag.id === tagThatChanged.id);
+        if (tagOnProject != null) {
```
hirokiterashima, line 59: When is this null check necessary?

hirokiterashima (approved):
LGTM. 

I added some comments inline. Also, remember to generate messages.

src/app/teacher/teacher-run-list-item/teacher-run-list-item.component.ts
geoffreykwan: If the run/project does not have the tag that was updated, the tagOnProject will be null.

PR #1717, opened 2024-04-05
feat(Microsoft Login): Register and Login Student

hirokiterashima:
## Notes
- This PR introduces a lot of code duplication with Google authentication and teacher registration, but you can ignore them for now. I'll address them in a separate PR to reduce the review scope of this PR.
- Don't worry about styles (e.g. MS button looks like Google button) for now. We'll also address that in a separate PR.

## Changes
- Add Register and Login with Microsoft features for Students

## Test Prep
- Test with https://github.com/WISE-Community/WISE-API/pull/265

## Test (for student accounts)
- Register/Sign in with Microsoft buttons
   - appear when properties are set in wise-dockerdev.properties.
   - do NOT appear when properties are NOT set in wise-dockerdev.properties.
- Register with Microsoft
   - with new MS account => lets you create new student account 
   - with existing MS account => shows page with message "Microsoft user already has a WISE account"
- Login with Microsoft
   - with already-registered MS account => works
   - with unregistered MS account => shows page with message "Sorry, your Microsoft User ID was not found in WISE."

### DIFF
--- a/src/app/register/register-student-complete/register-student-complete.component.html
+++ b/src/app/register/register-student-complete/register-student-complete.component.html
@@ -1,15 +1,15 @@
 <mat-card class="standalone__content standalone__content--sm mat-elevation-z4 center">
   <mat-card-content>
     <h2 class="standalone__title accent" i18n>Your WISE account has been created!</h2>
-    <p *ngIf="!isUsingGoogleId" class="mat-subtitle-2" i18n>
+    <p *ngIf="!socialAccount" class="mat-subtitle-2" i18n>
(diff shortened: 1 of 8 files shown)


geoffreykwan (approved):
Student register and sign in with Microsoft works.

I noticed that if Microsoft login is not enabled in the API application properties, it still shows the "Sign up with Microsoft" button in the student register and teacher register pages. Right now in ```register-student.component.ts``` and ```register-teacher.component.ts``` it checks config.microsoftClientId != null but it looks like we should also check for empty string or just have the config set microsoftClientId to be null on the API side.

PR #1718, opened 2024-04-05
feat(Tag): Implement delete tag

geoffreykwan:
## Changes
- In the Manage Tags dialog, added delete button next to each tag

## Test
- Test with https://github.com/WISE-Community/WISE-API/pull/266
- Go to the teacher run list
- Click the checkbox for a run
- Click "Manage Tags"
- Click on a delete button next to a tag
- A confirm message should pop up saying how many units use that tag
- Click "OK"
- The tag should immediately disappear from the "Manage Tags" popup
- The tag should immediately disappear from the "Apply Tags" menu
- If the tag was on any runs, it should immediately disappear from the run card

### DIFF
--- a/src/app/teacher/manage-tags-dialog/manage-tags-dialog.component.ts
+++ b/src/app/teacher/manage-tags-dialog/manage-tags-dialog.component.ts
@@ -91,18 +91,13 @@ export class ManageTagsDialogComponent implements OnInit {
   }
 
   private getDeleteMessage(runs: Run[], tag: Tag): string {
-    const numberOfProjectsWithTag = this.getNumberOfProjectsWithTag(runs, tag);
-    const numberOfProjectsMessage =
-      numberOfProjectsWithTag === 1
-        ? $localize`There is ${numberOfProjectsWithTag} unit with this tag.`
-        : $localize`There are ${numberOfProjectsWithTag} units with this tag.`;
-    return $localize`Are you sure you want to delete this tag? ` + numberOfProjectsMessage;
-  }
-
-  private getNumberOfProjectsWithTag(runs: Run[], tag: Tag) {
-    const projectsWithTag = runs.filter((run: Run) =>
+    const numProjectsWithTag = runs.filter((run: Run) =>
       run.project.tags.some((projectTag: Tag) => projectTag.id === tag.id)
-    );
-    return projectsWithTag.length;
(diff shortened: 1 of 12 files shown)


src/app/teacher/manage-tags-dialog/manage-tags-dialog.component.ts
```diff
@@ -70,4 +81,28 @@ export class ManageTagsDialogComponent implements OnInit {
...
+  protected delete(tag: Tag): void {
+    this.teacherService.getRuns().subscribe((runs) => {
+      if (confirm(this.getDeleteMessage(runs, tag))) {
+        this.projectTagService.deleteTag(tag);
+      }
+    });
+  }
+
+  private getDeleteMessage(runs: Run[], tag: Tag): string {
+    const numberOfProjectsWithTag = this.getNumberOfProjectsWithTag(runs, tag);
```
hirokiterashima: Shorten variable and function names? Same for ```numProjectsMessage```
```suggestion
    const numProjectsWithTag = this.getNumProjectsWithTag(runs, tag);
```

src/app/teacher/manage-tags-dialog/manage-tags-dialog.component.ts
```diff
@@ -70,4 +81,28 @@ export class ManageTagsDialogComponent implements OnInit {
...
+        ? $localize`There is ${numberOfProjectsWithTag} unit with this tag.`
+        : $localize`There are ${numberOfProjectsWithTag} units with this tag.`;
+    return $localize`Are you sure you want to delete this tag? ` + numberOfProjectsMessage;
+  }
+
+  private getNumberOfProjectsWithTag(runs: Run[], tag: Tag) {
+    const projectsWithTag = runs.filter((run: Run) =>
+      run.project.tags.some((projectTag: Tag) => projectTag.id === tag.id)
+    );
+    return projectsWithTag.length;
```
hirokiterashima: I don't think we need a variable here. We may not even need this function since it's only called in one place?
```suggestion
    return runs.filter((run: Run) =>
      run.project.tags.some((projectTag: Tag) => projectTag.id === tag.id)
    ).length;
```

hirokiterashima (approved):
LGTM. We can have @breity style this in a separate PR.

I added some code comments inline.

PR #1719, opened 2024-04-05
refactor(Register User): Extract AbstractRegisterUserComponent

hirokiterashima:
## Changes
- Create AbstractRegisterUserComponent that houses common code in Register[Student/Teacher]Component
- Rename ```isGoogleAuthenticationEnabled``` to ```googleAuthenticationEnabled```

## Test Prep
- Test with ```login-with-microsoft``` branch on API

## Test
- Teacher and Student user registration works as before
   - without social account
   - with Microsoft
   - with Google

### DIFF
--- a/src/app/register/abstract-register-user.component.ts
+++ b/src/app/register/abstract-register-user.component.ts
@@ -1,0 +1,53 @@
+import { Router } from '@angular/router';
+import { UserService } from '../services/user.service';
+import { ConfigService } from '../services/config.service';
+import { Directive, OnInit } from '@angular/core';
+import { GoogleUser } from '../modules/google-sign-in/GoogleUser';
+
+@Directive()
+export abstract class AbstractRegisterUserComponent implements OnInit {
+  protected googleAuthenticationEnabled: boolean = false;
+  protected microsoftAuthenticationEnabled: boolean;
+  protected abstract joinFormPath: string;
+
+  constructor(
+    private configService: ConfigService,
+    private router: Router,
+    private userService: UserService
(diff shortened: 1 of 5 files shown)


src/app/register/abstract-register-user.component.ts
```diff
@@ -0,0 +1,53 @@
...
+import { UserService } from '../services/user.service';
+import { ConfigService } from '../services/config.service';
+import { Directive, OnInit } from '@angular/core';
+import { GoogleUser } from '../modules/google-sign-in/GoogleUser';
+
+@Directive()
+export abstract class AbstractRegisterUserComponent implements OnInit {
+  protected googleAuthenticationEnabled: boolean = false;
+  protected microsoftAuthenticationEnabled: boolean;
+  protected abstract joinFormPath: string;
```
geoffreykwan: This can be ordered alphabetically.

```microsoftAuthenticationEnabled``` can be initialized to false to be consistent with ```googleAuthenticationEnabled``` being initialized to false.

geoffreykwan (approved):
Looks good.

src/app/register/abstract-register-user.component.ts
hirokiterashima: Something we can discuss as a team: everything would continue to work even when microsoftAuthenticationEnabled and googleAuthenticationEnabled aren't initialized (to false, in this case). It would default to ```undefined```, which is ```falsy```. We do this in many parts of our code, so it'd be good to establish a convention.

PR #1722, opened 2024-04-08
feat(Tag): Prevent creating a duplicate tag

geoffreykwan:
## Changes
- When creating a new tag, make sure it is not a duplicate

## Test
- Test with https://github.com/WISE-Community/WISE-API/pull/268
- Go to the teacher run list
- Click the checkbox for a run
- Click "Manage Tags"
- Click "New Tag"
- Type in a tag name that already exists
- You should not be able to click the "Create" button and you should see an error message that says "Tag already exists"
- Type in a tag name that does not exist
- Click the "Create" button
- The new tag should be created

### DIFF
--- a/src/app/teacher/create-tag-dialog/create-tag-dialog.component.ts
+++ b/src/app/teacher/create-tag-dialog/create-tag-dialog.component.ts
@@ -1,6 +1,14 @@
 import { CommonModule } from '@angular/common';
 import { Component, OnInit } from '@angular/core';
-import { FormsModule } from '@angular/forms';
+import {
+  AbstractControl,
+  FormControl,
+  FormsModule,
+  ReactiveFormsModule,
+  ValidationErrors,
+  ValidatorFn,
+  Validators
+} from '@angular/forms';
 import { MatButtonModule } from '@angular/material/button';
 import { MatDialogModule, MatDialogRef } from '@angular/material/dialog';
 import { MatFormFieldModule } from '@angular/material/form-field';
@@ -8,6 +16,7 @@ import { MatInputModule } from '@angular/material/input';
 import { MatSnackBar } from '@angular/material/snack-bar';
 import { Subscription } from 'rxjs';
 import { ProjectTagService } from '../../../assets/wise5/services/projectTagService';
+import { Tag } from '../../domain/tag';
 
 @Component({
   selector: 'create-tag-dialog',
@@ -20,12 +29,17 @@ import { ProjectTagService } from '../../../assets/wise5/services/projectTagServ
     MatButtonModule,
     MatDialogModule,
     MatFormFieldModule,
-    MatInputModule
+    MatInputModule,
+    ReactiveFormsModule
   ]
 })
 export class CreateTagDialogComponent implements OnInit {
   private subscriptions: Subscription = new Subscription();
-  protected tagName: string = '';
+  protected tagControl = new FormControl('', [
+    Validators.required,
+    this.createUniqueTagValidator()
+  ]);
+  private tags: Tag[] = [];
 
   constructor(
     private dialogRef: MatDialogRef<CreateTagDialogComponent>,
@@ -34,6 +48,9 @@ export class CreateTagDialogComponent implements OnInit {
   ) {}
 
   ngOnInit(): void {
+    this.projectTagService.retrieveUserTags().subscribe((tags: Tag[]) => {
+      this.tags = tags;
+    });
     this.subscriptions.add(
       this.projectTagService.newTag$.subscribe(() => {
         this.snackBar.open($localize`Tag created`);
@@ -45,6 +62,15 @@ export class CreateTagDialogComponent implements OnInit {
     this.subscriptions.unsubscribe();
   }
 
+  createUniqueTagValidator(): ValidatorFn {
+    return (control: AbstractControl): ValidationErrors | null => {
+      return this.tags != null &&
+        this.projectTagService.doesTagAlreadyExist(this.tags, control.value)
+        ? { tagAlreadyExists: true }
+        : null;
+    };
+  }
+
   protected keyPressed(event: KeyboardEvent): void {
     if (event.key === 'Enter') {
       this.create();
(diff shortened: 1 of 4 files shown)


src/app/teacher/create-tag-dialog/create-tag-dialog.component.ts
```diff
@@ -45,14 +62,32 @@ export class CreateTagDialogComponent implements OnInit {
     this.subscriptions.unsubscribe();
   }
 
+  createUniqueTagValidator(): ValidatorFn {
```
hirokiterashima: private?

src/assets/wise5/services/projectTagService.ts
```diff
@@ -38,10 +38,12 @@ export class ProjectTagService {
...
-    return this.http.post(`/api/user/tag`, { text: tagName }).subscribe((tag: Tag) => {
-      this.newTagSource.next(tag);
-    });
+  createTag(tagName: string): Observable<Object> {
+    return this.http.post(`/api/user/tag`, { text: tagName });
+  }
+
+  emitNewTag(tag: Tag): void {
+    this.newTagSource.next(tag);
   }
```
hirokiterashima: I'm guessing that we always want to emit the newTagSource event after each call to post() returns. If you pipe the result to post(), you can emit the newTagSource event and not need to define emitNewTag() function?
```suggestion
  createTag(tagName: string): Observable<Tag> {
    return this.http
      .post<Tag>(`/api/user/tag`, { text: tagName })
      .pipe(
        tap((tag) => {
          this.newTagSource.next(tag);
        })
      );
  }
```

hirokiterashima (commented):
Works well. I added some code comments inline. Lmk what you think.

geoffreykwan:
Ok, I've made the changes.

src/app/teacher/create-tag-dialog/create-tag-dialog.component.ts
```diff
@@ -45,14 +62,31 @@ export class CreateTagDialogComponent implements OnInit {
     this.subscriptions.unsubscribe();
   }
 
+  private createUniqueTagValidator(): ValidatorFn {
+    return (control: AbstractControl): ValidationErrors | null => {
+      return this.tags != null &&
```
hirokiterashima, line 67: Do we need this null check if we've already initialized it to an empty array?
```suggestion
      return 
```

src/app/teacher/create-tag-dialog/create-tag-dialog.component.ts
```diff
@@ -45,14 +62,31 @@ export class CreateTagDialogComponent implements OnInit {
...
+        this.projectTagService.doesTagAlreadyExist(this.tags, control.value)
```
hirokiterashima, line 68: Do we need this function to be defined in ProjectTagService? Since it's only used in this component, we can just define it in this same class?
```suggestion
        this.doesTagAlreadyExist(control.value)
```


Private function in CreateTagDialogComponent (untested):
```
  private doesTagAlreadyExist(tagText: string): boolean {
    return this.tags.some((tag: Tag) => tag.text.toLowerCase().trim() === tagText.toLowerCase().trim());
  }
```

Also, do we need to trim the existing tags? Would they already be trimmed?

src/app/teacher/create-tag-dialog/create-tag-dialog.component.ts
```diff
@@ -45,14 +62,31 @@ export class CreateTagDialogComponent implements OnInit {
...
-    this.dialogRef.close();
+    this.projectTagService.createTag(this.tagControl.value.trim()).subscribe({
+      next: () => {
+        this.dialogRef.close();
+      },
+      error: ({ error }) => {
+        if (error.messageCode === 'tagAlreadyExists') {
+          this.tagControl.setErrors({ tagAlreadyExists: true });
+        }
+      }
```
hirokiterashima, line 89: When does this code run? I don't see the server returning any error?

hirokiterashima (commented):
New changes look good. I found few more areas that might be improved.

src/assets/wise5/services/projectTagService.ts
```diff
@@ -38,10 +38,12 @@ export class ProjectTagService {
     });
   }
 
-  createTag(tagName: string): Subscription {
-    return this.http.post(`/api/user/tag`, { text: tagName }).subscribe((tag: Tag) => {
-      this.newTagSource.next(tag);
-    });
+  createTag(tagName: string): Observable<Object> {
```
hirokiterashima, line 41: This should return a Tag Observable?
```suggestion
  createTag(tagName: string): Observable<Tag> {
```

src/app/teacher/create-tag-dialog/create-tag-dialog.component.ts
geoffreykwan: This error is sent from the server here

https://github.com/WISE-Community/WISE-API/blob/25d19539b1c5475020733177edcd03d7b48dfac3/src/main/java/org/wise/portal/presentation/web/controllers/tag/UserTagController.java#L43

To trigger it
- Open the create new tag dialog in two different tabs
- In the first tab enter some text and click "Create"
- In the second tab enter the same text and click "Create"
- You should then see the error "Tag already exists"

geoffreykwan:
I'll fix those other issues in the next PR since I've already started working on adding the edit tag dialog and creating a parent abstract tag dialog which moves the code you pointed out.

src/app/teacher/create-tag-dialog/create-tag-dialog.component.ts
hirokiterashima: Got it. Must have missed it.

hirokiterashima:
Ok sounds good.

hirokiterashima (approved):

src/app/teacher/create-tag-dialog/create-tag-dialog.component.ts
geoffreykwan: I added the null check because for some reason the validator is called before this.tags is initialized.

PR #1725, opened 2024-04-10
feat(Translation): Convert milestone authoring view to translatable

geoffreykwan:
## Changes
- Convert milestone authoring view to translatable

## Test
- In the milestone authoring view make sure these fields are translatable
  - Milestone Name
  - Milestone Description
  - Template Description
  - Template Content
  - Template Recommendations


breity (approved):
Looks good. 👍 

I notice that sometimes we use self-closing tags for translatable inputs and text areas:
```
<translatable-input
...
/>
```
But in this PR, we don't. I think we should settle on a standard before the feature branch is merged.

Long content in the milestone reports or recommended actions result in the source language display overlapping content below:
Screenshot 2024-04-11 at 1 59 07 PM

As we discussed, we should move away from using <mat-hint> elements for the source language text in translation mode. Will do in another PR.

PR #1727, opened 2024-04-11
feat(Tag): Implement tag color

geoffreykwan:
## Changes
- Added ability to set color on a tag
- Created EditTagComponent
- Deleted AbstractTagDialogComponent
- Fixed a bug where "Tag already exists" error message does not show up immediately when it should
- Fixed a bug where the user could bypass tag validation checks by hitting enter

## Test
- Test with https://github.com/WISE-Community/WISE-API/pull/270
- Make sure 
  - you can create a tag with color
  - you can update a tag with color
  - tag color shows up in the manage tags dialog
  - tag color shows up in the apply tags menu
  - tag color shows up in the run card
  - tag color immediately updates in the manage tags dialog
  - tag color immediately updates in the apply tags menu
  - tag color immediately updates in the run card


hirokiterashima (commented):
Works as described. I have some comments.

What's the purpose for having the endpoints return ProjectAndTagsResponse? It doesn't look like this response is used in any of the components that call the endpoints. Should the endpoints return void instead?

CreateTagDialogComponent and EditTagDialogComponent are very similar. Should we extract common code to an abstract class?

Manage Tags UI/UX has veered a bit from our designs (https://balsamiq.cloud/s8brvqj/p9j1not/rEB79 and https://balsamiq.cloud/s8brvqj/p9j1not/r24F3). Should we allow editing (tag name, tag color) directly in the Manage Tags dialog?

Also, I don't think we should have the author type in a color because we don't want users typing in dark colors like "black" and "blue", which would make the tag text difficult to read. How difficult is it to add a color chooser, like a grid of pre-selected soft colors?

hirokiterashima (approved):
LGTM. We'll make improvements on UI/UX and code in the future PR's.

PR #1728, opened 2024-04-11
feat(Translation): Make embedded authoring inputs translatable

hirokiterashima:
## Notes
- Please style as you see fit. The url input and the asset choose button used to be on the same line (see production), but I couldn't figure out how to do that easily with the new elements.
- When you are translating to another language, the changes are saved, but the input field does not immediately update with the newly-selected asset at the moment. It takes a second or so to update.

## Changes
- Parameterize TranslatableAssetChooser button tooltip
- Convert embedded authoring
   - url input to TranslatableInput
   - asset chooser to TranslatableAssetChooser
   - Note: these changes removed the need to define ```assetSelected()``` and ```chooseModelFile()``` in EmbeddedAuthoringComponent

## Test Prep
- Test with the ```at-units-in-multiple-languages``` branch on API

## Test (In embedded component authoring)
- The tooltip text on the button for choosing model should say "Choose model file"
- Choosing asset in both non-multi-lingual unit and multi-lingual unit work when: 
   - typing in the asset url string in the input
   - using the "Choose model file" asset chooser

### DIFF
--- a/src/assets/wise5/authoringTool/components/translatable-asset-chooser/translatable-asset-chooser.component.html
+++ b/src/assets/wise5/authoringTool/components/translatable-asset-chooser/translatable-asset-chooser.component.html
@@ -2,9 +2,8 @@
   mat-raised-button
   color="primary"
   (click)="chooseAsset()"
-  matTooltip="Choose an image"
+  [matTooltip]="tooltip"
   matTooltipPosition="above"
-  i18n-matTooltip
 >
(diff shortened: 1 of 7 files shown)


breity (approved):
Looks good. 👍 

I update styles so input field and asset choose button are on the same line.

hirokiterashima:
Great. Thanks!

PR #1729, opened 2024-04-11
feat(Translation): Convert notebook authoring to translatable

geoffreykwan:
## Changes
- Convert notebook authoring to translatable
- Deleted a lot of code in notebook-authoring.component.ts because that logic is now handled in translatable-rich-text-editor.component.ts

## Test
- Make sure all the relevant notebook authoring fields are translatable


breity (commented):
Functionality looks good. I made some minor style updates.

I was wondering if it makes sense to make the Teacher Notebook fields translatable. As of now, this doesn't actually do anything because the Teacher Notebook shows up in the Classroom Monitor, not the VLE. So the translated Teacher Notebook fields are never used.

Also, adding translatable labels for the Notebook results in mixed language text in some places. For example:
Screenshot 2024-04-11 at 3 17 10 PM
Screenshot 2024-04-11 at 3 16 41 PM

Not sure the best way to handle these cases...

geoffreykwan:
I've reverted the teacher notebook fields to not be translatable.

breity (approved):
LGTM.

PR #1730, opened 2024-04-12
feat(Tag): Hide archived tag and prevent user from creating archived tag

geoffreykwan:
## Changes
- Hide archived tag in run card, apply tags list, and manage tags dialog
- Prevent user from creating a tag named archived

## Test
- Make sure archived tag does not show up in
  - run card
  - apply tags list
  - manage tags dialog
- Make sure the user can not create a tag named archived
- Make sure the user can not change a tag to be named archived

### DIFF
--- a/src/app/teacher/apply-tags-button/apply-tags-button.component.html
+++ b/src/app/teacher/apply-tags-button/apply-tags-button.component.html
@@ -10,18 +10,16 @@
 <mat-menu #applyTagsMenu>
   <div class="menu-info secondary-text" i18n>Apply tags</div>
   <mat-divider></mat-divider>
-  <ng-container *ngFor="let tag of tags">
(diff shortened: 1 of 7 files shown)


hirokiterashima (commented):
Works as described.

Should ```ProjectTagService.retrieveUserTags()``` exclude the archived tag before returning the array? This would remove the need for checks in parts that use this array. Since "Archived" is a reserved tag, it should not belong in the user tags array?
```
  retrieveUserTags(): Observable<Tag[]> {
    return this.http
      .get<Tag[]>(`/api/user/tags`)
      .pipe(map((tags) => tags.filter((tag) => tag.text != 'archived')));
  }
```

geoffreykwan:
Ah yea good idea. I've made the change.

hirokiterashima (approved):
LGTM

PR #1731, opened 2024-04-14
feat(Export): Create export gradebook route

### LINKED ISSUE #1698: feat(Export): Create export gradebook route
Move the "export one workgroup per row" code out of data-export.component.ts and into a new component/route named "export gradebook".

geoffreykwan:
## Changes
- Created export Gradebook route (this used to be called One Workgroup Per Row, the strategy is still called OneWorkgroupPerRowDataExportStrategy and we might want to rename that in the future)

## Test
- Make sure export Gradebook works

Closes #1698


src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractDataExportStrategy.ts
```diff
@@ -4,15 +4,14 @@ import { ConfigService } from '../../../services/configService';
...
-import { DataExportComponent } from '../data-export/data-export.component';
 import { DataExportContext } from '../DataExportContext';
 import { DataExportStrategy } from './DataExportStrategy';
 import { removeHTMLTags } from '../../../common/string/string';
 import { generateCSVFile } from '../../../common/csv/csv';
 
 export abstract class AbstractDataExportStrategy implements DataExportStrategy {
   context: DataExportContext;
-  controller: DataExportComponent;
+  controller: any;
```
hirokiterashima, line 14: Why are we relaxing the typing of the controller from DataExportComponent to any here?

geoffreykwan: This is due to the tight coupling between the strategies and the controller which I don't think should have been implemented like this in the first place. More refactoring in the export strategies will need to be done in the future to reduce the coupling.

hirokiterashima: I see. In this case, should we use a [union type](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#union-types) so we don't allow any type?

hirokiterashima (commented):
I'm having a hard time following the changes in this PR, because there are multiple changes going on at once and it's hard to wrap my head around them:
1. Renaming of OneWorkgroupPerRow feature and route to Gradebook, but still keeping the OneWorkgroupPerRowDataExportStrategy
2. Relaxing AbstractDataExportStrategy.controller type from ```DataExportComponent``` to ```any```
3. Refactoring of DataExportComponent.

Do they all need to happen at once, or can this be broken down into stages to make the changes easier for me to follow?

I'm also wondering why we're renaming the route and export to "grade book". The contents of the export didn’t quite match what I expect to see when I think of a grade book. When I think of a grade book, I just expect to see columns of scores, and I don't expect to see student work. I wonder if the researchers and teachers will have the same expectation? I think OneWorkgroupPerRow was a better name for this export.

geoffreykwan:
Ok, I've renamed Export Gradebook back to One Workgroup Per Row.

hirokiterashima (approved):
LGTM.

It looks like we don't show the Export button at the top anymore in these new routes, so you have to scroll to the bottom to find the Export button. This can be cumbersome for when choosing to export specific items for a large unit. Let's discuss a better design for these new routes soon.

PR #1732, opened 2024-04-15
feat(AI Chat): Check if chat gpt is enabled when adding items

geoffreykwan:
## Changes
- Check authoring config param "chatGptEnabled" to determine whether to show the AI Chat item when an author adds items

## Test
- Test with https://github.com/WISE-Community/WISE-API/pull/271
- If chat gpt is enabled, you should be able to add AI Chat items in the authoring
- If chat gpt is not enabled, you should not be able to add AI Chat items in the authoring


hirokiterashima (approved):
LGTM

PR #1734, opened 2024-04-15
chore(Angular): Remove ng2-file-upload dependency

hirokiterashima:
## Changes
- Remove ng2-file-upload dependency. This library was blocking us from upgrading to Angular 17
- Update ProjectAssetAuthoring to use DragAndDropDirective instead for uploading files. We use this directive for uploading assets to student notebooks

## Test
- In Authoring > Project Assets view, make sure that uploading files via click and drag/drop work as before


geoffreykwan (approved):
Looks good.

PR #1736, opened 2024-04-16
fix(Tag): Show indeterminate checkbox when only some projects have a tag

geoffreykwan:
## Changes
- In the apply tags menu, if only some of the selected units have a tag, show the indeterminate checkbox (dash instead of check mark)

## Test
- Select some units that have different tags
- Make sure
  - The tags that are not on any of the selected units are not checked
  - The tags that are on some of the selected units are indeterminate checked (dash)
  - The tags that are on all the selected units are checked (check mark)


src/app/teacher/apply-tags-button/apply-tags-button.component.ts
```diff
@@ -19,21 +21,37 @@ export class ApplyTagsButtonComponent implements OnInit {
...
   ngOnInit(): void {
-    this.retrieveUserTags();
     this.subscribeToTagUpdated();
     this.subscribeToNewTag();
     this.subscribeToTagDeleted();
   }
 
+  ngOnChanges(): void {
+    this.retrieveUserTags();
+  }
```
hirokiterashima: Why do we need to retrieve the user tags each time selectedProjects changes? Shouldn't the tags stay the same?

src/app/teacher/apply-tags-button/apply-tags-button.component.ts
```diff
@@ -19,21 +21,37 @@ export class ApplyTagsButtonComponent implements OnInit {
...
+    tag.checked = numProjectsWithTag === this.selectedProjects.length;
+    tag.indeterminateChecked =
+      numProjectsWithTag > 0 && numProjectsWithTag < this.selectedProjects.length;
+  }
+
+  private getNumProjectsWithTag(tag: Tag): number {
+    return this.selectedProjects.filter((project) =>
+      project.tags.some((projectTag) => projectTag.id === tag.id)
+    ).length;
+  }
```
hirokiterashima: Do we need to define a function here, or can we just call ```this.selectedProjects.filter(...)``` inline in updateTagCheckedValue()?

hirokiterashima (commented):
Works as described. I added some comments inline.

It seems like ```SelectAllItemsCheckboxComponent``` could be used here for showing checked/indetermined/none status on the checkbox? What do you think?

src/app/teacher/apply-tags-button/apply-tags-button.component.ts
geoffreykwan: If you modify the tags in a different tab it can change, but that's not very likely to happen so I've made the change.

src/app/teacher/apply-tags-button/apply-tags-button.component.ts
geoffreykwan: Ok, I've made this change.

geoffreykwan:
Using `SelectAllItemsCheckboxComponent` would require some change on `SelectAllItemsCheckboxComponent` to be able to pass in a value to be placed inside the mat-checkbox and to modify the tooltip.

geoffreykwan:
Apply tags now uses `SelectAllItemsCheckboxComponent`.

hirokiterashima (approved):
Looks good. 

Can SelectAllItemsCheckboxComponent be a standalone component and remove the need for SelectAllItemsCheckboxModule? If yes, can you change it before you merge?

geoffreykwan:
Ok, I've converted SelectAllItemsCheckboxComponent to standalone.

PR #1740, opened 2024-04-17
chore(Application): Upgrade to Angular v17

hirokiterashima:
## Changes
- Upgrade Angular version from 16 to 17
   -  Revert to using original ```angular-password-strength-meter``` dependency (before: we built and deployed our own ```@wise-community/angular-password-strength-meter```
   - Change instances where we were calling ```mutate()``` on signals, as this is no longer supported in Angular 17. Instead, we call ```update()```. If it's an object or array that we're updating, we need to make sure that we're returning a new reference for the change to propagate.
   - Update tests

## Test
- Everything works as before, in particular:
   - Password strength meter works as before in creating new account and updating existing account's password
   - Selecting/unselecting single/multiple/none units in Unit Library
   - Archiving/unarchiving unit
   - Authoring tool 
      - expand all/collapse all lessons and components
      - selecting steps/components and moving/deleting/copying them


src/app/modules/library/personal-library/personal-library.component.ts
```diff
@@ -7,6 +7,7 @@ import { ProjectFilterValues } from '../../../domain/projectFilterValues';
 import { ArchiveProjectService } from '../../../services/archive-project.service';
 import { PageEvent } from '@angular/material/paginator';
 import { ProjectSelectionEvent } from '../../../domain/projectSelectionEvent';
+import { copy } from '../../../../assets/wise5/common/object/object';
```
geoffreykwan: Unused import can be deleted.

geoffreykwan (changes requested):
I came upon this weird behavior in the VLE.

1. Go to the Authoring Tool to author a unit
2. Create a step that contains all item types
3. Create another step after it that contains one item like Open Response
4. Preview the unit
5. Go to the first step
6. Go to the second step and it will show a bunch of items when it should only show the Open Response item

src/app/modules/library/personal-library/personal-library.component.ts
hirokiterashima: done

hirokiterashima:
@geoffreykwan Good catch. This was due to an Angular Flex Layout issue. In order to work with Angular 17, we need to declare the class before we can add/remove it to an element. I fixed this in the discussion component and the About Us page.

geoffreykwan (approved):
Looks good.

breity (approved):
LGTM.

PR #1742, opened 2024-04-19
feat(Tag): Add color chooser when editing tag

geoffreykwan:
## Changes
- Add color chooser when editing a tag
- Added preview of color and text

## Test
- Make sure choosing a color using the color picker updates the color value

### DIFF
--- a/src/app/teacher/color-chooser/color-chooser.component.ts
+++ b/src/app/teacher/color-chooser/color-chooser.component.ts
@@ -1,0 +1,29 @@
+import { CommonModule } from '@angular/common';
+import { Component, EventEmitter, Input, Output } from '@angular/core';
+import { MatChipsModule } from '@angular/material/chips';
+
+@Component({
+  selector: 'color-chooser',
+  templateUrl: './color-chooser.component.html',
+  styleUrls: ['./color-chooser.component.scss'],
+  standalone: true,
+  imports: [CommonModule, MatChipsModule]
+})
+export class ColorChooserComponent {
+  @Output() chooseColorEvent: EventEmitter<string> = new EventEmitter();
+  @Input() chosenColor: string;
+  protected colors: string[] = [
+    '#66BB6A',
+    '#009688',
+    '#00B0FF',
+    '#1565C0',
+    '#673AB7',
+    '#AB47BC',
+    '#E91E63',
+    '#D50000',
+    '#F57C00',
+    '#FBC02D',
+    '#795548',
+    '#757575'
+  ];
+}
--- a/src/app/teacher/edit-tag/edit-tag.component.spec.ts
+++ b/src/app/teacher/edit-tag/edit-tag.component.spec.ts
@@ -4,7 +4,6 @@ import { ProjectTagService } from '../../../assets/wise5/services/projectTagServ
 import { HttpClientTestingModule } from '@angular/common/http/testing';
 import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
 import { MatSnackBarModule } from '@angular/material/snack-bar';
-import { MatChipsModule } from '@angular/material/chips';
 
 describe('EditTagComponent', () => {
   let component: EditTagComponent;
@@ -16,7 +15,6 @@ describe('EditTagComponent', () => {
         BrowserAnimationsModule,
         EditTagComponent,
         HttpClientTestingModule,
-        MatChipsModule,
         MatSnackBarModule
       ],
       providers: [ProjectTagService]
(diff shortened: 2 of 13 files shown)


hirokiterashima (commented):
Functionality works. I have a few comments:
- Should the color picker be its own component instead of being part of the EditTagComponent?
- The checkmark doesn't appear centered in the color box (seems slightly to the right of the center)
- Do we need to show the color text input and let user specify their own color, or is 12 default colors to choose from enough?
- Instead of the text "preview", should it show the actual text of the tag? (see GitHub label editing view)

geoffreykwan:
I've created the color picker component and also now show the text and color preview.

src/app/teacher/edit-tag/edit-tag.component.ts
```diff
@@ -16,17 +16,21 @@ import { Tag } from '../../domain/tag';
...
   templateUrl: './edit-tag.component.html',
   styleUrls: ['./edit-tag.component.scss'],
   standalone: true,
   imports: [
+    ColorChooserComponent,
     CommonModule,
     FormsModule,
     FlexLayoutModule,
     MatButtonModule,
+    MatChipsModule,
```
hirokiterashima: Remove?

src/app/teacher/edit-tag/edit-tag.component.ts
```diff
@@ -16,17 +16,21 @@ import { Tag } from '../../domain/tag';
 import { FlexLayoutModule } from '@angular/flex-layout';
 import { MatButtonModule } from '@angular/material/button';
 import { MatSnackBar } from '@angular/material/snack-bar';
+import { MatChipsModule } from '@angular/material/chips';
```
hirokiterashima: Remove?

src/app/teacher/edit-tag/edit-tag.component.ts
```diff
@@ -99,6 +103,10 @@ export class EditTagComponent {
     }
   }
 
+  protected chooseColor(event: any): void {
```
hirokiterashima: Is there a type for this event that is more specific? Is it MatChipSelectionChange? https://material.angular.io/components/chips/api#MatChipEditedEvent

src/app/teacher/edit-tag/edit-tag.component.spec.ts
```diff
@@ -4,6 +4,7 @@ import { ProjectTagService } from '../../../assets/wise5/services/projectTagServ
 import { HttpClientTestingModule } from '@angular/common/http/testing';
 import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
 import { MatSnackBarModule } from '@angular/material/snack-bar';
+import { MatChipsModule } from '@angular/material/chips';
```
hirokiterashima: Remove?

src/app/teacher/edit-tag/edit-tag.component.spec.ts
```diff
@@ -15,6 +16,7 @@ describe('EditTagComponent', () => {
         BrowserAnimationsModule,
         EditTagComponent,
         HttpClientTestingModule,
+        MatChipsModule,
```
hirokiterashima: Remove?

src/app/teacher/color-chooser/color-chooser.component.ts
```diff
@@ -0,0 +1,29 @@
...
+
+@Component({
+  selector: 'color-chooser',
+  templateUrl: './color-chooser.component.html',
+  styleUrls: ['./color-chooser.component.scss'],
+  standalone: true,
+  imports: [CommonModule, MatChipsModule]
+})
+export class ColorChooserComponent {
+  @Output() chooseColorEvent: EventEmitter<string> = new EventEmitter();
```
hirokiterashima, line 13: Does this emit a string or some sort of an event? Is it MatChipSelectionChange? https://material.angular.io/components/chips/api#MatChipEditedEvent

hirokiterashima (commented):
The new preview looks good.

I added some comments inline.

Also- ColorChooserComponent seems to emit MatChipSelectionChange event (https://material.angular.io/components/chips/api#MatChipEditedEvent)? We then use it to check for selected and then set the color in EditTagComponent.chooseColor(). I'm wondering if ColorChooserComponent should always emit a string representing the color that was chosen from the component, and emit an empty string if the user doesn't want to use any color (we'd have to add this no-color option). I think this would make the abstraction between the two components a bit better. Ideally, EditTagComponent wouldn't have to deal with MatChipSelectionChange event; only deal with the string that represent the color that was chosen. What do you think?

geoffreykwan:
Ok, I've made the color chooser now send a color string.

hirokiterashima (approved):
LGTM.

This template section:
```
<hr *ngIf="(first && !showCreateTag) || (!first && !idToEditing[tags[tagIndex - 1].id])" />
```
is a bit hard to follow. Let's address this in a separate refactor. Maybe we can use borders around the tags instead of hr's to organize the layout.

I'm still not sold on showing the color input. I think it adds complexity to the page layout and the code, and probably won't be used by that many people, so it's not worth the effort. Let's revisit this before merging this feature to develop.

PR #1743, opened 2024-04-19
fix(Tag): Fix new tag checked value

geoffreykwan:
## Changes
- When a new tag is created, it used to show up as checked in the apply tags menu but now it should not

## Test
- Create a new tag
- Make sure the new tag in the apply tags menu is not checked (it used to be checked)


hirokiterashima (approved):
LGTM

PR #1748, opened 2024-04-22
fix(Application): Retrieve config during app initialization

### LINKED ISSUE #1747: bug(Application): Config is not always retrieved in time during initialization
When the server (API) is slow to respond (for example, due to latency), the config object is not retrieved in time for processing by parts that need it.

You can see this by adding a breakpoint in UserAPIController.getConfig() function and loading the login page. You'll see this error:
```
ERROR TypeError: Cannot read properties of null (reading 'recaptchaPublicKey')
    at L.getRecaptchaPublicKey (config.service.ts:60:25)
    at Object.useFactory (app.module.ts:97:30)
```

Modify the startup behavior to ensure that the config object is retrieved during the APP_INITIALIZATION phase before proceeding.

hirokiterashima:
## Changes
- Use promise and resolve() to ensure that the config object is retrieved during APP_INITIALIZATION. This was causing issues in pages that required the config object to be present but the API was slow to respond due to network latency.

## Test
1. Add a breakpoint at the top of UserAPIController.getConfig() function and start a debugging session on API
2. Go directly to the login page by copying&pasting this url to the browser: http://localhost/login
   - This should break at the UserAPIController.getConfig() breakpoint
   - You should not see any js errors
3. Release the breakpoint
   - The login page should load 
   - You should not see any js errors

Test that directly going to other pages work as before, like register student/teacher form pages.

Closes #1747

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -30,9 +30,13 @@ export function initialize(
   userService: UserService
 ): () => Promise<any> {
   return (): Promise<any> => {
-    return userService.retrieveUserPromise().then((user) => {
-      userService.getUser().subscribe((user) => {
-        configService.retrieveConfig();
+    return new Promise((resolve) => {
+      userService.retrieveUserPromise().then(() => {
+        return userService.getUser().subscribe(() => {
+          return configService.retrieveConfig().subscribe((config) => {
+            resolve(config);
+          });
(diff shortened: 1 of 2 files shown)


geoffreykwan (approved):
Looks good.

PR #1749, opened 2024-04-23
feat(Authoring): Add branch show branch path parameters

geoffreykwan:
## Changes
- When "How to Choose Among Paths" is set to "Score" or "Choice Chosen" we now show inputs for each path to allow the author to enter the score or choice id

Note: the "Choice Chosen" input fields will be converted to a select drop down in the future

## Test
- When "Score" is chosen, make sure a number input shows up for each path
- When "Choice Chosen" is chosen, make sure an input shows up for each path
- If you change item, the path inputs are cleared
- If "How to Choose Among Paths" is set to "Choice Chosen" and you choose a "Multiple Choice" item, it will auto populate the choice ids
- If you change "How to Choose Among Paths" to "Choice Chosen", and there is only one "Multiple Choice" item, it will auto populate the choice ids
- Make sure "Submit" is only enabled when everything is filled out

### DIFF
--- a/src/assets/wise5/authoringTool/add-branch/add-branch.component.html
+++ b/src/assets/wise5/authoringTool/add-branch/add-branch.component.html
@@ -62,6 +62,23 @@
         </mat-select>
       </mat-form-field>
     </ng-container>
+    <ng-container *ngIf="formGroup.controls['componentId'].value !== ''">
+      <div *ngFor="let pathFormControl of pathFormControls; index as index">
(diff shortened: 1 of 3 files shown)


hirokiterashima (approved):
Works as described.

AddBranchComponent template and code are getting big and complex. It's becoming difficult to review. Can you refactor them before you add more features to the feature branch? Maybe you can introduce new components to handle the different cases?

PR #1750, opened 2024-04-23
feat(Tag): Add search input to the apply tags menu

geoffreykwan:
## Changes
- Added search input to the apply tags menu
- Added $event.stopPropagation() to the search bar when clicking on the mat-form-field and when typing in the input

## Test
- Make sure the apply tags search filters the tags
- Make sure search still works in
  - Unit library
  - Teacher run list
  - Student run list

### DIFF
--- a/src/app/modules/shared/search-bar/search-bar.component.html
+++ b/src/app/modules/shared/search-bar/search-bar.component.html
@@ -1,12 +1,12 @@
-<mat-form-field class="search-bar" appearance="fill">
+<mat-form-field class="search-bar" appearance="fill" (click)="$event.stopPropagation()">
   <mat-label>{{ placeholderText }}</mat-label>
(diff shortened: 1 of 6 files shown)


hirokiterashima:
Works as described.

I wonder if we should clear the search text every time you open the menu instead of remembering what you typed last time. I was expecting it to be cleared, so I can type a new search text without having to clear it. I don't think I would re-use my previous search text that often, and if I did, I can just re-type it. What do you think?

geoffreykwan:
Ah good catch. It also looks like we'll need to keep the [value] input in SearchBarComponent since it's needed to fix this problem.

hirokiterashima (approved):
LGTM. 

Should menuClosed() be called clearSearchText()? Or, do we need the menuClosed() function- can we just set the searchText to empty in the template?

PR #1753, opened 2024-04-24
feat(Authoring): Add translatable asset choosers

breity:
## Changes
- Add processAsset @input to TranslatableAssetChooserComponent which allows for custom processing of selected asset by component types like Match and MultipleChoice, which wrap an image tag around the filename.
- Replace file upload buttons in Authoring Tool with translatable assets choosers in the following component types:
  - Animation (Image Moving Left, Image Moving Right, Change to Image)
  - Concept Map (Background Image, Node images)
  - Draw (Background Image)
  - Graph (Background Image)
  - Label (Background Image)
  - Match (Choice, Bucket)
  - Multiple Choice (Choice)

## Test
Make sure translatable asset choosers work properly for the listed component fields above.


geoffreykwan (approved):
Looks good.

I noticed that for Concept Map, Draw, and Label, the preview component used to "save starter state" does not show the translated values. We can handle this in a different PR.

breity:
@geoffreykwan Right, I noticed this too. Translatable starter states will take some additional work, so makes sense to address in a different PR. Thanks.

PR #1756, opened 2024-04-25
feat(MultipleLanguages): Delete translations when components are deleted

hirokiterashima:
## Changes
- Delete translations associated with deleted components. Note: to keep the review scope small, I didn't implement deleting translations associated with deleted nodes in this PR.

## Test
- Multilingual unit:
   - Associated translation entries are deleted from translations.[language].json files (for all supported languages) when you delete one component and when you delete multiple components.
- Non-multilingual unit:
   - Deleting component(s) work as before

### DIFF
--- a/src/app/services/removeTranslationsService.spec.ts
+++ b/src/app/services/removeTranslationsService.spec.ts
@@ -34,12 +34,14 @@ function removeComponent() {
       );
       spyOn(configService, 'getProjectId').and.returnValue('123');
       spyOn(configService, 'getConfigParam').and.returnValue('/123/project.json');
-      service.removeComponent({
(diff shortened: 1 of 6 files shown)


geoffreykwan (changes requested):
Deleting one component seems to work but deleting multiple components at the same time does not delete all the translations that should be deleted. If a translation save request is made for each component that is deleted, it leads to a race condition since it saves all the translations in each request.

hirokiterashima:
Good catch. I've changed the service function to handle removing multiple components' translations at at time to prevent overwriting. PTAL

geoffreykwan (approved):
Looks good.

PR #1760, opened 2024-04-26
fix(Authoring): Peer Chat authoring error when peer grouping is null

### LINKED ISSUE #1759: bug(Authoring): Peer Chat authoring error when peer grouping is null
2. Create a step with a Peer Chat item
3. Go into the step
4. Expand the Peer Chat item. The Peer Chat authoring UI will not render properly and you will see this error in the console
```
core.mjs:6531 ERROR TypeError: Cannot read properties of undefined (reading 'name')
    at Tv (select-peer-grouping-authoring.component.html:5:28)
```

I think this error was introduced when we upgraded to Angular 17.

geoffreykwan:
## Changes
- Fixed a null pointer error in Peer Chat authoring when peer grouping is null
- I think this was introduced when upgrading to Angular 17

## Test
2. Create a step with a Peer Chat item
3. Go into the step
4. Expand the Peer Chat item. The Peer Chat authoring UI should render properly. It used to not render properly and would show this error in the console
```
core.mjs:6531 ERROR TypeError: Cannot read properties of undefined (reading 'name')
    at Tv (select-peer-grouping-authoring.component.html:5:28)
```

Closes #1759

### DIFF
--- a/src/assets/wise5/authoringTool/peer-grouping/select-peer-grouping-authoring/select-peer-grouping-authoring.component.html
+++ b/src/assets/wise5/authoringTool/peer-grouping/select-peer-grouping-authoring/select-peer-grouping-authoring.component.html
@@ -1,9 +1,7 @@
 <div fxLayout="row" fxLayoutAlign="start center" fxLayoutGap="10px" class="bottom-spacing">
   <mat-label class="bold" i18n>Grouping Logic:</mat-label>
-  <ng-container [ngSwitch]="peerGrouping">
-    <span *ngSwitchCase="null" i18n>(None Selected)</span>
-    <span *ngSwitchDefault>{{ peerGrouping.name }}</span>
-  </ng-container>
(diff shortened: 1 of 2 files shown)


hirokiterashima (approved):
LGTM. Do you need to generate messages.xlf?

Also, I'm not sure when this feature broke (or did it ever work?) but ```this.peerGroupingAuthoringService.getPeerGrouping(this.tag)``` either returns a PeerGroup instance or ```undefined```. So in the original code, ngSwitch should have been checking for```undefined``` instead of ```null```. It doesn't appear related to upgrade to Angular 17.

geoffreykwan:
The Peer Chat authoring works in 51d0a9c2ee9a4ca26d961ca90c1b5041bab139ce but does not work in the commit after that when we upgraded to Angular 17 6a01e3c8071a46ab5c7bb12a217d22705a992703. It looks like previously the ngSwitchCase would treat undefined the same as null but now it doesn't.

PR #1761, opened 2024-04-26
feat(Tag): Add filter units by tag

geoffreykwan:
## Changes
- Added filter units by tag
- Created AbstractTagsMenuComponent that ApplyTagsButtonComponent and TagsFilterComponent both extend

## Test
- Make sure you can filter units in the teacher run list by tag
- Make sure the apply tags menu still works


src/app/teacher/tags-filter/tags-filter.component.ts
```diff
@@ -0,0 +1,18 @@
...
+import { SharedModule } from '../../modules/shared/shared.module';
+import { Tag } from '../../domain/tag';
+import { AbstractTagsMenuComponent } from '../abstract-tags-menu/abstract-tags-menu.component';
+
+@Component({
+  selector: 'tags-filter',
+  templateUrl: './tags-filter.component.html',
+  styleUrls: ['./tags-filter.component.scss'],
+  standalone: true,
+  imports: [CommonModule, MatDividerModule, SharedModule]
```
hirokiterashima: This component imports SharedModule, but it only needs some of the components and modules within it. Is it possible to only import those dependencies instead?

src/app/teacher/tags-filter/tags-filter.component.ts
```diff
@@ -0,0 +1,18 @@
...
+})
+export class TagsFilterComponent extends AbstractTagsMenuComponent {
```
hirokiterashima: Should this component be called SelectTagsComponent instead? 

TagsFilterComponent => SelectTagsComponent
tagsFilterChanged => selectedTagsEvent

hirokiterashima (commented):
Works as described.

I think some of the class and variable namings can be improved a bit. See inline comments.

src/app/teacher/tags-filter/tags-filter.component.ts
geoffreykwan: Right now I don't think you can because we need SearchBarComponent which is not standalone and is declared in SharedModule.

hirokiterashima: Ok- let's make SearchBarComponent into a standalone component soon.

hirokiterashima (approved):
LGTM. 

I noticed a few more naming improvements to consider:
1. Should ```TeacherRunListComponent.tagsFilterChanged()``` be called ```tagsSelected()``` or ```selectTags()```?
2. Should ```TeacherRunListComponent.tagsToFilterBy``` be called ```selectedTags```?

PR #1762, opened 2024-04-26
refactor(HeaderLinksComponent): Convert to standalone component

hirokiterashima:
## Changes
- Convert HeaderLinksComponent to standalone
- Clean up HeaderLinks component and spec files

## Test
- Header links displays and works as before
   - not signed in: should show "Features", "About Us", "Community", "Research", "Register" and "Sign in" links
   - signed in: should show "Welcome, {first name}" and the account menu button

## Note
It looks like ```<app-header-signin>``` block could (and should?) be move out of HeaderLinksComponent template and into the header component template instead? It's not really a link. We can discuss and address this in a separate task.


breity (approved):
Looks good. 👍

PR #1767, opened 2024-04-29
feat(MultipleLanguages): Delete translations when steps and lessons are deleted

hirokiterashima:
## Changes
- Delete translations associated with deleted step(s) and lesson(s). 

## Test Prep
- Test with ```issue-259-implement-unit-tagging``` API branch

## Test
- Multilingual unit:
   - Associated translation entries are deleted from translations.[language].json files (for all supported languages) when you delete 
      - one step 
      - multiple steps together
      - one lesson
      - multiple lessons together
- Non-multilingual unit:
   - Deleting steps and lessons work as before

### DIFF
--- a/src/app/services/deleteTranslationsService.spec.ts
+++ b/src/app/services/deleteTranslationsService.spec.ts
@@ -23,18 +23,18 @@ describe('DeleteTranslationsService', () => {
     projectService = TestBed.inject(TeacherProjectService);
     service = TestBed.inject(DeleteTranslationsService);
   });
-  deleteComponents();
+  tryDeleteComponents();
 });
 
-function deleteComponents() {
-  describe('deleteComponents()', () => {
+function tryDeleteComponents() {
+  describe('tryDeleteComponents()', () => {
     it('fetches all supported translations', () => {
       spyOn(projectService, 'getLocale').and.returnValue(
         new ProjectLocale({ default: 'en_us', supported: ['es', 'ja'] })
       );
       spyOn(configService, 'getProjectId').and.returnValue('123');
       spyOn(configService, 'getConfigParam').and.returnValue('/123/project.json');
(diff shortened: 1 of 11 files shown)


geoffreykwan:
Functionality works well.

Did you mean `at-units-in-multiple-languages` API branch?

I think the comments should be placed on the line before the code so that the lines don't exceed 100 characters.

I wonder if it would be good to have `deleteNodeService.deleteNode()` return the nodes that it deletes and then we get the components from that instead of getting the components before hand. We could do this in a future PR if we really felt like it.

It seems like there is commonality in all 3 places where we get the components. Perhaps this can be refactored. Maybe we can have a function that returns the components in a step and another function that returns the components in a lesson. Then we can use those functions as building blocks to get components from multiple steps and multiple lessons.

There seems to be some duplication where this code is in 3 locations.
```
      if (this.projectService.getLocale().hasTranslations()) {
        this.deleteTranslationsService.deleteComponents(components);
      }
```

hirokiterashima:
Yes, I see the duplication and opportunities to write it better, but I think we should refactor this in a separate task. 

The way I see it, biggest issue is that ```ProjectAuthoringStepComponent.delete()```, ```ProjectAuthoringLessonComponent.delete()``` and ```ProjectAuthoringComponent.deleteSelectedNodes()``` are similar to start with, but it's hard to refactor them right now because we have to wait for the node deletions to save to the server (and refresh the project) before removing the translation strings. Ideally, we'd just call ```this.deleteNodeService.deleteNode()```, which would handle everything (save deleted nodes, refresh project, and save deleted translations). But making this change would be out of the scope of this PR.

hirokiterashima:
Refactored some duplicate code to TeacherProjectService. PTAL

geoffreykwan (approved):
Looks good.

PR #1769, opened 2024-04-30
feat(Tag): Add custom color picker

geoffreykwan:
## Changes
- Added custom color picker for tags

You can see the component parameters [here](https://zefoy.github.io/ngx-color-picker/).

## Test
- Run `npm install`
- Make sure the custom color chooser works when creating a new tag and editing an existing tag


geoffreykwan:
We decided to use a different color picker.

PR #1774, opened 2024-05-01
feat(Tag): Show selected tags list

geoffreykwan:
## Changes
- We now show the list of selected tags to filter by
- Added clear filters button

## Test
- Make sure tags that are selected using the tags filter drop down show up in the list
- Make sure removing a tag using the tag x button removes the tag from the list and also unselects it in the tags filter drop down
- Make sure the clear filters button unselects all the tags to filter by

### DIFF
--- a/src/app/teacher/select-tags/select-tags.component.ts
+++ b/src/app/teacher/select-tags/select-tags.component.ts
@@ -1,4 +1,4 @@
-import { Component, EventEmitter, Output } from '@angular/core';
+import { Component, EventEmitter, Input, Output } from '@angular/core';
 import { CommonModule } from '@angular/common';
 import { MatDividerModule } from '@angular/material/divider';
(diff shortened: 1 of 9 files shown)


breity (changes requested):
Functionality works well. I added keyboard accessibility to the remove filter 'x' buttons and also updated the mat-chip label font-size to match mat-button font-size.

As discussed, I think we can take out the "Clear filters" option added in this set of changes and use the existing "Clear filters" option that is visible when a search term has been entered. So that option would appear when any tag or search filters are present and would clear out both.

geoffreykwan:
Ok, I've changed it so that we use the "Clear filters" link to clear out the tags. I also fixed the messages when there are no active runs or no archived runs.

breity (approved):
Looks good. 👍

PR #1775, opened 2024-05-02
feat(Tag): Use mat chip for run card tags

geoffreykwan:
## Changes
- Use mat chip for run card tags

## Test
- Make sure the run card tags show up properly


breity (approved):
Looks good. 👍 

I added an `aria-label` descriptor to the list of tags.

PR #1776, opened 2024-05-03
refactor(Authoring): Create branch path authoring component

geoffreykwan:
## Changes
- Created BranchPathAuthoringComponent
- Create Branch view now uses the BranchPathAuthoringComponent

## Test
- Make sure the create branch form still works properly and validates properly
- Test out all the criteria types
  - Workgroup ID
  - Score
  - Choice Chosen
  - Random
  - Tag

### DIFF
--- a/src/assets/wise5/authoringTool/add-branch/add-branch.component.ts
+++ b/src/assets/wise5/authoringTool/add-branch/add-branch.component.ts
@@ -13,9 +13,11 @@ import { MatInputModule } from '@angular/material/input';
 import { CommonModule } from '@angular/common';
 import { MatButtonModule } from '@angular/material/button';
 import { FlexLayoutModule } from '@angular/flex-layout';
-import { Choice } from '../../components/match/choice';
 import { SelectStepComponent } from '../../../../app/authoring-tool/select-step/select-step.component';
 import { SelectComponentComponent } from '../../../../app/authoring-tool/select-component/select-component.component';
+import { BranchPathAuthoringComponent } from '../branch-path-authoring/branch-path-authoring.component';
+import { RouterModule } from '@angular/router';
+import { BranchService } from '../../services/branchService';
 
 @Component({
   selector: 'add-branch',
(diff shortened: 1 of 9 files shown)


src/assets/wise5/authoringTool/add-branch/add-branch.component.ts
```diff
@@ -13,30 +13,34 @@ import { MatInputModule } from '@angular/material/input';
...
   imports: [
+    BranchPathAuthoringComponent,
     CommonModule,
     FlexLayoutModule,
     MatButtonModule,
     MatFormFieldModule,
     MatInputModule,
     MatSelectModule,
     ReactiveFormsModule,
+    RouterModule,
```
hirokiterashima, line 41: Where do we use this import?

src/assets/wise5/services/branchService.ts
```diff
@@ -5,6 +5,9 @@ import { PathService } from './pathService';
 
 @Injectable()
 export class BranchService {
+  readonly CHOICE_CHOSEN: string = 'choiceChosen';
+  readonly SCORE: string = 'score';
```
hirokiterashima: Should these variables be in a separate file, outside of BranchService? Having to depend on BranchService just to use these variables like we do now in BranchPathAuthoringComponent doesn't seem ideal.

hirokiterashima (commented):
Works as described. I added some comments inline about code. Lmk what you think.

src/assets/wise5/authoringTool/add-branch/add-branch.component.ts
geoffreykwan: The RouterModule is required for the cancel button routerlink in the template.

src/assets/wise5/services/branchService.ts
geoffreykwan: Good idea, I've made the changes.

hirokiterashima (approved):
LGTM. Looks like you can remove branchService dependency from BranchPathAuthoringComponent?

PR #1777, opened 2024-05-06
feat(Translation): Make source language text display more visible

breity:
## Changes
- Add translation icon and give source language text a background to make more visible to translators
- Fix long source language text overlapping content below

## Test
- Make sure source language text still displays properly when in translation mode in the AT.
- Try adding some source language text content that is long so that the mat-hint element takes up multiple lines. Ensure that the source language mat-hint pushes content below down and doesn't overlap.


hirokiterashima (approved):
LGTM

PR #1781, opened 2024-05-06
feat(Authoring): Create select merge step component

geoffreykwan:
## Changes
- Create SelectMergeStepComponent

## Test
- Make sure the SelectMergeStepComponent shows up once all the other necessary branch parameters are set
- If the author is creating a branch on the last step in a lesson, the only option should be "Create New Step"
- If the author is creating a branch on a step that is not the last in a lesson, the two options should be "Use Next Step In Unit" and "Create New Step"

### DIFF
--- a/src/assets/wise5/authoringTool/add-branch/add-branch.component.html
+++ b/src/assets/wise5/authoringTool/add-branch/add-branch.component.html
@@ -41,6 +41,20 @@
     [componentId]="formGroup.controls['componentId']?.value"
     [pathFormGroup]="formGroup.controls['pathFormGroup']"
   ></branch-path-authoring>
+  <select-merge-step
+    *ngIf="
+      formGroup.controls['criteria'].value === WORKGROUP_ID_VALUE ||
+      formGroup.controls['criteria'].value === RANDOM_VALUE ||
+      formGroup.controls['criteria'].value === TAG_VALUE ||
+      ((formGroup.controls['criteria'].value === SCORE_VALUE ||
+        formGroup.controls['criteria'].value === CHOICE_CHOSEN_VALUE) &&
+        formGroup.controls['nodeId'].value !== '' &&
+        formGroup.controls['componentId'].value !== '' &&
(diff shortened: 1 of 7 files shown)


hirokiterashima (commented):
Works a described. Seeing "Use Next Step in Unit" option made me wonder what that step is. (The interface gives no clue about where we're adding the branch...we might want to add this info somewhere on the page). Can we show the actual step title instead? Like "1.2 Burning Coal" Or something like "Use next Step in Unit: 1.2 Burning Coal".

src/assets/wise5/authoringTool/add-branch/add-branch.component.html
```diff
@@ -41,6 +41,20 @@ <h3 i18n>Create Branch</h3>
...
+    *ngIf="
+      formGroup.controls['criteria'].value === WORKGROUP_ID_VALUE ||
+      formGroup.controls['criteria'].value === RANDOM_VALUE ||
+      formGroup.controls['criteria'].value === TAG_VALUE ||
+      ((formGroup.controls['criteria'].value === SCORE_VALUE ||
+        formGroup.controls['criteria'].value === CHOICE_CHOSEN_VALUE) &&
+        formGroup.controls['nodeId'].value !== '' &&
+        formGroup.controls['componentId'].value !== '' &&
+        formGroup.controls['pathFormGroup'].valid)
+    "
```
hirokiterashima: Should this be a function, like ```showSelectMergeStep(): boolean {}```?

geoffreykwan:
I've added the branch step to the title of the page and the merge step title in the select option. I've also created the function showSelectMergeStep().

hirokiterashima (approved):
LGTM. The new text gives me more support as I make the branch.

A minor issue is that ```showSelectMergeStep()``` has a complexity of 9. I think using ```[WORKGROUP_ID_VALUE, ...].includes(this.formGroup.controls['criteria'].value)``` would make it a bit easier to read.

PR #1782, opened 2024-05-06
feat(Tag): Add tag functionality to personal library

geoffreykwan:
## Changes
- In personal library added
  - Apply tags button
  - Tags filter
  - Tags displayed in library card
- In selected tags component, added optional clear filters button

@breity The select tags filter alignment is a little off in the personal library. I tried adding `[subscriptSizing]="'dynamic'"` to the mat-form-field which fixes the problem in the personal library view but then the alignment is off in the class schedule. Can you look into figuring out a good solution for this?

## Test
- Test with https://github.com/WISE-Community/WISE-API/pull/274
- Make sure all the tag functionality works properly in the personal library
- Make sure the clear filters button works

### DIFF
--- a/src/app/modules/library/home-page-project-library/home-page-project-library.component.html
+++ b/src/app/modules/library/home-page-project-library/home-page-project-library.component.html
@@ -1,6 +1,6 @@
 <div class="library library-home notice-bg-bg">
   <div fxLayout="column" fxLayout.gt-sm="row">
-    <div class="content-block controls dark-theme library__header primary-bg" fxFlex.gt-sm="33">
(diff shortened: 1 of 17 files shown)


breity (changes requested):
Functionality looks great. I updated the tags filter alignment and some styles.

I found one bug: Checkboxes in the Apply Tags dropdown menu sometimes don't correctly reflect the units that are checked. For example:
Screenshot 2024-05-14 at 12 24 11 PM

geoffreykwan:
Ah good catch. I've fixed the problem with the apply tags checkbox values not updating correctly.

breity (approved):
Looks good. 👍

PR #1783, opened 2024-05-07
refactor(PossibleScoresComponent): Convert to standalone

hirokiterashima:
## Changes
- Convert PossibleScoresComponent to standalone component
- Clean up PossibleScoresComponent by removing unused code and adding types

## Note
- This changes required changes in some component spec files (e.g. AudioOscillator, DialogGuidance, AiChat, ...) but not others, so we're not being consistent across all the components. I think it'd be nice if we didn't have to import the PossibleScoresComponent in the component spec files. We should take a look in a separate PR.

## Test (in student VLE)
- PossibleScoresComponent loads and displays the possible score as before in the component header, alongside the prompt
    - ex: ```Who hears Fin first? (5 points)```


geoffreykwan (approved):
Looks good.

PR #1785, opened 2024-05-07
feat(Authoring): Use select for choice chosen when authoring branch path

geoffreykwan:
## Changes
- When authoring a branch with "Choice Chosen", use select drop down to choose choices

## Test
- Author a branch that uses "Choice Chosen" and make sure the paths use a select drop down and populates the options with the choice values


hirokiterashima (approved):
LGTM. 

Unrelated to this PR, but should this if check be outside of BranchPathAuthoring template, and instead be done in AddBranch template? Seems like this check should be done before BranchPathAuthoringComponent is initialized.
```
  *ngIf="
    nodeId !== '' &&
    componentId !== '' &&
    (criteria === SCORE_VALUE || criteria === CHOICE_CHOSEN_VALUE)
  "
```

geoffreykwan:
I originally tried to put that check in add-branch.component.html but it caused some issues with BranchPathAuthoringComponent not receiving the ngOnChanges events.

PR #1786, opened 2024-05-07
refactor(ComponentStateInfoComponent): Convert to standalone

hirokiterashima:
## Changes
- Convert ComponentStateInfoComponent to standalone component
- Remove ComponentStateInfoModule
- Clean up ComponentStateInfoComponent

## Notes
- Since ComponentStateInfoComponent's template only uses ```<save-time-message...```, maybe we can remove this component and directly us SaveTimeMessageComponent in the future.

## Test
- In the VLE and the Grading Tool, ComponentStateInfoComponent 
   - correctly displays the save/submit times for students' work as before
   - tooltip appears on hover, displaying the timestamp


geoffreykwan (approved):
Looks good.

PR #1807, opened 2024-05-20
refactor(DiscourseComponents): Convert to standalone

hirokiterashima:
## Changes
- Convert DiscourseLatestNewsComponent and DiscourseRecentActivityComponent to standalone components
- Move test from DiscourseFeedComponent to DiscourseLatestNewsComponent. DiscourseFeedComponent is abstract class, and we should be testing concrete classes for existing of list items in templates
- Clean up code
   - use self-closing tags
   - add private/protected modifiers

## Test
- Discourse feed shows up in the main home page and in the teacher home page, and the links to posts work as before


breity (approved):
Looks good. 👍

PR #1808, opened 2024-05-21
fix(Authoring): Fix add step before for branch path steps

geoffreykwan:
## Changes
- Performing add step before on the first step in a branch path now works

Note:
- Add step before on the merge step does not work
- Add step before on first step in a branch path that is in an inactive lesson does not work

## Test
- Add step before on the first step in a branch path and make sure it works for all 4 ways you can add a step
  - Create Your Own
  - Import From Another Unit
  - Automated Assessment
  - Interactive Simulation


src/assets/wise5/authoringTool/addNode/abstract-import-step/abstract-import-step.component.ts
```diff
@@ -29,7 +31,7 @@ export abstract class AbstractImportStepComponent implements OnInit {
       .copyNodes(nodesToImport, this.importProjectId, this.configService.getProjectId())
       .subscribe((copiedNodes: any[]) => {
         const nodesWithNewNodeIds = this.projectService.getNodesWithNewIds(copiedNodes);
-        this.insertNodesService.insertNodes(nodesWithNewNodeIds, this.targetId);
+        this.insertNodesService.insertNodes(nodesWithNewNodeIds, this.targetId, this.nextId);
```
hirokiterashima: Should we check for ```isFirstNodeInBranchPath(this.nextId)``` here, like we do in ```AddYourOwnNode.submit()```? This way, insertNodeService won't need to depend on CreateNodeBetweenService.

```
if (this.projectService.isFirstNodeInBranchPath(this.nextId)) {
      this.createNodeBetweenService.createNodeBetween(newNode, this.targetId, this.nextId);
    }
```

hirokiterashima (commented):
Functionality works, but the code is a bit difficult to follow.

Should ```CreateNodeBetweenService``` be called something more specific, like ```InsertFirstNodeInBranchPathService```? This is a specific procedure that we're trying to handle. Also, we're inserting node(s), not creating them.

The nextId variable name, when used with targetId is confusing. Since this variable is only used when we're adding a node at the beginning of a branch path, should we call it something more specific, like firstNodeInBranchPath?

On an unrelated note, I think that if we simplify (i.e. restrict) the UI a bit, we can simplify the code. Right now, we allow author to add step before every node. If we restrict it so that they can only add step before the first node in the lesson or first node in a branch path, the code would become easier to follow.

hirokiterashima:
I forgot to mention that the tests are failing too: ConfigureAutomatedAssessmentComponent and ChooseSimulationComponent.

src/assets/wise5/authoringTool/addNode/abstract-import-step/abstract-import-step.component.ts
geoffreykwan: I don't think we can do that since insertNodes() for loops over the nodes.

geoffreykwan:
Do we want to change all instances of nextId to firstNodeInBranchPath? It might seem a little weird to be setting a parameter named firstNodeInBranchPath in all the routes when not all of them use it.

geoffreykwan:
I've made the changes we discussed last week.

src/assets/wise5/authoringTool/add-step-button/add-step-button.component.ts
```diff
@@ -23,18 +23,53 @@ export class AddStepButtonComponent {
...
+      const previousNodeId: string = previousNodes[0].id;
+      if (this.projectService.isFirstNodeInBranchPath(this.nodeId)) {
+        this.goToAddStepViewForFirstStepInBranchPath(previousNodeId, this.nodeId);
+      } else {
+        this.goToAddStepViewForAfter(previousNodeId);
+      }
+    }
   }
 
   protected addStepAfter(): void {
```
hirokiterashima: Remove function and call ```goToAddStepViewForAfter(nodeId)``` directly from template?

hirokiterashima (commented):
New changes made the code easier to follow.

Is it possible to encapsulate the target fields (targetType, targetId, branchNodeId, firstNodeIdInBranchPath)  in a ```AddStepTarget``` class and pass that around instead of having to pass 4 fields everywhere? This would also reduce duplication.

```
Before:
    this.targetType = history.state.targetType;
    this.targetId = history.state.targetId;
    this.branchNodeId = history.state.branchNodeId;
    this.firstNodeIdInBranchPath = history.state.firstNodeIdInBranchPath;

After:
    this.target = history.state.target;
```

```
Before:
      state: {
        importProjectId: project.id,
        targetType: this.targetType,
        targetId: this.targetId,
        branchNodeId: this.branchNodeId,
        firstNodeIdInBranchPath: this.firstNodeIdInBranchPath
      }

After:
      state: {
        importProjectId: project.id,
        target: this.target,
      }
```

```
class AddStepTarget {
  branchNodeId?: string;
  firstNodeIdInBranchPath?: string;
  targetId: string;
  type: string;
}
```

I also added a comment inline.

geoffreykwan:
Ok, I've made the changes.

src/assets/wise5/authoringTool/add-step-button/add-step-button.component.ts
```diff
@@ -23,18 +24,45 @@ export class AddStepButtonComponent {
...
   }
 
-  protected addStepAfter(): void {
-    this.goToAddStepView(this.nodeId);
+  private goToAddStepViewForIn(groupId: string): void {
+    this.router.navigate(['add-node', 'choose-template'], {
+      relativeTo: this.route,
+      state: {
+        target: new AddStepTarget('in', groupId)
+      }
```
hirokiterashima: Do we need to add the target field, or can we just pass in the AddStepTarget as the state? What else do we need to pass in to the next route?
```suggestion
      state:  new AddStepTarget('in', groupId)
```

src/assets/wise5/authoringTool/add-step-button/add-step-button.component.ts
```diff
@@ -23,18 +24,45 @@ export class AddStepButtonComponent {
...
+    });
+  }
+
+  protected goToAddStepViewForAfter(previousNodeId: string): void {
+    this.router.navigate(['add-node', 'choose-template'], {
+      relativeTo: this.route,
+      state: {
+        target: new AddStepTarget('after', previousNodeId)
+      }
+    });
```
hirokiterashima: Do we need to add the target field, or can we just pass in the AddStepTarget as the state? What else do we need to pass in to the next route?
```suggestion
      state:  new AddStepTarget('after', previousNodeId)
```

src/assets/wise5/authoringTool/add-step-button/add-step-button.component.ts
```diff
@@ -23,18 +24,45 @@ export class AddStepButtonComponent {
...
   }
 
-  private goToAddStepView(nodeId: string): void {
+  private goToAddStepViewForFirstStepInBranchPath(
+    previousNodeId: string,
+    nextNodeId: string
+  ): void {
     this.router.navigate(['add-node', 'choose-template'], {
       relativeTo: this.route,
       state: {
```
hirokiterashima: Do we need to add the target field, or can we just pass in the AddStepTarget as the state? What else do we need to pass in to the next route?
```suggestion
      state: new AddStepTarget('firstStepInBranchPath', null, previousNodeId, nextNodeId)
```

src/assets/wise5/services/teacherProjectService.ts
```diff
@@ -344,7 +344,7 @@ export class TeacherProjectService extends ProjectService {
    * @param node the new node
    * @param nodeId the node to add after
    */
-  createNodeAfter(newNode, nodeId) {
+  createNodeAfter(newNode: any, nodeId: string) {
```
hirokiterashima: Also add return type?
```suggestion
  createNodeAfter(newNode: any, nodeId: string): void {
```

hirokiterashima (commented):
Looks good. I added new comments based on your latest changes. The idea is to only send AddStepTarget as the state (```state: new AddStepTarget(...)```), instead of wrapping it in an object that only contains the target field (```state: { target: new AddStepTarget(...) }```) in some areas to reduce extra wrapping/unwrapping. Lmk what you think.

geoffreykwan:
Ok, we set the target object as the state now instead of wrapping it.

hirokiterashima (approved):
LGTM. 

Some functions in InsertFirstNodeInBranchPathService don't have a return type, so we should add that before merge.

PR #1811, opened 2024-05-22
refactor(SaveStarterStateComponent): Convert to standalone

hirokiterashima:
## Changes
- Convert SaveStarterStateComponent to standalone component
- Clean up code
   - use self-closing tags
   - change texts "isDirty" -> "dirty", "Save Starter State" -> "Save starter state", etc

## Test
- Saving starter state in AT -> Preview Component dialog works as before


geoffreykwan (approved):
Looks good.

PR #1815, opened 2024-05-24
refactor(PreviewComponents): Convert to standalone

hirokiterashima:
## Changes
- Convert PreviewComponentButtonComponent and PreviewComponentDialogComponent to standalone components
- Clean up code
   - use self-closing tag
   - add private/protected modifier
   - Remove PreviewComponentModule and import components directly

## Test
- In AT, preview component works as before
   - starter state saves as before
- In CM, component preview appears as before in "Step Info" dialog


geoffreykwan (approved):
Looks good.

PR #1818, opened 2024-05-30
refactor(HeaderComponent): Convert to standalone

hirokiterashima:
## Changes
- Convert HeaderComponent to standalone component
- Remove HeaderModule and import HeaderComponent directly
- Clean up code
   - use self-closing tag
   - use new template control flow blocks
   - add private/protected modifiers
   
## Test
- In the home pages (home, about us, feature, etc), the HeaderComponent displays as before, and the links in the component work
   - in a small window, the mobile menu (3 horizontal lines) appears instead


breity (approved):
Looks good. 👍

PR #1837, opened 2024-06-12
feat(Authoring): Only allow certain add options for branch steps

### LINKED ISSUE #1836: feat(Authoring): Only allow certain add options for branch steps
Branch step
- Add step before

Branch path step
- Add step before
- Add step after

Branch merge step
- Add step after
- Branch off this step

geoffreykwan:
## Changes
Only allow these add step options for the branch steps

Branch step
- Add step before

Branch path step
- Add step before
- Add step after

Branch merge step
- Add step after
- Branch off this step

For regular steps the options should be
- Add step before
- Add step after
- Branch off this step

## Test
- Make sure the correct add step options show up for the branch step, branch path steps, and branch merge step

Closes #1836

### DIFF
--- a/src/assets/wise5/authoringTool/add-step-button/add-step-button.component.html
+++ b/src/assets/wise5/authoringTool/add-step-button/add-step-button.component.html
@@ -10,13 +10,17 @@
   <mat-icon>add_circle</mat-icon>
 </button>
 <mat-menu #step="matMenu">
+  @if(!branchMergePoint) {
   <button mat-menu-item (click)="addStepBefore()">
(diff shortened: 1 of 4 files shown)


breity (approved):
Functionality works well.

One issue I found is that if you add a step before a branch step that is the first step in the lesson, the new step is added to the end of the previous lesson, when it should be added to the beginning of the current lesson. I think this is an existing issue, not related to this PR.

I believe we also decided that we're going to move this functionality away from the "Add Step" button. The "Add Step" button will always add a step after and then we're going to add a "Branch Off This Step" icon button to the step list item itself. But I guess this will be an intermediate step? Hopefully it won't be too confusing for users.

PR #1839, opened 2024-06-14
refactor(Teacher StepTools): Convert to standalone

geoffreykwan:
## Changes
- Converted teacher StepTools to standalone

## Test
- Make sure the step tools in the Authoring Tool and Grading Tool work so that you can change step by using the drop down or previous and next buttons


hirokiterashima (approved):
LGTM.

Remove redundant test for ```showStepTools```
```
    @if (showStepTools) {
      <step-tools *ngIf="showStepTools" />
    }
```

Also, step-tools.component.spec.ts "should create" test case body is empty, so it's not really testing anything. We should either test that component exists (``` expect(component).toBeTruthy();```) or remove the test altogether?

PR #1842, opened 2024-06-19
refactor(Authoring): Create select path count component

geoffreykwan:
## Changes
- Created SelectPathCountComponent

## Test
- Make sure in the add branch view that changing the number of branch paths works properly


hirokiterashima (approved):
LGTM. Self-close the ```<select-path-count>``` element?

PR #1844, opened 2024-06-21
fix(AT): Copied unit does not immediately appear

### LINKED ISSUE #1843: bug(AT): Copied unit does not immediately appear
### Description

In AT > Unit list view, copying a unit does not immediately display the copied unit

hirokiterashima:
## Changes
- Fix issue where copied unit didn't immediately appear in AT > unit list view

## Test
- In AT > unit list view, copy a unit. The copied unit should appear immediately and be highlighted. Before, nothing happened.

Closes #1843


geoffreykwan (approved):
Looks good.

PR #1847, opened 2024-06-24
refactor(EditConstraintRemovalCriteria): Convert to standalone

geoffreykwan:
## Changes
- Convert EditConstraintRemovalCriteriaComponent to standalone

## Test
- Make sure constraint removal criteria authoring works in
  - Step constraints
  - Component (visibility) constraints


hirokiterashima (approved):
LGTM. Should some of the ```@if () {...}``` be ```@else if () {...}``` instead, like

```
    @else if (param.value === 'requiredSubmitCount') {
...
    @else if (
      param.value === 'choiceIds' &&
...
```

PR #1850, opened 2024-06-26
feat(Translation): Display current language in language chooser

breity:
## Changes
- Show current selected language for translating in authoring top bar.

## Test
- Make sure current language is displayed when switching languages.


hirokiterashima (approved):
LGTM. Should the tooltip be "Language to translate" instead of "Language to Translate"? We've been only capitalizing the first word recently (e.g. "Move step", "Add step", "Add lesson").

PR #1852, opened 2024-07-05
feat(Multilingual Unit): Copy single component generates new translations

hirokiterashima:
## Changes
- Copying a single component generates new translations ids and copies over translations for the new component
- Note: to keep the scope small, I did not implement copying components via other methods.

## Test
- In the node authoring view, click on the "Copy component" button for a component (not the checkbox -> copy method). This should:
   - copy the component
   - generate new translation ids for the new component (look at project.json)
   - copy the translations that were in the original component, but with new the new translation ids (look at translations.[language].json)
- Deleting components deletes translations from the translations.[language].json files as before. I extracted common code to a new EditTranslationsService class.


geoffreykwan (approved):
Looks good.

PR #1853, opened 2024-07-05
refactor(GenerateImageDialogComponent): Convert to standalone

hirokiterashima:
## Changes
- Convert GenerateImageDialogComponent to standalone component
- Clean up code
   
## Test
- Generating image of component and setting it as background (e.g. setting draw as background to graph) work as before


geoffreykwan (approved):
Looks good.

PR #1856, opened 2024-07-08
feat(Multilingual Unit): Copy multiple components generates new translations

hirokiterashima:
## Changes
- Copying multiple components generates new translations ids and copies over translations for the new components

## Test
- In the node authoring view, select multiple components (using the checkbox) and then click on the "Copy Components". Select a location for the copied components. This should:
   - copy the components
   - generate new translation ids for the new components (look at project.json)
   - copy the translations that were in the original components, but with new the new translation ids (look at translations.[language].json)

### DIFF
--- a/src/assets/wise5/authoringTool/node/chooseComponentLocation/choose-component-location.component.spec.ts
+++ b/src/assets/wise5/authoringTool/node/chooseComponentLocation/choose-component-location.component.spec.ts
@@ -5,6 +5,7 @@ import { ComponentTypeService } from '../../../services/componentTypeService';
 import { ConfigService } from '../../../services/configService';
 import { TeacherDataService } from '../../../services/teacherDataService';
 import { TeacherProjectService } from '../../../services/teacherProjectService';
+import { CopyTranslationsService } from '../../../services/copyTranslationsService';
 
 const nodeId = 'node1';
 const components = [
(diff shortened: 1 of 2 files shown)


geoffreykwan (changes requested):
I noticed two bugs when copying a component (using the copy single component button or the copy multiple components button).

Bug 1
1. Create a component and fill in the prompt for all the languages
2. Copy a component in the English (default) language
3. Switch to another language like Spanish
4. The Spanish prompt will sometimes be blank even though it should contain text

Bug 2
1. Create a component and fill in the prompt for all the languages
2. Copy a component in the English (default) language
3. Delete the new component
4. Refresh the Authoring Tool
5. Switch to another language like Spanish
6. The Spanish prompt in the original component will sometimes be blank even though it should contain text

Reproducing these bugs may depend on whether you previously viewed the Spanish view or had the component expanded or collapsed.

hirokiterashima:
This looks like an issue related to how the wise-docker-dev Nginx is configured. Following the example (see below) on the Nginx doc fixes some issues, but I still see a timeout after multiple consecutive copies right after each other. The problem does not seem to occur if you wait ~5 seconds after each action (copy, delete) before taking the next action. Since this seems like a corner case, I think we should proceed as is and look into it more closely if it starts to become a major problem with our users.

https://nginx.org/en/docs/http/ngx_http_upstream_module.html#keepalive

For HTTP, the [proxy_http_version](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_http_version) directive should be set to “1.1” and the “Connection” header field should be cleared:
```
upstream http_backend {
    server 127.0.0.1:8080;

    keepalive 16;
}

server {
    ...

    location /http/ {
        proxy_pass http://http_backend;
        proxy_http_version 1.1;
        proxy_set_header Connection "";
        ...
    }
}
```

geoffreykwan (approved):
Looks good.

PR #1857, opened 2024-07-09
refactor(RunEndedAndLockedMessage): Convert to standalone

hirokiterashima:
## Note
- @breity Can you style the message? When it's displayed in a large window, appears to be vertically overlapped by the top bar (see screenshot). When it's displayed in a smaller window, it appears ok.
Screenshot 2024-07-09 at 9 23 11 AM

## Changes
- Convert RunEndedAndLockedMessage to standalone component
- Clean up code
   
## Test
- As a teacher for a run, set its end date to the past and check "Lock After End Date" option in the run setting
- As a student for the run, verify that the message appears when you launch the run.


breity (approved):
Looks good. 👍

PR #1862, opened 2024-07-11
feat(Multilingual Unit): Copy multiple steps generates new translations

hirokiterashima:
## Changes
- Copying a single step and multiple steps generates new translations ids and copies over translations for the new steps

## Test (in the project authoring view)
- Click on the "Copy step" button for a single step. This should:
   - copy the step
   - generate new translation ids for the new step (look at project.json)
   - copy the translations that were in the original step, but with new the new translation ids (look at translations.[language].json)
- Select multiple components (using the checkbox) and then click on the "Copy" button. Select a location for the copied steps. This should:
   - copy the steps
   - generate new translation ids for the new steps (look at project.json)
   - copy the translations that were in the original steps, but with new the new translation ids (look at translations.[language].json)


geoffreykwan (approved):
Looks good.

PR #1864, opened 2024-07-16
refactor(ChooseNewComponent): Convert to standalone

hirokiterashima:
## Changes
- Convert ChooseNewComponent to standalone component
- Clean up code
   
## Test
- In the add new component dialog in the AT
   - the component options appear and work as before
   - "cancel" button works as before
   - "import" button works as before


geoffreykwan (changes requested):
The add component view works but the create new step view no longer shows the component type buttons.

hirokiterashima:
Good catch. Fixed. PTAL.

geoffreykwan (approved):
Looks good.

PR #1866, opened 2024-07-18
feat(Multilingual Unit): Log selected language and load previous-selected language on VLE start

hirokiterashima:
## Changes
- When in student run mode
   - Log language selection events in the event table. Every time the student switches a language, we save an event.
   - Display unit in previously-selected language, if applicable

## Test
- In a student run for a unit with multiple languages
   - switch between different languages on different steps. In the database, check that the events are saved in the events table.
      - The event name is "languageSelected"
      - The event data is ```{"requester":"user","language":"[LOCALE]"}```
      - You should also see a row for ```{"requester":"system","language":"[LOCALE]"}```. This is saved when the VLE loads, it looks up the last-saved language in StudentStatus and switches to it
   - Refresh the browser, or exit -> reload the unit. The unit should be rendered in your last-saved language. 
- In a student run for a unit without multiple languages, everything works as before


breity (approved):
Looks good. 👍

geoffreykwan (approved):
Looks good.

PR #1867, opened 2024-07-20
refactor(NotebookLauncherComponent): Convert to standalone

hirokiterashima:
## Changes
- Convert NotebookLauncherComponent to standalone component
- Clean up code
   
## Test
- In the VLE, the notebook launcher button (fab icon at the bottom-right corner)
   - appears and works as before when it is enabled
   - is hidden when it is disabled


breity (approved):
Looks good.

PR #1868, opened 2024-07-20
refactor(EditComponentWidthComponent): Convert to standalone

hirokiterashima:
## Changes
- Convert EditComponentWidthComponent to standalone component
- Clean up code
   
## Test
- In AT, for a component's advanced editing view, set the component's (of HTML, label, graph, draw, etc) width to 50 and verify in preview mode that it takes up half of the screen's width.


geoffreykwan (approved):
Looks good.

PR #1869, opened 2024-07-21
refactor(ChooseNewNodeTemplateComponent): Convert to standalone

hirokiterashima:
## Changes
- Convert ChooseNewNodeTemplateComponent to standalone component
- Clean up code
   
## Test
- In AT's Add Step view (where you choose between "create your own", "import from another unit...", etc), the page displays the options and works as before.
   - The 4 different options take you to the right pages 
   - "Cancel" takes you back to unit view


breity (approved):
Looks good.

PR #1870, opened 2024-07-25
feat(Multilingual Unit): Disable elements while translating

hirokiterashima:
## Changes
- Disable elements to disallow translators from making changes to the unit when they've chosen a non-default language. When they choose another language, we want them to focus on translating. This constraint also helps simplify our code.

## Test
In the AT for a multi-lingual unit
- authoring in default language works as before
- when you switch to a non-default language,
   - you can translate text as before
   - you can't make changes to the unit in any other way, for example:
      - add/delete/move/copy steps/components
      - edit component content, like switching "Selection Type" or adding new choices in MC component


breity (approved):
Looks good. 👍

PR #1872, opened 2024-07-29
fix(Events): Node exited event now saves when changing steps

### LINKED ISSUE #1871: bug(Events): Node exited event is not saved
1. Sign in as a student
2. Launch a unit
3. Navigate to different steps by using the previous or next step buttons or the select step drop down
4. Check the events database table. You will only see nodeEntered events and will not see any nodeExited events.

geoffreykwan:
## Changes
- Properly save nodeExited event when the student changes step. This likely stopped working when we converted to Angular. Previously we were saving the nodeExited event when the node component was destroyed but the node component no longer gets destroyed in Angular. Now we save the nodeExited event on this.studentDataService.currentNodeChanged$.

## Test
1. Sign in as a student
2. Launch a unit
3. Navigate to different steps by using the previous or next step buttons or the select step drop down
4. Check the events database table. You should see nodeExited events. Previously nodeExited events would not be saved when you changed steps.

Closes #1871

### DIFF
--- a/src/assets/wise5/vle/node/node.component.ts
+++ b/src/assets/wise5/vle/node/node.component.ts
@@ -134,6 +134,10 @@ export class NodeComponent implements OnInit {
     );
 
     this.studentDataService.currentNodeChanged$.subscribe(() => {
+      this.nodeUnloaded(this.node.id);
+      if (this.node.isEvaluateTransitionLogicOn('exitNode')) {
+        this.nodeService.evaluateTransitionLogic();
+      }
       this.initializeNode();
     });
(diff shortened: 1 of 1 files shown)


hirokiterashima:
I'm not sure if this is related to this PR, but while testing, I found that when I go to the project view, it adds an additional "nodeEntered" event for the group node **each time** in the events table, and additional "nodeExited" event for the group node and any subsequent node visits there after. Is this something that's easy to address in this PR, or should we do it in a separate PR?

hirokiterashima:
We decided to address the issue in a separate task

hirokiterashima (approved):
LGTM

PR #1879, opened 2024-08-06
feat(Branching): Implement edit branch submit changes

geoffreykwan:
## Changes
- Edit branch submit button now applies changes
- Note: Adding and removing branch paths will be implemented in a different PR

## Test
- Go to the Edit Branch view for a branch
- Make changes to the branch parameters
- Click "Submit"
- Make sure the changes are applied

### DIFF
--- a/src/assets/wise5/authoringTool/edit-branch/edit-branch.component.html
+++ b/src/assets/wise5/authoringTool/edit-branch/edit-branch.component.html
@@ -48,7 +48,9 @@
       (click)="submit()"
       class="button--progress"
     >
-      <mat-progress-bar *ngIf="submitting" mode="indeterminate"></mat-progress-bar>
+      @if (submitting) {
+        <mat-progress-bar mode="indeterminate" />
+      }
       <span i18n>Submit</span>
     </button>
   </div>
--- a/src/assets/wise5/authoringTool/edit-branch/edit-branch.component.html
+++ b/src/assets/wise5/authoringTool/edit-branch/edit-branch.component.html
@@ -1,6 +1,8 @@
 <div fxLayout="row wrap" fxLayoutAlign="space-between center">
   <h3 class="title" i18n>Edit Branch</h3>
-  <button mat-raised-button color="warn" (click)="removeBranch()" i18n>Remove Branch</button>
+  <button mat-raised-button color="warn" (click)="removeBranch()" [disabled]="submitting" i18n>
+    Remove Branch
+  </button>
 </div>
 <h5 i18n>Branch Step: {{ targetTitle }}</h5>
 <form fxLayout="column" [formGroup]="formGroup">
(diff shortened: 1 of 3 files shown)


src/assets/wise5/authoringTool/edit-branch/edit-branch.component.ts
```diff
@@ -208,8 +213,51 @@ export class EditBranchComponent extends AbstractBranchAuthoringComponent {
...
+      }
+    }
+  }
+
+  private updateTransitionLogic(params: any): void {
+    if (params.criteria === this.WORKGROUP_ID_VALUE || params.criteria === this.RANDOM_VALUE) {
+      this.node.transitionLogic.howToChooseAmongAvailablePaths = params.criteria;
+      this.node.transitionLogic.whenToChoosePath = 'enterNode';
+    } else {
+      this.node.transitionLogic.howToChooseAmongAvailablePaths = this.RANDOM_VALUE;
```
hirokiterashima, line 259: If we choose "choice chosen" in the UI, why do we set the field to "random" in the JSON here?

src/assets/wise5/authoringTool/edit-branch/edit-branch.component.html
```diff
@@ -39,8 +41,15 @@ <h5 i18n>Branch Step: {{ targetTitle }}</h5>
...
-    <button mat-raised-button color="primary" [disabled]="!formGroup.valid" (click)="submit()" i18n>
-      Submit
+    <button
+      mat-raised-button
+      color="primary"
+      [disabled]="!formGroup.valid || submitting"
+      (click)="submit()"
+      class="button--progress"
+    >
+      <mat-progress-bar *ngIf="submitting" mode="indeterminate"></mat-progress-bar>
```
hirokiterashima: use ```@if (submitting) {``` ?

hirokiterashima (commented):
Functionality seems to work. I had a question about the code. See inline comments.

src/assets/wise5/authoringTool/edit-branch/edit-branch.component.ts
geoffreykwan: If multiple paths are satisfied, we choose one of them randomly. For example, if one of the paths is set to a score of 1 and another path is set to a score of 1, we will randomly choose between those two paths. This is how it has always been. I don't know if this logic is ever utilized and we could probably change it to just choose the first path that is satisfied if we wanted in the future.

hirokiterashima: I see. We should change this in the future so that the field says "choiceChosen" instead.

hirokiterashima (approved):
LGTM. We should disable the submit button until the author makes a chance. This can be done in a future branch before merging to develop.

PR #1883, opened 2024-08-12
refactor(TeacherProjectService): Extract newProjectTemplate to its own file

hirokiterashima:
## Changes
- Move newProjectTemplate from TeacherProjectService to its own file

## Test
- Creating new unit works as before in the AT


geoffreykwan (approved):
Looks good.

PR #1884, opened 2024-08-13
refactor(TeacherProjectService): Extract NotifyAuthorService

hirokiterashima:
## Changes
- Extract code to notify concurrent authors for a unit to NotifyAuthorService

## Test
- Concurrent authors message ("Also currently editing this unit...") appears at the top of the AT as before when two authors are editing the same unit. The message should disappear when one is no longer authoring the unit (closes the AT or signs out)


geoffreykwan (approved):
Looks good.

PR #1886, opened 2024-08-13
refactor(ImportComponentService): Return Observable instead of Promise

hirokiterashima:
## Changes
- ImportComponentService.importComponents() now returns an Observable instead of a Promise. This addresses future deprecation issue

## Test
- Importing component(s) in the AT works as before


geoffreykwan (approved):
Looks good.

PR #1892, opened 2024-08-15
refactor(TeacherProjectService): Move constraint-related functions to ConstraintService

hirokiterashima:
## Changes
- Move constraint-related functions ```getConstraintDescriptions()``` and ```getCriteriaMessage()``` to ConstraintService and update references
- Remove first null-check in ```getCriteriaMessage()```. This function should take in a non-null criteria

## Test
- In AT unit view, if a node has any constraint, the constraint icon should appear next to the node title. Hovering over it should display the message describing the constraint
- In Student VLE, if a node has a constraint (for example, "must complete step before moving on" constraint), it should display the dialog with the constraint description when you try to move on before completing the step.


geoffreykwan (approved):
Looks good.

PR #1893, opened 2024-08-15
refactor(GradingEditComponentMaxScore): Set max score

hirokiterashima:
## Changes
- Get the Component and set max score directly instead of calling ```TeacherProjectService.setMaxScoreForComponent()```
- Remove ```TeacherProjectService.setMaxScoreForComponent()```
- Remove unused function ```TeacherProjectService.isNodeIdUsed()```

## Test
- In CM > grade by step or team, changing the max score for a component works as before
   - should save changes and update total max score for the node


breity (approved):
Looks good.

PR #1899, opened 2024-08-19
refactor(PathService): Clean up arePathsEmpty()

hirokiterashima:
## Changes
- Remove unnecessary null-checks
- Use ```array.every()``` to simplify logic
- Add test

## Test
- Verify that null-checks weren't really necessary
- Verify that the function works as intended


geoffreykwan (approved):
Looks good.

PR #1904, opened 2024-08-20
refactor(TeacherProjectService): Move doesAnyComponentHaveWork() to NodeAuthoringComponent

hirokiterashima:
## Changes
- Move ```TeacherProjectService.doesAnyComponentHaveWork()``` to ```NodeAuthoringComponent.hasComponentsWithWork()```
- Clean up ```NodeAuthoringComponent.hasComponentsWithWork()```
   - use ```array.some()```
   - remove unnecessary checks 

## Test
- In the AT > step editing view, the node's "show save/submit button" settings are properly set when you delete component(s) as before, e.g.,
   - if no component has submit button but there is at least one component that saves work, the node's save button should be set to show


geoffreykwan (approved):
Looks good.

PR #1913, opened 2024-08-21
refactor(GroupTabsComponent): Convert to standalone

hirokiterashima:
## Changes
- Convert GroupTabsComponent to standalone
- Clean up code
- Update tests to use harness instead of making the component's functions public (just for the sake of testing)

## Test Prep
Make a unit use the tabbed navigation theme. In AT > Project JSON view, add this as the top JSON attribute:
```
    "theme": "tab",
```

## Test (in VLE)
- For a unit with tab theme (see "Test Prep" above), the lessons appears as tabs, and you can click on the tab to go to the lesson
- For a unit with regular (non-tab) theme, the VLE loads and works as before


breity (approved):
LGTM.

PR #1914, opened 2024-08-22
refactor(TeacherProjectService): Extract RemoveNodeIdFromTransitionsService

hirokiterashima:
## Changes
- Extract code related to removing node from transitions (when deleting and moving steps in the AT) to a new class, RemoveNodeIdFromTransitionsService
- Add private modifiers and return types to the functions that were moved over to the new class
   - Note: I kept removeTransitionsOutOfGroup() public because we had a test for it, and I just moved the test to RemoveNodeIdFromTransitionsService spec file.  In the future, we should make this function private and add tests for the remove() function.

## Test
- In the AT > unit view, try moving and deleting step(s) and lesson(s). They should work as before
   - the UI should update and correctly display the new state after move/deletion
   - the node transition(s) should update correctly


geoffreykwan (approved):
Looks good.

PR #1915, opened 2024-08-22
feat(Edit Branch View): Display branch path steps when no parameters

geoffreykwan:
## Changes
- Edit Branch view now displays branch path steps when "How to Choose Among Paths" is set to "Workgroup ID" or "Random"

## Test
- Open the edit branch view for a branch and make sure the steps in the paths are displayed when "How to Choose Among Paths" is set to "Workgroup ID" or "Random"


hirokiterashima (approved):
LGTM.

One thing I noticed was how we check for
```@if (nodeId !== '' && componentId !== '') {``` at the root of the template. Since we don't show anything if this evaluates to false, should we do this check in the templates that embed ```<branch-path-authoring>```, e.g., in add-branch.component.html and edit-branch.component.html?

PR #1920, opened 2024-08-26
refactor(TeacherProjectService): Move removeNodeFromInactiveNodes() to MoveNodesService

hirokiterashima:
## Changes
- Move ```TeacherProjectService.removeNodeFromInactiveNodes()``` and related functions to MoveNodesService. This is used when inactive nodes/groups are moved to the active section in the authoring tool.
- Remove unnecessary null checks in ```removeNodeFromInactiveNodes()```

## Test
- Verify that the null checks in ```removeNodeFromInactiveNodes()``` were not really necessary
- In the AT, moving nodes and groups works as before, in these cases:
   - inactive to active
   - inactive to inactive

### DIFF
--- a/src/assets/wise5/services/moveNodesService.ts
+++ b/src/assets/wise5/services/moveNodesService.ts
@@ -206,12 +206,11 @@ export class MoveNodesService {
 
   private removeNodeFromInactiveNodes(nodeId: string): void {
     let node = null;
-    let parentGroup = this.projectService.getParentGroup(nodeId);
-    if (parentGroup != null) {
+    if (this.projectService.getParentGroup(nodeId) != null) {
       this.projectService.removeChildFromParent(nodeId);
     }
 
-    let inactiveNodes = this.projectService.project.inactiveNodes;
+    const inactiveNodes = this.projectService.project.inactiveNodes;
     for (let i = 0; i < inactiveNodes.length; i++) {
       let inactiveNode = inactiveNodes[i];
       if (inactiveNode.id === nodeId) {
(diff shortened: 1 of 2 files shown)


src/assets/wise5/services/moveNodesService.ts
```diff
@@ -200,7 +200,46 @@ export class MoveNodesService {
...
+    this.removeNodeFromInactiveNodes(node.id);
     this.projectService.addInactiveNodeInsertAfter(node, nodeIdToInsertAfter);
   }
+
+  private removeNodeFromInactiveNodes(nodeId: string): void {
+    let node = null;
+    let parentGroup = this.projectService.getParentGroup(nodeId);
+    if (parentGroup != null) {
+      this.projectService.removeChildFromParent(nodeId);
+    }
```
geoffreykwan: We can call the function in the if conditional.

```suggestion
    if (this.projectService.getParentGroup(nodeId) != null) {
      this.projectService.removeChildFromParent(nodeId);
    }
```

src/assets/wise5/services/moveNodesService.ts
```diff
@@ -200,7 +200,46 @@ export class MoveNodesService {
...
+
+    let inactiveNodes = this.projectService.project.inactiveNodes;
```
geoffreykwan: We can change the let to const.

```suggestion
    const inactiveNodes = this.projectService.project.inactiveNodes;
```

geoffreykwan (approved):
Looks good.

PR #1925, opened 2024-08-28
refactor(TeacherProjectService): getNodesByToNodeId()

hirokiterashima:
## Changes
- Simplify ```TeacherProjectService.getNodesByToNodeId()``` using ```concat``` and ```filter()```
- Remove unnecessary null check in ```RemoveNodeIdFromTransitionsService.remove()```. The call to ```TeacherProjectService.getNodesByToNodeId()``` should never return an array with null element(s).
- Remove line that was never used in ```RemoveNodeIdFromTransitionsService```:
```
        // get all the nodes that have a transition to the node we are removing
        const nodesByToNodeId = this.projectService.getNodesByToNodeId(groupIdWeAreMoving);
```
- Clean up other related code

## Test
- Verify that the null check in ```RemoveNodeIdFromTransitionsService.remove()``` was really not necessary.
- These actions work as before in the AT > unit view:
   - Moving step(s)/group(s) in/between active and inactive sections
   - Deleting step(s)/group(s) in/between active and inactive sections

### DIFF
--- a/src/assets/wise5/services/nodeService.ts
+++ b/src/assets/wise5/services/nodeService.ts
@@ -8,6 +8,7 @@ import { ChooseBranchPathDialogComponent } from '../../../app/preview/modules/ch
 import { DataService } from '../../../app/services/data.service';
 import { Observable, Subject } from 'rxjs';
 import { ConstraintService } from './constraintService';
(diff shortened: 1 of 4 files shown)


geoffreykwan (approved):
Looks good.

PR #1926, opened 2024-08-28
refactor(NodeService): chooseTransition()

hirokiterashima:
## Changes
- Remove null checks in ```NodeService.chooseTransition()```
   - transitionLogic that is passed in to this function should always be non-null
   - ```transitionLogic.transitions``` should never be null. It should always be an array (can be empty)
- Clean up getAvailableTransitions using ```filter()```
- Clean up related code

## Test
- Transitions work as before in student VLE, a unit with/without branches

### DIFF
--- a/src/assets/wise5/services/nodeService.ts
+++ b/src/assets/wise5/services/nodeService.ts
@@ -8,6 +8,7 @@ import { ChooseBranchPathDialogComponent } from '../../../app/preview/modules/ch
 import { DataService } from '../../../app/services/data.service';
 import { Observable, Subject } from 'rxjs';
 import { ConstraintService } from './constraintService';
+import { TransitionLogic } from '../common/TransitionLogic';
 
 @Injectable()
 export class NodeService {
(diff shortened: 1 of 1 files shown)


src/assets/wise5/services/nodeService.ts
```diff
@@ -176,92 +177,87 @@ export class NodeService {
...
     if (existingPromise != null) {
       return existingPromise;
     }
     const promise = new Promise((resolve, reject) => {
       let transitionResult = this.getTransitionResultByNodeId(nodeId);
-      if (
-        transitionResult == null ||
-        (transitionLogic != null && transitionLogic.canChangePath == true)
-      ) {
+      if (transitionResult == null || transitionLogic.canChangePath == true) {
```
geoffreykwan: I think you can delete the "== true" in the if conditional.

```suggestion
      if (transitionResult == null || transitionLogic.canChangePath) {
```

geoffreykwan (approved):
Looks good.

PR #1927, opened 2024-08-28
refactor(ProjectService): getNodesWithTransitionToNodeId()

hirokiterashima:
## Changes
- Remove ```ProjectService.getNodesWithTransitionToNodeId()``` and write the logic directly in ```NodeService.getPrevNodeId()```, the only place where it's used
- Remove unnecessary null check (```nodeIdsByToNodeId == null```) in ```NodeService.getPrevNodeId()```. This should always be an array.
- Update tests

## Test
- In the Student VLE, navigating to previous node works as before


geoffreykwan (approved):
Looks good.

PR #1933, opened 2024-09-03
refactor(ProjectService): getAllPaths()

hirokiterashima:
## Changes
- Break down function to getStepPaths() and getLessonPaths(), and clean up getLessonPaths()

## Test
- getAllPaths() works as before. Test functions that call it and ensure that features that rely on it work as before.


geoffreykwan (approved):
Looks good.

PR #1940, opened 2024-09-13
Fix unit icon URL and Chinese filenames not auto-completing URLs.

VM8gkAs:
## Changes
1. Updated the `setFeaturedProjectIcon` method in `project-info-authoring.component.ts` to correctly construct the unit icon URL.
2. Modified the regex in `projectService.ts` to support Chinese characters in filenames.

## Test
 - I haven’t had the chance to fully test these changes. Could you please help by verifying the following?
 Thank you!

## Pic
 - Attached is a screenshot illustrating the original issue with the incorrect unit icon URL and the URL auto-completion for Chinese filenames.

1. 
 image

2. 
 image


hirokiterashima (approved):
LGTM

PR #1941, opened 2024-09-13
refactor(Authoring Tool): Remove old branch authoring route and component

hirokiterashima:
## Changes
- Removed route and component that we used to create and edit branches. 

## Test
- AT loads and works as before, and the old branch authoring route is no longer accessible.


breity (approved):
Looks good.

PR #1944, opened 2024-09-18
Exit CodeBuild with status 1 when build prod command fails

hirokiterashima:
This was allowing the codepipeline to continue execution when it should not have


hirokiterashima:
Since the task failed as expected during ```npm run build-prod```, we'll merge this PR.

PR #1945, opened 2024-09-18
fix(EmbeddedComponent): Break up i18n block in authoring

hirokiterashima:
## Changes
- Break up the checkbox span in EmbeddedComponent so it becomes easier to translate

## Test
- Build succeeds. Before, the build was failing when trying to put together the bundles for different languages


hirokiterashima:
Build passed, will merge.

PR #1958, opened 2024-09-27
Improved default language switching functionality

VM8gkAs:
## Changes
- Added `mat-radio-button` that needs to be disabled.
- Resolved the issue where a page refresh was required to edit after switching the default language.

## Test
After changing the default language, it will switch to the new default language.
image

## Other Issues
After switching to another language, `defaultLanguage` does not change.
In the file
`src/assets/wise5/authoringTool/components/abstract-translatable-field/abstract-translatable-field.component.ts`
```typescript
protected defaultLanguage: Language = this.projectService.getLocale().getDefaultLanguage();
``` 

image

### DIFF
--- a/src/assets/wise5/authoringTool/authoring-tool.component.ts
+++ b/src/assets/wise5/authoringTool/authoring-tool.component.ts
@@ -232,7 +232,7 @@ export class AuthoringToolComponent {
   }
 
   private getElements(): any[] {
-    const elementsToDisable = 'button,input,mat-checkbox,textarea,mat-icon[cdkdraghandle]';
+    const elementsToDisable = 'button,input,textarea,mat-radio-button,mat-checkbox,mat-icon[cdkdraghandle]';
     return Array.from(
       this.elem.nativeElement.querySelectorAll(`div.main-content ${elementsToDisable}`)
     ).concat(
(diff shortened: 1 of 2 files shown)


hirokiterashima:
@VM8gkAs Thanks for the contribution. 

Your changes should work as expected when the unit has not been translated into different languages. 

However, an issue arises when the unit (originally authored in lang1) has already been translated (to lang2, for example), and you switch the default language (from lang1 to lang2). You'd expect the unit to load with text in lang2, but right now, this is not the case because of the way the code works. To get this to work, switching the default language from lang1 to lang2 needs to swap out the translatable strings in project.json (which was saved in lang1) with values in translations.[lang2].json and then create translations.[lang1].json and copy strings over from project.json.

Until we have this swapping logic implemented, we were thinking about showing a message informing the author about this limitation, and to discourage author from switching default language if the unit has already been translated into another language. What do you think?

VM8gkAs:
**Option 1:**
Initially, all text in project.json is stored by default in `translations.[def_lang].json`.

**Option 2:**
When switching the default language, create or overwrite `translations.[lang_old].json`,
and attempt to load `translations.[lang_new].json`.
- 2.A. When translations.[lang_new].json exists:
Load translations.[lang_new].json and replace all text in project.json.
- 2.B. When translations.[lang_new].json does not exist:
Do not change any text in project.json.

**Option 3:**
When switching the default language, do not change any text in `project.json`.

I think showing a message to inform the author about the limitation is a good temporary solution. Since switching the default language is likely to occur in the early stages of project creation, sticking with **Option 3** and adding a reminder is feasible.
If we want to implement switching to a language included in the Additional languages, it may take some time to complete **Option 2.**

hirokiterashima:
LGTM. In addition, we will show a message to inform the author about the limitation as a temporary solution.

PR #2029, opened 2024-12-20
feat(node-authoring): Move add component button position

breity:
## Changes
- Move add component buttons to in between each component
- Add divider with animation on hover
- Only scroll component into view if necessary

## Test
- Adding and moving components both work properly


hirokiterashima (approved):
LGTM

PR #2036, opened 2025-01-09
feat(LibraryFiltersComponent): add research project filtering functionality

### LINKED ISSUE #2035: feat(Unit Library): Add Research Project unit filter
### Description

Add a new feature to let teachers filter units based on Research Projects, like ARISE and NLP-TIPS. This will allow teachers and grant reviewers to quickly search for units that were created for certain research projects.

hirokiterashima:
## Changes
- This change set allows authors to specify ResearchProject (RP) tags ("NLP-TIPS", "ARISE", etc) to library units. This is represented in the project JSON in ```metadata.researchProjects``` array. This will allow teachers and grant reviewers to quickly search for units that were created for certain research projects.
- Add select menu that appears only when there are library units with RP tags
- Update the filtered units according to the users' selections of the new RP tags

## Setup
- To add RP tag to a unit, open the unit's JSON editing view in the AT and add the following field to the metadata block:
```
        "researchProjects": [
            "ARISE",
            "NLP-TIPS"
        ],
```

## Test
- Test with https://github.com/WISE-Community/WISE-API/pull/284
- When no units with RP tags exists
- When one or more units with RP tags exist

Closes #2035

### DIFF
--- a/src/app/modules/library/standard.ts
+++ b/src/app/modules/library/standard.ts
@@ -3,3 +3,9 @@ export class Standard {
   name: string = '';
   children: Standard[] = [];
 }
+
+export type ResearchProjectTypes = 'ARISE' | 'NLP-TIPS';
+
+export class ResearchProject extends Standard {
+  name: ResearchProjectTypes;
+}
--- a/src/app/domain/projectFilterValues.ts
+++ b/src/app/domain/projectFilterValues.ts
@@ -1,6 +1,9 @@
+import { ResearchProjectTypes } from '../modules/library/standard';
+
 export class ProjectFilterValues {
   searchValue: string = '';
   disciplineValue: string[] = [];
   dciArrangementValue: string[] = [];
   peValue: string[] = [];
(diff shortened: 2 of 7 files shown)


src/app/modules/library/standard.ts
```diff
@@ -3,3 +3,9 @@ export class Standard {
   name: string = '';
   children: Standard[] = [];
 }
+
+export type ResearchProjectTypes = 'ARISE' | 'NLP-TIPS';
```
breity, line 7: I'm thinking we may want to make ResearchProjectTypes more generic and not restricted to specific values, similar to how we handle the standards now. That way, we don't need to change the code whenever we want to add a new research project option.

breity (approved):
Looks good. :+1:

I made one code comment.

src/app/modules/library/standard.ts
hirokiterashima: Sounds good. Let's address in a future task.

PR #2059, opened 2025-01-21
style(Import Component): Use same UI/UX flow as Import Step

### LINKED ISSUE #2056: style(Import Component): Use  same UI/UX as Import Step
## Description
- Use the same tab-system ("My Units", "Library Units") for letting author choose a unit as the import step wizard
- Use a similar UI/UX for letting author choose a component to import as the import step wizard

hirokiterashima:
## Notes
- @breity Please style as you see fit

## Changes
- Change UI/UX of the import component feature to match that of the import step feature
   - Step 1: select unit. Units are organized in "My Units" and "Library Units" tabs
   - Step 2: select component(s), and submit
   - (before, these steps were combined in one page)
- Clean up surrounding code
   - Convert ChooseImportComponentComponent to standalone
   - Remove ImportComponentModule

## Test
- Import component works as described above
- Import step works as before
- Try going back and canceling while going through the above scenarios. 
   - Note: when you click back on the the import component, we take them back to the step authoring view, but we don't show the "Add new component" popup again. Showing the popup involves a bit more book-keeping/code and I wasn't sure if this was worth the added complexity. Let me know if you think we should add this in.

Closes #2056

### DIFF
--- a/src/app/authoring-tool/import-step/choose-import-step/choose-import-step.component.html
+++ b/src/app/authoring-tool/import-step/choose-import-step/choose-import-step.component.html
@@ -1,4 +1,4 @@
-<h5 i18n>Choose the step(s) that you want to import, then select Next.</h5>
+<h5 i18n>Choose the step(s) that you want to import, then select Submit.</h5>
 <p fxLayout="row wrap" fxLayoutAlign="start center" fxLayoutGap="8px">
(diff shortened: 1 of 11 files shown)


breity (approved):
Looks good. 👍 

Functionality works and appearance mirrors import step UI/UX. I think it's fine that we don't re-open the add new component dialog when cancelling.

PR #2054, opened 2025-01-17
feat(MatchStudent): import ideas detected in Dialog Guidance as choices

hirokiterashima:
## Changes
- We now show ideas detected in a DG component as choices in a Match component. We use the "Connected component" feature. You can now add a connection to a DG component in Match's advanced > connected components.
- We can specify now the cRaterRubric in the DG component. It looks like this:
```
"cRaterRubric": {
        "ideas": [
            {
                "name": "1",
                "studentText": "Off topic"
            },
            {
                "name": "2",
                "studentText": "Confused speed of travel and amplitude"
            },
            {
                "name": "3",
                "studentText": "Confused speed of travel and pitch"
            },
            {
                "name": "4",
                "studentText": "vague: density and medium affect speed of travel"
            },
            {
                "name": "4a",
                "studentText": "Accurate density idea"
            },
            {
                "name": "5",
                "studentText": "Inaccurate conclusion: sound travels the same speed"
            },
            {
                "name": "5a",
                "studentText": "Inaccurate conclusion: sound is faster in air/slower"
            },
            {
                "name": "6",
                "studentText": "Inaccurate explanation: air is less dense than water"
            },
            {
                "name": "7",
                "studentText": "Inaccurate explanation: uses personal experience as evidence (sound is distorted under water)"
            },
            {
                "name": "7a",
                "studentText": "Inaccurate explanation: sound is distorted/muffled"
            },
            {
                "name": "8",
                "studentText": "Inaccurate explanation: uses personal experience as evidence (harder to walk under water)"
            },
            {
                "name": "9",
                "studentText": "Inaccurate explanation: water blocks sound/air lets sound pass"
            },
            {
                "name": "10",
                "studentText": "Inaccurate explanation: more resistance, more pressure under water"
            },
            {
                "name": "11",
                "studentText": "Inaccurate explanation: Uses context as evidence (dolphins are made to communicate below water)"
            },
            {
                "name": "12",
                "studentText": "Inaccurate explanation: Uses distance as evidence (same time, same distance, same speed)"
            },
            {
                "name": "13",
                "studentText": "Accurate conclusion: Sound travels faster in water"
            },
            {
                "name": "14",
                "studentText": "Accurate conclusion: Happy hears Fin first and Bubbles hears Fin later"
            },
            {
                "name": "16",
                "studentText": "Accurate explanation: Sound travels via vibrations"
            },
            {
                "name": "17",
                "studentText": "Accurate explanation: Engery is transferred from particle to particle"
            }
        ]
[...]

PR #2081, opened 2025-02-04
feat(MatchComponentDataExportStrategy): Include ideas detected by C-Rater

### LINKED ISSUE #2080: feat(MatchExport): Export choices that are ideas detected by CRater
Building on work that was done in this [issue](https://github.com/WISE-Community/WISE-Client/issues/2038), we now want to export the ideas that were used as choices in the Match component export.

Every idea that was detected should be in its own column. If a student had that idea, show which bucket it was placed in.

hirokiterashima:
## Changes
When C-Rater ideas are used as choices in the Match component, they will be exported when the user requests the the Match item export.

## Test Prep
- Use the same run as https://github.com/WISE-Community/WISE-Client/pull/2054. 
- Make sure you've done some work on this run with multiple students who had different ideas.

## Test
- Try exporting a Match component that contains C-Rater items as choices. Go to CM > Data Export > Export Item Data > Match. Try exporting in both Latest and All 
   - Every detected idea by all students in the class should have a column in the header row.
   - If a student had that idea, the cell should contain which bucket that idea is in.
   - If a student didn't have that idea, the cell should be blank.
   - Check other fields in the export and make sure they look correct
      - component revision number looks ok
      - correctness looks ok, etc
- Regression tests
   - Try exporting a Match component that doesn't contain C-Rater item, and verify that it works as before
   - Try exporting another component item (e.g., OR, Label) and verify that it works as before

Closes #2080

### DIFF
--- a/src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
+++ b/src/assets/wise5/classroomMonitor/dataExport/strategies/AbstractComponentDataExportStrategy.ts
@@ -101,10 +101,15 @@ export abstract class AbstractComponentDataExportStrategy extends AbstractDataEx
 
   protected abstract getComponentTypeWithUnderscore(): string;
 
-  protected getComponentStates(component: any): ComponentState[] {
+  protected getComponentStates(
+    component: any,
+    calculateRevisions: boolean = true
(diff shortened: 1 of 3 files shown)


breity (approved):
LGTM. 👍

PR #2087, opened 2025-02-06
chore(Application): Upgrade to Angular 19

hirokiterashima:
## Changes
- Upgrade to Angular 19
- Upgrade dependencies
   - wise-draw: was using fabric (depended on jsdom, which was causing building issues), changed to fabric-browser
   - fabric (depended on jsdom, which was causing building issues): changed to fabric-browser 
   - password-strength-meter: was not compatible with Angular 19
- Remove dom-autoscroller library since it wasn't doing anything
- Components are standalone by default. This is why there are many changed components

## Test
- Everything works as before


hirokiterashima:
@breity starting the dev-server shows a lot of angular-sass deprecation warnings. Can you try addressing them? It looks like it's 2 or 3 issues that occur in many files, so addressing one should remove many warnings. 

Also, can you address the minor styling issues that we discussed last week (border around topbar, notebook color, etc)?

breity (approved):
Looks good. 👍 

I fixed the styling we discussed and fixed the Back to Top button so that it appears again when scrolling up. I'm leaving the SASS deprecation warnings for now. These will be addressed when we update the Angular Material themes to v19 syntax.

Aaron-Detre (approved):
I found some potential issues while testing it out, but all of those issues are also present on the current Angular 18 version of the site. So those might be things to look at some other time, but as for this upgrade, everything looked good.

hirokiterashima:
@Aaron-Detre Great. Thanks for testing. Let's discuss the issue you found.

Issue #2093, opened 2025-02-13
refactor(SummaryDisplay): Clean code


Aaron-Detre:
The SummaryDisplay component which is used for the TeacherSummaryDisplayComponent seems confusing and might benefit from refactoring.

PR #2100, opened 2025-02-24
refactor(SummaryDisplay): Created class for series data

Aaron-Detre:
## Changes
- Created classes to hold series data so that it's easier to tell what series data is and how to use it.

## Test
- Make sure summary displays (step info, edit step with summary component, preview step with summary component, student run summary component) work the same as before.

#2093


hirokiterashima (approved):
LGTM

PR #2108, opened 2025-03-04
build(Application): Remove Webpack dependencies

hirokiterashima:
Clean up dependencies related to development build (as opposed to production)

## Changes
- Remove Webpack dependencies. These are already included/depended by Angular, so we don't need to declare them ourselves.
- Update @types/node. New dependency updates were throwing node:stream not found error.

## Test prep
- Stop Docker
- Install dependencies ```npm i```
- Start Docker

## Test
- Build server starts as before
- Local development loop works as before (make change to file and verify that it's updated in the browser)


breity (approved):
LGTM

PR #2090, opened 2025-02-12
feat(Grade by step): Display component summary views

hirokiterashima:
## Changes
This PR adds summaries for each component in a step in a set of tabs. Each component shows the prompt and completion %. Some components also show summary data (Multiple Choice, Match, items with idea detection enabled, any items with scores). It adds a select drop-down to choose which components to view in the student work view.

## Test
Choose a step in the Grade-by-step view. It should show the summary view for the first component in the step.

Verify the following:
   - Step completion and step mean score should be displayed at the top of the page
   - If the step has multiple components, you should see a select drop-down, and this will let you switch between the components.
   - If the component has a summary display (e.g. choices chosen on a MC component), it should show the summary view.
   - If the component has scores, it should show the score distributions in the summary view panel
   - Sort the student list by team/status/score
   - Expand and collapse each student, and expand all students and collapse all students
   - See students' work for the component, and provide scores and comments
   - (for @breity)
      - Milestone grading view works as before
      - PeerChat component shows Peer Grouping button
      - Milestone component shows Report button

Closes #2099

PR #2112, opened 2025-03-07
feat(Grade By Component): Show component average score

hirokiterashima:
This is a new feature for the grade by component feature branch.

This PR adds a new component, ComponentAverageScoreComponent, that shows the average score of the students in the period for the component. Previously, we were only able to show the average score for a node.

## Changes
- ComponentAverageScoreComponent shows
  - average score for the current period, if available
     - we only count workgroups that have a score 
  - shows "N/A", if not available

## Test
- Choose different nodes, components, and periods (including all periods). The average score should display as described above.

### DIFF
--- a/.github/workflows/test-coverage.yml
+++ b/.github/workflows/test-coverage.yml
@@ -6,7 +6,6 @@ on:
   push:
     branches: [develop]
   pull_request:
-    branches: [develop]
 jobs:
   test-coverage:
     runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


breity (changes requested):
Looks good. Functionality works.

One suggestion: Maybe we shouldn't show the max score if there is no average score. For example, "N/A / 2 Mean Score" looks confusing. I think it would be better to just show "N/A Mean Score" like we do for the component score.

I will work on styling in #2090.

breity (approved):
LGTM.

I changed `N/A` to `-`, as we discussed.

PR #2113, opened 2025-03-10
feat(Grade By Component): Show component completion

hirokiterashima:
This is a new feature for the grade by component feature branch.

This PR adds a new component, ComponentCompletionComponent, that shows the percentage of workgroups in the period that has completed the component. Previously, we were only able to show the completion for a node.

## Changes
- ComponentCompletionComponent shows percentage of workgroups in the period that has completed the component, from 0% to 100%

## Test
- Choose different nodes, components, and periods (including all periods). The completion percentage should display as described above.


breity (approved):
Looks good. I updated to round the completion percentage to the nearest whole number.

Will work on styling in #2090.

Issue #294, opened 2021-08-17
Remove ng-switch-default from component.ts and previewComponent.ts


geoffreykwan:
Also in nodeController.ts, remove getComponentStatePromiseFromChildScope() and getComponentStateFromChildScope().

PR #2128, opened 2025-03-25
feat(Analytics): Add Google Tag Manager support

breity:
## Changes
- Add Google Tag Manager support

## Test
- Review GTM setup
- Depends on WISE-API [feat(Analytics): Add Google Tag Manager support #294](https://github.com/WISE-Community/WISE-API/pull/294)
- Note: Testing will only work on production environments or if you temporarily disable the `if (environment.production) {` check in `app.component.ts`


hirokiterashima (approved):
LGTM. 

I wasn't able to test locally, I kept getting 403 Forbidden. But this may be because I didn't set up my GTM correctly to handle requests from localhost. I saw that it was making the request to the correct URL, so we should test together against production and if it works there, we're good!

PR #2129, opened 2025-03-25
refactor(EditOpenResponseAdvancedComponent): Replace ngIf and ngFor statements

Aaron-Detre:
## Changes
- Replaced ngIf and ngFor statements with @if and @for statements
- Made other minor improvements

## Test
- Make sure open response advanced settings editor works normally


hirokiterashima (approved):
LGTM.

Remember that ```<div>``` by default breaks to the next line, so removing it will affect the layout.

PR #2136, opened 2025-04-11
feat(Embedded): Grow model file name input vertically

### LINKED ISSUE #2123: feat(Embedded): Grow model filename input field vertically in authoring
In the Embedded authoring view, if you enter a long text in the "Model file name" input field, the input field textbox stays the same width and height, so you have to scroll to the left to see the first characters. This makes it difficult to see and edit a really long url, like with the plant glucose simulations.

Contrast this behavior with that of when you're typing a prompt for an Open Response component. The text input height grows as you enter more text. This lets you see the entire text and easily edit specific parts.

Change the embedded authoring to behave the same way.

Aaron-Detre:
## Changes
- Changed the file name input field in the embedded component authoring tool to expand vertically for long URLs.

## Test
- Make sure that the input field grows vertically for long URLs.
- Make sure that text entered in the input field is still saved and the desired file can still be retrieved.

Closes #2123


hirokiterashima (approved):
LGTM

PR #2142, opened 2025-04-16
Fixed advanced editing on existing components with CRater enabled

### LINKED ISSUE #2141: bug(OpenResponse): Advanced editing is broken on existing components with CRater enabled
## Description
- If you try to edit the advanced settings on an existing open response component with CRater enabled, there are tons of errors, and sections fail to render.
- If you click to disable CRater and then re-enable it, it works normally.

Aaron-Detre:
## Changes
- Create cRater and empty cRater.rubric when advanced settings are opened if none exist.

## Test
- Make sure you can open advanced editing on a unit with a pre-existing CRater-enabled open response component.
    - There should be no errors, and everything should render normally. 

Closes #2141


hirokiterashima:
It looks like this change will add the cRater field to any OpenResponse component, even if CRater is not used. Can we modify it so it's only set for cRater-enabled items?

hirokiterashima (approved):
LGTM

PR #2149, opened 2025-04-21
fix(AT): Clicking lesson/step title does not check checkbox

### LINKED ISSUE #2127: bug(AT): Cannot open step when using phone
## Description

When I open a unit in the AT, I cannot click on the step title to open it. AT thinks that I selected the checkbox. 

Image

## Fix
Apply these fixes to both unit editing (list of steps) and step editing (list of components) view:
- open the step/component for editing when user clicks on the step/component title
- only clicking on the checkbox selects the item

Aaron-Detre:
## Changes
- For lessons, clicking the title expands/collapses the lesson instead of checking the checkbox.
- For steps, clicking the title opens the step editor instead of checking the checkbox.

## Test
- Clicking the titles no longer checks the checkbox, and instead does its new expected behavior.

Closes #2127


hirokiterashima (approved):
LGTM

PR #2157, opened 2025-04-23
build(Application): Only generate messages if tests pass

### LINKED ISSUE #2156: build(Application): Only generate messages if tests pass
## Description
- If tests fail, committing changes to messages.xlf can be confusing since the checks on the updated messages commit will pass, making it look like the PR has passed all checks. Also, there is no point in extracting messages if the tests fail.

Aaron-Detre:
## Changes
- Run test-coverage before generate-messages in workflows

## Test
- If tests fail, messages should not be extracted.
- If tests pass, the workflow should attempt to extract messages.
- (Seems to be working as expected if you check commit history--ignore commits from other branch that go mixed in when testing on non-develop branch...)

Closes #2156


hirokiterashima (approved):
LGTM.

PR #2158, opened 2025-04-25
feat(Curriculum): Add My Units tab

Aaron-Detre:
## Changes
- Added a My Units tab to the curriculum page when a user is logged in as a teacher.
- Added a link to the authoring tool on the curriculum page when a user is logged in as a teacher.

## Test
- Personal and shared units appear in the My Units tab.
- Active vs archived units are shown when you toggle the radio buttons.
- Can add new tags and filter by tags.
- Tab labels show the correct number of projects.
    - Filtering by tag and toggling active/archived does not affect the label, which seems strange to me, but that's how it is currently working in the teacher library, so maybe that's intentional?
    - Also, the changes I made broke the label for community built projects in the teacher library (always says 0 now), but since we're getting rid of that soon and that's such a small thing, I figured it would be alright. I can change it back to how it was if that's a problem.


src/app/curriculum/curriculum.component.ts
```diff
@@ -1,12 +1,70 @@
...
+import { PublicLibraryComponent } from '../modules/library/public-library/public-library.component';
+import { UserService } from '../services/user.service';
 
 @Component({
-  selector: 'curriculum',
-  imports: [LibraryFiltersComponent, MatIconModule, PublicLibraryComponent],
-  templateUrl: './curriculum.component.html',
-  styleUrl: './curriculum.component.scss'
+  imports: [
+    LibraryFiltersComponent,
```
hirokiterashima, line 17: Add CommonModule?

src/app/curriculum/curriculum.component.ts
```diff
@@ -1,12 +1,70 @@
...
+  ngOnInit(): void {
+    this.setShowMyUnits();
+    this.getLibraryProjects();
+    this.subscribeNumUnitsVisible();
+  }
+
+  private setShowMyUnits(): void {
+    this.userService.getUser()?.subscribe((user) => {
+      this.showMyUnits = user.roles && user.roles.includes('teacher');
+    });
```
hirokiterashima: Is there a need to check the result of calling getUser() since the function always returns a BehaviorSubject? Also, use userService.isTeacher()?
```suggestion
    this.userService.getUser().subscribe(() => {
      this.showMyUnits = this.userService.isTeacher();
    });
```

src/app/curriculum/curriculum.component.ts
```diff
@@ -1,12 +1,70 @@
...
+    }
+  }
+
+  private subscribeNumUnitsVisible(): void {
+    this.libraryService.numberOfPersonalProjectsVisible$.subscribe(
+      (num) => (this.numMyUnitsVisible = num)
+    );
+    this.libraryService.numberOfPublicProjectsVisible$.subscribe(
+      (num) => (this.numPublicUnitsVisible = num)
+    );
```
hirokiterashima: We need to keep track of these subscriptions (using ```subscriptions.add()```) and unsubscribe (using ```subscriptions.unsubscribe()```) when this component is destroyed. See how other components do this for reference.

src/app/curriculum/curriculum.component.ts
```diff
@@ -1,12 +1,70 @@
...
+  }
+
+  protected getPublicTabLabel(): string {
+    return `Public (${this.numPublicUnitsVisible})`;
+  }
+
+  protected getMyUnitsTabLabel(): string {
+    return `My Units (${this.numMyUnitsVisible})`;
+  }
```
hirokiterashima: $localize ?

hirokiterashima (changes requested):
Behavior seems to work as expected. Your comments about Community library sounds good.

See inline code comments.

breity (approved):
Functionality looks good. 👍 I updated the styles and layout.

hirokiterashima (approved):
LGTM. Let's look more closely at the thumb nail issue in the next task

PR #2162, opened 2025-04-29
feat(Site): Remove unit library from teacher home

breity:
## Changes
Remove unit library from teacher home.

- Redirect `/teacher/home/schedule` to `/teacher/home`.
- Redirect `/teacher/home/library/`* to `/curriculum`.
- Always show site header links when logged in as teacher.

## Test
- Make sure /teacher/home still works and shows Class Schedule (run list).
- Make sure /curriculum unit libraries work as before.
- Make sure header links (Features, About Us, Curriculum, etc.) always appear when logged in as a teacher.


hirokiterashima (changes requested):
Looking great. 👍 

I realized that we actually need to implement the ```/curriculum/personal``` path in this PR, and go there after copying a unit and flash/animate the newly copied unit to draw attention to it, as we did before. Right now, when we copy a Public unit, the user won't see any visual queues that the project has been copied into My Units.

We should also fix the issue where the thumbnails for My Units appear as gray boxes before this PR gets merged into develop. I'll look into this.

hirokiterashima:
Changes look great. I really like the new icons and ability to toggle between class schedule and unit library.

I noticed that copying run unit takes me to My Units, but it doesn’t highlight/flash the new unit. It works when I copy a unit that is in the "Public" or "My Units". It would be nice to draw attention to a new unit, but if it's not an easy fix, let's not sweat it.

Also, it looks like we can now remove PersonalLibraryDetailsComponent and the dialog dependency from LibraryComponent and its child components?

hirokiterashima:
I found a new issue: I can't set up a run from the Library. When I click on "use with class", I see the "Use with Class" dialog but the body is empty, and there's an error in the js console:

```
core.mjs:6673 ERROR Error: MatDatepicker: No provider found for DateAdapter. You must add one of the following to your app config: provideNativeDateAdapter, provideDateFnsAdapter, provideLuxonDateAdapter, provideMomentDateAdapter, or provide a custom implementation.
    at createMissingDateImplError (datepicker.mjs:236:10)
    at new _MatDatepickerInputBase (datepicker.mjs:4172:15)
    at new _MatDatepickerInput (datepicker.mjs:4438:5)
```

Strange... @Aaron-Detre and I saw this error before, and we fixed it by adding ```provideNativeDateAdapter()``` in CreateRunDialogComponent's providers array. It worked then, and since CreateRunDialogComponent is a standalone component, I'd thought it'd continue to work in the new /curriculum route. 🤔

hirokiterashima:
When I cleared out my .angular/cache and node_modules folder and restarted the server, the seems to have fixed itself.

PR #2164, opened 2025-05-03
feat(Match): Summary display

Aaron-Detre:
## Changes
- Added summary display for match components.
    - Unordered components have actual display, ordered components show a message saying summary display currently only for unordered match.

## Test
- Buckets contain the bucket name and the choices placed in that bucket.
- Choices contain the choice name and the number of students who placed that choice in that bucket.
- The same choice can appear in multiple buckets if multiple students placed that choice in different buckets.
- Choices are sorted descending by count (number of times placed in that bucket).
- Buckets without any choices placed in them are not shown.
- Buckets with more than 3 choices have a 'Show More' button which can be toggled to see all the choices placed in the bucket.
- If no choices have been placed in any buckets, show a message letting teachers know the summary will be shown once students complete the activity.


breity (approved):
Looks good. 👍

hirokiterashima (approved):
LGTM

PR #2169, opened 2025-05-07
refactor(LibraryService): Move ProjectFilterValues out of class

hirokiterashima:
## Changes
We used to store an instance of ProjectFilterValue in LibraryService. Now, we use the provider field instead to share the instance among the components (ProjectFilter, Curriculum, ProjectLibrary)

## Test
- Filters in the homepage and curriculum pages work as before, as anonymous user and signed in user.
- Try switching back and forth between those pages and makes sure that each page starts with fresh (empty) set of filters.

### DIFF
--- a/src/app/curriculum/curriculum.component.ts
+++ b/src/app/curriculum/curriculum.component.ts
@@ -11,6 +11,7 @@ import { Subscription } from 'rxjs';
 import { UserService } from '../services/user.service';
 import { MatButtonModule } from '@angular/material/button';
 import { Router, RouterModule } from '@angular/router';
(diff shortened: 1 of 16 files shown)


Aaron-Detre (approved):
Everything looks good, but I did find one problem, though I'm not sure if it's an issue to do with this pull request. When I type in "localhost/curriculum/personal" (or /public), I get a "Whitelabel Error Page." The same thing happens if I go from the curriculum to the community or research sites and then click the go back button. I checked on production and it's not happening there, so I'm not sure whether it's an issue with this branch or just my local environment.

PR #2177, opened 2025-05-12
feat(CRater): Ping endpoint

### LINKED ISSUE #2159: feat(CRater): Ping endpoint
## Description

When a student enters a step with a C-Rater enabled component, we want to make sure that the endpoint is available, and not require the student to wait while the model does a cold-start.

The expected behavior is as follows:
- Ping the item endpoint when student enters a step that contains a C-Rater item
- Ping every 3 minutes thereafter
- Stop ping when the student leaves the step or exits the unit

Aaron-Detre:
## Changes
- When you enter a step with a Dialog Guidance or CRater Open Response component, PingEndpointService collects the item IDs.
- When the student focuses on the textbox, the client sends ping requests for all collected item IDs to the API.
- Pings will be sent every 4 minutes and 55 seconds regardless of whether or not they are still focusing on the textbox or not.
    - For simplicity + edge case where student types out response > unfocuses > model sleeps > refocuses > sends immediately
- When the student leaves the step, the item IDs are cleared out and pings are no longer sent. 

## Test
- Make sure that ~30 seconds after focusing on a "berkeley_" item ID that hasn't been used recently, scoring requests come back quickly without the cold start waiting time.
- Requires [this PR](https://github.com/WISE-Community/WISE-API/pull/304) for the API.

Closes #2159


hirokiterashima (approved):
LGTM

PR #2185, opened 2025-05-20
feat(Match): Filter out detected ideas from pre-authored choices #2184

### LINKED ISSUE #2184: feat(Match): Filter out detected ideas from pre-authored choices
### Description
Request from a researcher:
We want to continue collecting data on how students perceive the way their ideas are detected in the dialog. I piloted this in X's classroom before:

Image

We are thinking of a new research design for this activity. Here is the prompt:
"Here are some ideas the thought buddy found in student answers.
Sort them into: Ideas I have, other ideas I agree with, ideas i don’t agree with"

To ensure all students have something to sort, apart from the ideas detected in their dialog, we’d also like to include three additional ideas for every student:
1. A non-normative idea that they likely won’t select
2. A sophisticated idea that is unlikely to be included in their initial response
3. A common idea that many students typically include

To avoid redundancy, we’d want the system to check whether a student already expressed any of these three additional ideas and exclude them from the sorting activity if so.

hirokiterashima:
## Changes
If the student has an idea that is authored as one of the choices, remove it from the choices to avoid duplication. We consider it a duplication if choice.id (authored choice) and idea.name (detected student idea) are the same.

## Test Prep
For a unit with DG connected to a Match component, author choices in the match component. Using the advanced JSON editing, give the choices the same ids as the choice names in the CRaterRubric.ideas.name.

## Test
Ensure that choices with the same id do not appear more than once in the choices array. 

Closes #2184

### DIFF
--- a/src/assets/wise5/components/match/match-student/match-student-default/match-student-default.component.ts
+++ b/src/assets/wise5/components/match/match-student/match-student-default/match-student-default.component.ts
@@ -613,11 +613,17 @@ export class MatchStudentDefaultComponent extends ComponentStudent {
   }
 
   private addIdeasToSourceBucket(responses: any[], rubric: CRaterRubric): void {
-    getUniqueIdeas(responses, rubric).forEach((idea) => {
-      const choice = new Choice(idea.name, idea.text);
-      this.choices.push(choice);
-      this.getBucketById(this.sourceBucketId).items.push(choice);
-    });
+    getUniqueIdeas(responses, rubric)
+      .filter((idea) => !this.isInSourceBucket(idea.name))
+      .forEach((idea) => {
+        const choice = new Choice(idea.name, idea.text);
+        this.choices.push(choice);
+        this.getBucketById(this.sourceBucketId).items.push(choice);
+      });
+  }
+
+  private isInSourceBucket(choiceId: string): boolean {
+    return this.sourceBucket.items.some((item) => item.id === choiceId);
   }
 
   protected addChoice(): void {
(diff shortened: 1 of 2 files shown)


src/assets/wise5/components/match/match-student/match-student-default/match-student-default.component.ts
```diff
@@ -613,11 +613,17 @@ export class MatchStudentDefaultComponent extends ComponentStudent {
...
+      .filter((idea) => !this.isInSourceBucket(idea.name))
+      .forEach((idea) => {
+        const choice = new Choice(idea.name, idea.text);
+        this.choices.push(choice);
+        this.getBucketById(this.sourceBucketId).items.push(choice);
+      });
+  }
+
+  private isInSourceBucket(choiceId: string): boolean {
+    return this.sourceBucket.items.some((item) => item.id === choiceId);
```
breity:
```suggestion
      .filter((idea) => !this.isInSourceBucket(idea.text))
      .forEach((idea) => {
        const choice = new Choice(idea.name, idea.text);
        this.choices.push(choice);
        this.getBucketById(this.sourceBucketId).items.push(choice);
      });
  }

  private isInSourceBucket(choiceId: string): boolean {
    return this.sourceBucket.items.some((item) => item.value === choiceId);
```

breity: I believe we should be passing in `idea.text` instead of `idea.name` because we should be comparing the text description of the idea to the choice (`item.name` is just the idea number). We should also be checking against `item.value` which is the text content of the choice (not `item.id`, which is the randomly generated string identifier for the choice). This will be easier for authors than needing to the change the json.

breity (changes requested):
Looks good, but I needed to make a couple changes to get it to work properly. See code suggestions and explanations.

breity (approved):

PR #2189, opened 2025-05-27
Grade by component show all components

hirokiterashima:
(Reminder: this should be a merge commit instead of squash)

## Changes
- Revert to showing all components in the node
- Revert router so it only goes up to ".../node/nodeId" (before: it included "/component/componentId")
- Hide summary display (this will be addressed in a separate PR before merging to develop)
- Add FilterComponentsComponent to let teacher select which components to show
   - Filtering components now updates the status, score and max score column fields appropriately. 

## Test
- Node grading works as it currently does in production


breity (approved):
Looks good. 👍 

I updated styles to match [mockup](https://balsamiq.cloud/s8brvqj/p9j1not/r793B).

PR #2195, opened 2025-06-16
refactor(AT): build optimizations June 2025

hirokiterashima:
This PR converts more components to stand-alone and lazy-load the routes where possible. By doing this, the resulting build files will contain more files with smaller size. So loading the AT will become faster for users. This will also let us build faster because each subsequent re-builds should compile to smaller file sizes.

## Changes
For routes in "/edit" (authoring tool):
- Convert components to standalone
- Convert sub-routes to lazy-load
- Clean up components if easy
- Remove unused code

## Test
- Test with https://github.com/WISE-Community/WISE-Docker-Dev/pull/18
- Make sure that AT loads and pages work as before


breity (approved):
Looks good to me.

I found one issue that should be fixed before merging: The edit component advanced button isn't showing up when a component is focused in editing mode. The component DOM element is there, but there is no content.

Screenshot 2025-06-27 at 11 21 14 AM

PR #2198, opened 2025-06-25
feat(TimedSteps): New type of step containing a series of timed components

Aaron-Detre:
## Changes
- Added authoring for timed steps on the node authoring page.
- Steps marked as timed display a series of components one at a time, each with its own countdown timer.
- When the timer reaches 0 or the proceed button is clicked, that component is hidden and the next one is made visible.
- When all components have been completed, a message saying the step has been completed is displayed.
- When working on a timed step, navigation to other steps using the step toolbar is disabled. (You could still edit the URL in theory)

## Test
- Make sure everything works as described above.
- If you have viewed or worked on a timed component, and then you close the tab and reopen it, you should see the next component instead of the one you were looking at before.
- Components with no time limit specified and negative time limits are treated the same as a time limit of 0 and basically skipped.
- Previewing should function the same as working on a timed step normally, except when you return to the timed step after completing it, you should start from the beginning rather than seeing the step completed message.

## Other
- I'm not currently saving time spent on each component, so that might be something to add later. I guess you would just store how long it takes them to press the proceed button if they press it at all? It wouldn't necessarily tell you how long it took them to work on the component, since they could finish it and then wait for time to run out. I guess it would be on the unit author to make it clear that they should click proceed as soon as they're done working.
- I think originally we were talking about proceeding when the student clicks submit on the component, but I'm not sure it would be clear that that is what's going to happen when they click submit, and not all component types have a submit button. I think that a dedicated proceed button right next to the timer makes it pretty obvious, but it does make knowing the actual time spent on the component a little trickier.

### DIFF
--- a/src/assets/wise5/authoringTool/node/node-authoring/node-authoring.component.ts
+++ b/src/assets/wise5/authoringTool/node/node-authoring/node-authoring.component.ts
@@ -221,7 +221,8 @@ export class NodeAuthoringComponent implements OnInit {
   }
 
   protected isTimed(): boolean {
-    return this.node.timed !== null && this.node.timed;
+    const isTimed = this.projectService.getNodeIsTimed(this.node.id);
+    return isTimed !== null && isTimed;
   }
 
   protected saveProject(): Promise<any> {
--- a/src/assets/wise5/authoringTool/node/node-authoring/node-authoring.component.ts
+++ b/src/assets/wise5/authoringTool/node/node-authoring/node-authoring.component.ts
@@ -12,10 +12,10 @@ import { TeacherNodeService } from '../../../services/teacherNodeService';
 import { DeleteTranslationsService } from '../../../services/deleteTranslationsService';
 
 @Component({
-    selector: 'node-authoring',
-    templateUrl: './node-authoring.component.html',
-    styleUrls: ['./node-authoring.component.scss'],
-    standalone: false
+  selector: 'node-authoring',
+  templateUrl: './node-authoring.component.html',
+  styleUrls: ['./node-authoring.component.scss'],
+  standalone: false
 })
 export class NodeAuthoringComponent implements OnInit {
   components: ComponentContent[] = [];
@@ -119,7 +119,7 @@ export class NodeAuthoringComponent implements OnInit {
     if (parseProject) {
       this.projectService.parseProject();
     }
-    return this.projectService.saveProject();
+    return this.saveProject();
   }
 
   protected deleteComponent(
@@ -139,7 +139,7 @@ export class NodeAuthoringComponent implements OnInit {
 
   private deleteComponentsOnServer(components: ComponentContent[]): void {
     this.checkIfNeedToShowNodeSaveOrNodeSubmitButtons();
-    this.projectService.saveProject().then(() => {
+    this.saveProject().then(() => {
       this.deleteTranslationsService.tryDeleteComponents(components);
     });
   }
@@ -213,10 +213,18 @@ export class NodeAuthoringComponent implements OnInit {
     if (scroll) {
       this.highlightComponents([this.components[currentIndex]]);
     }
-    this.projectService.saveProject();
+    this.saveProject();
   }
 
   protected editComponent(componentId: string): void {
     this.editingComponentId = componentId;
+  }
+
+  protected isTimed(): boolean {
+    return this.node.timed !== null && this.node.timed;
+  }
+
+  protected saveProject(): Promise<any> {
+    return this.projectService.saveProject();
   }
 }
--- a/src/assets/wise5/authoringTool/node/node-authoring/node-authoring.component.ts
+++ b/src/assets/wise5/authoringTool/node/node-authoring/node-authoring.component.ts
@@ -221,8 +221,8 @@ export class NodeAuthoringComponent implements OnInit {
   }
 
   protected isTimed(): boolean {
-    const isTimed = this.projectService.getNodeIsTimed(this.node.id);
-    return isTimed !== null && isTimed;
+    const node = this.projectService.getNodeById(this.node.id);
+    return node.timed !== null && node.timed;
   }
 
   protected saveProject(): Promise<any> {
--- a/src/assets/wise5/themes/default/themeComponents/stepTools/step-tools.component.ts
+++ b/src/assets/wise5/themes/default/themeComponents/stepTools/step-tools.component.ts
@@ -50,10 +50,10 @@ export class StepToolsComponent implements OnInit {
   protected toNodeId: string;
 
   constructor(
-    protected nodeService: NodeService,
-    protected nodeStatusService: NodeStatusService,
-    protected projectService: ProjectService,
-    protected studentDataService: StudentDataService,
+    private nodeService: NodeService,
+    private nodeStatusService: NodeStatusService,
+    private projectService: ProjectService,
+    private studentDataService: StudentDataService,
     private timedNodeService: TimedNodeService
   ) {}
 
--- a/src/assets/wise5/themes/default/themeComponents/stepTools/step-tools.component.ts
+++ b/src/assets/wise5/themes/default/themeComponents/stepTools/step-tools.component.ts
@@ -1,5 +1,5 @@
 import { CommonModule } from '@angular/common';
-import { Component, OnInit, ViewEncapsulation } from '@angular/core';
+import { Component, Input, OnInit, ViewEncapsulation } from '@angular/core';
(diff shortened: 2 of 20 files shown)


src/assets/wise5/themes/default/themeComponents/stepTools/step-tools.component.ts
```diff
@@ -124,4 +133,17 @@ export class StepToolsComponent implements OnInit {
...
+    const noNode = direction === 'prev' ? !this.prevId : !this.nextId;
+    return noNode || this.isUnfinishedTimedStep();
+  }
+
+  protected isUnfinishedTimedStep(): boolean {
+    if (this.timedStep) {
+      return !this.timedStepCompleted;
+    } else {
+      return false;
+    }
```
hirokiterashima, line 147: Write in one line?

src/assets/wise5/themes/default/themeComponents/stepTools/step-tools.component.ts
```diff
@@ -43,14 +44,17 @@ export class StepToolsComponent implements OnInit {
   protected nodeStatus: any;
   protected nodeStatuses: any;
   protected prevId: string;
+  @Input() private timedStep: boolean;
```
hirokiterashima, line 47: Instead of making timedStep an input to this component, can we set this in updateModel()?
```suggestion
  private timedStep: boolean;
```

src/assets/wise5/vle/vle.component.html
```diff
@@ -26,7 +26,7 @@
 
 <ng-template #defaultVLETemplate>
   @if (layoutState === 'node') {
-    <step-tools class="control-bg-bg mat-elevation-z1" />
+    <step-tools class="control-bg-bg mat-elevation-z1" [timedStep]="isCurrentNodeTimed()" />
```
hirokiterashima, line 29: Calculate timedStep boolean in  StepToolsComponent.updateModel()? Remove isCurrentNodeTimed() function?
```suggestion
    <step-tools class="control-bg-bg mat-elevation-z1" />
```

src/assets/wise5/authoringTool/node/node-authoring/node-authoring.component.ts
```diff
@@ -213,10 +213,19 @@ export class NodeAuthoringComponent implements OnInit {
...
+    this.saveProject();
   }
 
   protected editComponent(componentId: string): void {
     this.editingComponentId = componentId;
   }
+
+  protected isTimed(): boolean {
+    const node = this.projectService.getNodeById(this.node.id);
+    return node.timed !== null && node.timed;
```
hirokiterashima, line 225: Simplify?
```suggestion
    return this.projectService.getNodeById(this.node.id).timed;
```

src/assets/wise5/vle/timed-node/timed-node.component.ts
```diff
@@ -0,0 +1,191 @@
...
+    SubmitSurveyComponent
+  ],
+  selector: 'timed-node',
+  styleUrl: './timed-node.component.scss',
+  templateUrl: './timed-node.component.html'
+})
+export class TimedNodeComponent extends NodeComponent {
+  private componentTimers: number[];
+  private currentComponentIndex = 0;
+  private currentInterval: any;
```
hirokiterashima, line 42:
```suggestion
  private currentInterval: number;
```

src/assets/wise5/vle/timed-node/timed-node.component.ts
```diff
@@ -0,0 +1,191 @@
...
+      });
+      this.components.forEach((component, index) => {
+        this.componentToVisible[component.id] = index === this.currentComponentIndex;
+      });
+      this.startComponentTimer();
+    }
+  }
+
+  private async skipViewedComponents(): Promise<void> {
+    if (!this.isPreview()) {
```
hirokiterashima, line 85: Do this !isPreview() check before we call this function?

src/assets/wise5/vle/timed-node/timed-node.component.ts
```diff
@@ -0,0 +1,191 @@
...
+      if (componentsViewed.includes(component.id) && componentIndex >= this.currentComponentIndex) {
+        this.currentComponentIndex = componentIndex + 1;
+      }
+    });
+  }
+
+  private setStepCompletedIfNecessary() {
+    if (this.currentComponentIndex >= this.node.components.length) {
+      this.stepCompleted = true;
+    }
```
hirokiterashima, line 106: Write in one line and rename to setStepCompleted()?

src/assets/wise5/vle/timed-node/timed-node.component.ts
```diff
@@ -0,0 +1,191 @@
...
+      'componentViewed',
+      {}
+    );
+  }
+
+  private getCurrentComponentId(): string {
+    return this.components.at(this.currentComponentIndex).id;
+  }
+
+  private getComponentSubmitArgs(componentId: string) {
```
hirokiterashima, line 137:
```suggestion
  private getComponentSubmitArgs(componentId: string): any {
```

src/assets/wise5/vle/timed-node/timed-node.component.ts
```diff
@@ -0,0 +1,191 @@
...
+    this.hideComponents(++this.currentComponentIndex);
+    if (this.currentComponentIndex < this.node.components.length) {
+      this.startComponentTimer();
+    } else {
+      this.stepCompleted = true;
+      this.timedNodeService.broadcastIsNodeCompleted(true);
+    }
+  }
+
+  private saveUnsavedWork() {
```
hirokiterashima, line 155:
```suggestion
  private saveUnsavedWork(): void {
```

src/assets/wise5/vle/timed-node/timed-node.component.ts
```diff
@@ -0,0 +1,191 @@
...
+    if (!this.isPreview()) {
```
hirokiterashima, line 156: Check !isPreview() before calling this function?

src/assets/wise5/vle/timed-node/timed-node.component.ts
```diff
@@ -0,0 +1,191 @@
...
+    super.ngOnInit();
+    await this.skipViewedComponents();
+    this.timedNodeService.broadcastIsNodeCompleted(this.stepCompleted);
+    if (!this.stepCompleted) {
+      this.componentTimers = this.node.components.map((component: ComponentContent) => {
+        return component.timeLimit ?? 0;
+      });
+      this.components.forEach((component, index) => {
+        this.componentToVisible[component.id] = index === this.currentComponentIndex;
+      });
```
hirokiterashima, line 79: Can we call this.hideComponents() here?
```suggestion
      this.components.forEach((component, index) => {
        this.componentToVisible[component.id] = index === this.currentComponentIndex;
      });
```

src/assets/wise5/vle/timed-node/timed-node.component.ts
```diff
@@ -0,0 +1,191 @@
...
+    return `${this.addZeroPadding(minutes)}${minutes}:${this.addZeroPadding(seconds)}${seconds}`;
+  }
+
+  private addZeroPadding(digits: string): string {
+    const numZeros = 2 - digits.length;
+    return '0'.repeat(numZeros > 0 ? numZeros : 0);
+  }
+
+  protected updateComponentVisibility(): void {
+    return;
```
hirokiterashima, line 184: It'd be good to have a comment about why we're overloading it with an empty body.
```suggestion
    // do nothing because ...
```

hirokiterashima (changes requested):
Nice job! 👍  The functionality mostly seems to work as you described. I agree that the having the "proceed" button next to the timer makes it clearer for the user. 

I noticed some minor issues (below) and code improvement suggestions (inline). 

## Issues, improvements
- See code improvement suggestions inline
- In preview, I finish the step and see the "You have completed this step” message. I then go to previous step and come back, and it lets me work on the first component again.
- Countdown glitch. Sometimes, due to some combination of letting the component timer finish and clicking on "proceed" for others, the timer seems to decrease erratically, like in 2-second decrements. See this video:

https://github.com/user-attachments/assets/9281d6dc-912e-4cc8-bd7b-5fa4cda7c146

## Questions
- Should we default to 60(?) seconds for each timed component in the authoring tool? We should require a non-null number for the timeLimit value to simplify code. Having a default can help ensure that.
- Right now, the timeLimit input is always shown for all components in the step in the. authoring tool. Should we only show it when the component is being edited? In this scenario, it would be good to show the time limit value for each component.

PR #2211, opened 2025-08-16
build(Application): Upgrade to Angular 20

hirokiterashima:
## Changes
- Upgrade Angular from v19 to v20

## Test
- CM, AT, VLE work as before


breity (approved):
LGTM. 👍

PR #2216, opened 2025-08-27
feat(Site): Add more info about run types to create run dialog

breity:
## Changes
- Displays more information about Default and Survey runs in the create run dialog.

## Test
- Make sure create run dialog works as before.


hirokiterashima (approved):
LGTM

PR #2220, opened 2025-09-09
feat(GraphComponent): Conditionally disable decimals for yAxis based on authored content

hirokiterashima:
## Changes
- Conditionally disable decimal labels for y-axis. Previously, we were always disabling decimal labels. This was leading to strange behavior when the data points and max values were decimal values, as reported by Kelly.

## Test prep
- Download unit 44494 and import it locally. Steps 3.7 and 3.8 have the data explorer with decimal values.
- Enable decimals for y-axis for the graph component. 
   - Open JSON authoring for the graph components on steps 3.7 and 3.8 and set ```allowDecimals``` to true for yAxis. Note that there are multiple yAxis in this component, and they are all set to false. You’ll need to set it to true for all of them.

## Test
- Preview the steps. Make sure that the y-axis shows the correct labels as you choose different options to graph.

### DIFF
--- a/src/assets/wise5/components/graph/graph-student/graph-student.component.ts
+++ b/src/assets/wise5/components/graph/graph-student/graph-student.component.ts
@@ -602,7 +602,9 @@ export class GraphStudent extends ComponentStudent {
    */
   drawGraphHelper(resolve) {
     this.turnOffXAxisDecimals();
-    this.turnOffYAxisDecimals();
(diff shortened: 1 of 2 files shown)


breity (approved):
Looks good. The fix works as described. 👍 

We should also decide on a way to fix this without having to edit the component JSON. Maybe set `allowDecimals` to true for all y-axes if "Round values to" is set to anything other than integer or something like that?

PR #2222, opened 2025-09-11
chore(Application): Upgrade to Tailwind 4

breity:
## Changes
- Upgrade Tailwind dependency to v4.

## Test
- Make sure styles and layout appear and function as before.


hirokiterashima (approved):
Looks good. 

Just out of curiosity- why do we need to include
```
@reference "tailwindcss";
```

in only some files (e.g., crater-rubric.component.scss, ideas-summary.component.ts, ...) but not in others (e.g., edit-crater-description.component.scss)?

Is there a rule to follow on when to include this line?

breity:
> Just out of curiosity- why do we need to include
> 
> ```
> @reference "tailwindcss";
> ```
> 
> in only some files (e.g., crater-rubric.component.scss, ideas-summary.component.ts, ...) but not in others (e.g., edit-crater-description.component.scss)?
> 
> Is there a rule to follow on when to include this line?

Yep, we only need to include the reference to Tailwind in any files that use `@apply` to apply Tailwind styles  to a css selector (like "px-2 py-1 text-start align-top" in crater-rubric.component.scss, for example).

PR #2243, opened 2025-11-13
style(AuthoringTool): Organize unit listing page using tabs

hirokiterashima:
## Notes
- Please style as you see fit
- This is the landing page when users clicked on the "Authoring Tool" button from the Unit Library or Teacher Home. The page looked out-dated, so I changed the buttons to match and used MatTabs to organize the two sections better.
- I thought it might be good to add "Authoring Tool" to the left of the "Create New Unit" and "Class Schedule" buttons to match the Unit Library, but since we already display "Authoring Tool" in the top bar, I omitted it.

## Changes
- Organize unit listing page with tabs ("My Units" and "Shared Units")
- Add text to "Create New Unit" and "Class Schedule" buttons and move them to the right to match the rest of the site
- Move copy and preview buttons to the right, change hover background color, and add "Edit unit" tooltip to match the rest of the AT

## Test
- AT unit listing page works as before and the page layout looks better than before


breity (approved):
Looks good. I simplified the template and cleaned up styles. Switched to using MatActionList for lists of units and added keyboard accessibility for unit selection.

I kept the hard-coded blue background color on hover for now, since it matches other parts of the AT. But this should be removed/replaced when refreshing the AT design.

PR #2244, opened 2025-11-14
feat(HomeComponent): Defer loading library until on viewport

hirokiterashima:
## Changes
- Convert HomePageProjectLibrary and ShareProjectDialog to standalone
- Defer library loading on the homepage until it's in view, i.e., when you scroll down to it. This will save users from having to load the units and their thumbnails until they actually scroll to the section, which will improve the initial page load time.

## Test
- Homepage loads as before
- "Curriculum Offerings" section of the homepage should load when it comes into view. You can test this by seeing that project_thumb.png's for the units are loaded when you scroll down to the section, not when the page is first loaded.
- Unit Library page (/curriculum) works as before
- Sharing units works as before

### DIFF
--- a/src/app/home/home.component.html
+++ b/src/app/home/home.component.html
@@ -93,7 +93,11 @@
         >
           Curriculum Offerings
         </h1>
-        <app-home-page-project-library></app-home-page-project-library>
+        @defer (on viewport) {
+          <app-home-page-project-library />
+        } @placeholder {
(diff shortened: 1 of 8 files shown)


breity (approved):
Looks good. 👍 

Is the `@placeholder` div necessary or can we just include the `@defer` content?

hirokiterashima:
Thanks. Yes, removing it will result in a compile-time error.

PR #2246, opened 2025-11-22
refactor(VLE): Remove modules part 2

hirokiterashima:
This is part 2 of the work to remove modules from WISE-Client and replacing them with standalone components. Also, I removed RoutingModules and replaced them with simple arrays instead. Since the change set has already grown large, I broke up this work into multiple parts to make it easier to review. 

## Changes
- Remove modules and replace them with directly importing standalone components
- Remove routing modules and replace them with simple arrays
- Clean up related code

## Test
- VLE, AT, and CM work as before


src/assets/wise5/authoringTool/components/component-type-selector/component-type-selector.component.spec.ts
```diff
@@ -16,14 +17,37 @@ let userService: UserService;
...
+      { type: 'MultipleChoice', name: 'Multiple Choice' },
+      { type: 'OpenResponse', name: 'Open Response' },
+      { type: 'OutsideURL', name: 'Outside URL' },
+      { type: 'PeerChat', name: 'Peer Chat' },
+      { type: 'HTML', name: 'HTML' },
+      { type: 'ShowGroupWork', name: 'Show Group Work' },
+      { type: 'ShowMyWork', name: 'Show My Work' },
+      { type: 'Summary', name: 'Summary' },
+      { type: 'Table', name: 'Table' }
+    ]);
```
breity, line 49: Perhaps this could be done without hard-coding. Because we'll need to update whenever a component type is added, removed, or edited, correct?

breity (approved):
Looks good 👍

PR #2253, opened 2026-01-08
feat(NodeAuthoring): Show component preview in current language

hirokiterashima:
## Changes
- In node authoring, show the component preview in the currently-selected language. Before, it was always showing the preview in default language.

## Test
- For units with translations
   - Component previews in node authoring view are displayed in the currently-selected language
   - Try switching languages and ensure that component previews update to display that language
- For units without translations: AT loads and works as before


breity (approved):
Looks good. 👍

PR #2254, opened 2026-01-16
feat(Notes): Display in dialog instead of mat-drawer

hirokiterashima:
## Changes
- Display notes in a dialog. Before, we were showing it in a mat-drawer on the right-side.
- Clean up code
   - use dialog.open() directly instead of using observables to control opening behavior
   - pass in insertMode as Dialog param, to indicate if we are in editing mode, or inserting mode

## Test
- Notes appear in a dialog
- Existing features of the notes work as before
   - Add/Edit/Delete notes
   - Insert note into report


breity (approved):
Looks good. 👍 

I increased the width of the notes dialog and increased the number of notes per row on wider screens.

PR #2256, opened 2026-01-16
feat(VLE): Show Step Tools when viewing Unit Plan

breity:
## Changes
The Step Tools toolbar now remains visible when viewing the Unit Plan in the VLE. This ensures that the buttons for viewing Notes and toggling the AI Assistant remain visible at all times.

Previous behavior:
Screenshot 2026-01-16 at 11 19 07 AM

New behavior:
Screenshot 2026-01-16 at 11 18 42 AM

## Test
Make sure nav view (Unit Plan) and node view (step view) work as before and that you can view Notes and toggle AI Assistant in both modes.


hirokiterashima (approved):
LGTM

PR #2257, opened 2026-01-20
fix(Unit Plan): Zoom to and highlight previous step

hirokiterashima:
Please style as you see fit.

## Changes
- When going to the unit plan, zoom to the previous step and highlight it

## Test
- above works as described


breity (approved):
Looks good. 👍

PR #2260, opened 2026-01-27
fix(style): Remove bullets from match and idea summary lists

breity:
## Changes
Fixes the list styles for match and idea summaries by removing the bullets and spacing.

Current:
Screenshot 2026-01-27 at 9 56 45 AM

Fixed:
Screenshot 2026-01-26 at 10 41 32 AM

Also reverts font-weight for `b` and `strong` tags and `strong` class to `bolder` to fix some content (especially in HTML components) not showing up as bold when it should.

## Test
Make sure match and idea summaries in the grading tool display correctly and work as before.


hirokiterashima:
Idea summary looks good 👍 

Match summary still shows bullets. Seems like we should remove them?
Screenshot 2026-01-27 at 2 51 47 PM

breity:
@hirokiterashima Ah, yes. The match inline Tailwind styles were be overwritten at times. Should be fixed now. Thanks.

hirokiterashima (approved):
LGTM

PR #2263, opened 2026-01-28
fix(Match Summary): Don't wrap choices bucket

### LINKED ISSUE #2261: style(Match Summary): bucket wrapping
Bucket wraps to the next column. In this case, I think it doesn't seem unnecessary? It'd be much easier to read if it's all in one block.
Image

breity:
## Changes
- Don't allow Choices bucket in Match summary display to break across columns. Wrapping to multiple columns was making the display difficult to read.
- Limit Choices bucket to 3 visible items by default and add a "Show more" link, like we do with the other buckets.

Before:
Screenshot 2026-01-28 at 3 32 27 PM

After:
Screenshot 2026-01-28 at 3 34 39 PM

## Test
- Make sure Match summary displays still in the grading tool.
- Make sure Choices bucket limits to 3 visible items and that "Show more" and "Show less" links toggle the rest of the items.

fixes #2261

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24329,7 +24329,6 @@
       "dev": true,
       "inBundle": true,
       "license": "ISC",
-      "peer": true,
       "dependencies": {
         "string-width": "^5.1.2",
         "string-width-cjs": "npm:string-width@^4.2.0",
@@ -24347,7 +24346,6 @@
       "dev": true,
       "inBundle": true,
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=12"
       },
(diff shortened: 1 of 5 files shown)


hirokiterashima (approved):
LGTM. AwsBedRockService isn't related to this PR, but it's fine to include. We should parameterize the model string.

breity:
@hirokiterashima Ah, sorry. Didn't mean to include the AwsBedRockService for this PR. I'll revert since it's not related.

I'm thinking it might make sense to switch to Gemma, though, since it seemed to have a bit more current info.

PR #2264, opened 2026-01-29
feat(GradeByStep): Show teams on node

hirokiterashima:
## Notes
Please style as you see fit. The person icon is not quite centered vertically. :p

I'm wondering if we need to show the "(Period 1)" and "(All Periods)" in the tooltip. It clutters up the tooltip text a bit and makes it hard to read. Would the teachers usually know which period they're currently looking at and can get this info by looking at the period select drop-down?

## Changes
- Show teams on nodes (lessons and steps) in the Grade by Step view, if any
   - An icon, with a count appears if there are any teams on that node.
   - A tooltip should show students on that step
- Changing periods should update the display to match the selected period (including all periods) 
- This should update in real-time as students move around in the unit.

## Test
- Above works as described


breity (approved):
Looks good. 👍 

I updated styles and removed the period name from NavItemProgress and TeamsOnNode tooltips.

PR #2265, opened 2026-02-03
feat(Grade by Student): Visual location display

hirokiterashima:
## Notes
Please style as you see fit! I had trouble with these, as you can see in the image below:
- The highlighted segment is not the same size as the non-highlighted segments. This seems to happen for only units with multiple lessons. It looks like the width of the segment is matching the width of the step position ("1.2")
- Text overlaps with the display above. 

## Changes 
Screenshot 2026-02-02 at 11 01 18 PM

In Grade by Student view, add a visual display (see above) to show where students are in the unit. Before, this was text-only, like "1.2: Let's Discuss". The visual should be more useful because it will give a quick overview of where the students are.
- If the unit has just 1 lesson, the segments are the steps in that lesson.
- If the unit has more than 1 lesson, the segments are the lessons in the unit.
- The segment that the student is on has a fill-color, a location marker, and the step number (e.g., "1.2")
- Hovering over the component shows step position text (e.g., "1.2: Let's Discuss")
- If the student hasn't done any work, it still shows the segments, but none of the segments are highlighted

## Test
- Above works as described
- Sorting by columns (location, completion, etc.) works as before

### DIFF
--- a/src/assets/wise5/classroomMonitor/classroomMonitorComponents/shared/testing/ClassroomMonitorTestHelper.ts
+++ b/src/assets/wise5/classroomMonitor/classroomMonitorComponents/shared/testing/ClassroomMonitorTestHelper.ts
@@ -1,4 +1,4 @@
-import { StudentProgress } from '../../../student-progress/student-progress.component';
+import { StudentProgress } from '../../../student-progress/StudentProgress';
 
 export class ClassroomMonitorTestHelper {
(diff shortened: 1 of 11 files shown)


breity (approved):
Looks great. 👍 I updated styles and location icon now hides when lesson segments are small (<2.5rem/40px width).

PR #2268, opened 2026-02-07
feat(IdeasSummary): Show student responses

hirokiterashima:
## Notes
Please style as you see fit. It would be great to distinguish between the responses a bit more, with maybe a border or background. Maybe make the responses italic?

Also, as you can see in the image below, the expanded idea has 3 students who had this idea, but we only show 2 users. Would this be a bit confusing because users might expect to see 3 students? Maybe we should explain this in the UI, like displaying a message "showing 2 of 3 responses" or "sample responses" in the expanded panel, or "click to see sample responses" when hovering over the expand icon?

## Changes
Screenshot 2026-02-06 at 4 06 06 PM

- For Dialog Guidance and Open Response idea summaries in the teacher tools, clicking on the ideas will show up to 2 examples of the students' actual responses that contained that idea.
- These are toggled divs, so you can have multiple details opened at the same time
- This should work in both the condensed and full views of the summary display.

## Test
- Test both Dialog Guidance and Open Response items and make sure that the feature works as described above.

### DIFF
--- a/src/app/modules/library/library-project-details/library-project-details.component.scss
+++ b/src/app/modules/library/library-project-details/library-project-details.component.scss
@@ -1,9 +1,6 @@
 @use '@angular/material' as mat;
-
 @use 'style/abstracts/functions';
-
-.library-project-details {
-}
+@reference "tailwindcss";
 
 .info-block {
   padding: 12px;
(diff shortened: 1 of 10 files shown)


breity (approved):
Functionality is great. I updated to utilize MatExpansionPanel for expanding and collapsing and updated styles. I also added "Sample responses:" text above the responses display.

PR #2269, opened 2026-02-09
feat(GradeByStep): Sync class responses with summary

hirokiterashima:
## Changes
This change makes the grade-by-step view less confusing, by only displaying one component when the page is first loaded. Before, we were showing all the components (and their combined score) in the step, but were showing one summary, so this mismatch was confusing.

## Test
- In the grade-by-step view, Class Responses section now shows the same component as in the summary tab.
- When the step is first loaded, we show the first component in the step.
- Choosing a different component in the summary updates the class responses section to show responses for that same component.


breity (approved):
LGTM 👍

PR #2276, opened 2026-02-25
feat(Application): Change "component" to "activity"

hirokiterashima:
To make our UI more user-friendly, we're moving away from using the term "component", and replacing it with "activity". These changes are mostly in the authoring tool.

Examples of changes:
- "add component" -> "add activity"
- "component width" -> "activity width"
- "connected components" -> "connected activities"

## Changes
- Change instances of "component" to "activity" and "components" to "activities", while keeping the letter-casing (e.g., Component -> Activity)
- Note that the code still uses the term "component" everywhere (e.g. ```node.addComponent()```)

## Test
- Make sure that you don't see the term "component" anywhere as a teacher and student. Test the authoring tool in particular, because this was where it was heavily used.


breity (approved):
Looks good. 👍

Issue #2287, opened 2026-03-11
bug(Authoring Tool): Moving branch step breaks transition


hirokiterashima:
Moving a branching step anywhere causes the transition to break

PR #2288, opened 2026-03-11
feat(Authoring Tool): Drag and drop lessons and steps

hirokiterashima:
We add a new default mode to the authoring tool unit editing view, which allows the author to move lessons and steps using drag and drop. The existing behavior (choosing lesson(s) or step(s) using the checkbox and then choosing a new location) has been moved to what we now call "batch edit mode".

## Notes
Please style as you see fit. Things we discussed:
- hiding controls (e.g., add step button) while dragging
- indenting drop preview to make it clear that the step will be dragged in a branch path

## Changes (in the unit editing view in the AT)
- Drag and drop lessons and steps in the default view
- Toggle "batch edit mode". This is what the edit unit view used to be 
- Disable moving a branch step. We found a bug https://github.com/WISE-Community/WISE-Client/issues/2287 while working on this PR. We'll disable moving a branching step in default mode and batch edit mode to prevent authors from breaking their units. 

## Test
- The new drag and drop feature works as expected. Test all scenarios (branch, inactive lessons, inactive nodes, etc)
- Existing batch edit mode works as before

### DIFF
--- a/src/assets/wise5/authoringTool/choose-node-location/node-icon-and-title/node-icon-and-title.component.html
+++ b/src/assets/wise5/authoringTool/choose-node-location/node-icon-and-title/node-icon-and-title.component.html
@@ -1,7 +1,7 @@
 <div class="flex flex-row justify-start items-center gap-1">
   <node-icon [nodeId]="nodeId" size="18" />&nbsp;
   @if (showPosition) {
-    <span class="step-number">{{ getNodePosition(nodeId) }}: </span>
+    <span class="step-number">{{ nodePosition }}: </span>
   }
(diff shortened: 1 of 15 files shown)


breity (approved):
Looks good. 👍 

I changed some styles, made some UI updates, and fixed most of the dragging visual bugs. Dragging to and from the unused sections still results in visual errors. But we decided to merge as is and re-design the unused section in a subsequent PR.

There's also still an issue dragging to/from the last lesson in the unit that prevents placing a step in certain positions in other lessons. Good to merge once this is fixed.

PR #2289, opened 2026-03-12
feat(Discussion): Anonymize responses

hirokiterashima:
## Changes
- Student responses in the Discussion activity can now be displayed with anonymous names instead of real names. The anonymous names are chosen from a list of animal names (Anonymous Lion, Anonymous Yeti, etc...)
- Add option to anonymize student responses in the authoring tool. By default, this option is disabled.
- Add option to toggle anonymizing student responses in the teacher tool > discussion summary view.

## Test
- Above works as described

### DIFF
--- a/src/assets/wise5/common/ComponentContent.ts
+++ b/src/assets/wise5/common/ComponentContent.ts
@@ -4,6 +4,7 @@ import { DynamicPrompt } from '../directives/dynamic-prompt/DynamicPrompt';
 
 export interface ComponentContent {
   id: string;
+  anonymizeResponses?: boolean;
   connectedComponents?: any[];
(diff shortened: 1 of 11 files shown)


breity (approved):
Looks great. I updated some styles and text.

I had a couple thoughts:
1. I think students should always be able to see their own names in the discussion, even if it is set to be anonymous. As of now, there's no way for students to quickly find and remember the posts and comments they made.
2. It would be nice if the anonymous setting in the teacher summary view could sync between the regular and expanded views. This could come in a separate PR.

hirokiterashima:
Your changes look great. Also, good ideas for improvement. I was able to implement #1. Let's save #2 for another time.

PR #2290, opened 2026-03-16
feat(Step Icon): Improve default option

### LINKED ISSUE #2286: feat(Step Icon): Improve default option
Currently, we assign a gray, hat icon for new steps and lessons. This makes the unit rather dull. Instead, do the following:
- New lesson: choose random color, abstract grid icon as default
- New step: same color as lesson, hat icon

hirokiterashima:
Currently, we assign a gray hat icon for new steps and lessons. This makes the unit feel gray and dull. This PR tries to improve this by making the unit more colorful and organized. We do this by changing the following:

- New lesson icon will have a random color, grid icon ("dashboard" material icon)
- New step icon will have the same color as the lesson, hat icon ("school" material icon)
   - This includes branch path steps and the merge step 
- Imported step(s) icon(s) will have the same color as lesson they get imported into, same icon as before

## Test
- Above works as described

Closes #2286

### DIFF
--- a/src/app/services/editBranchService.spec.ts
+++ b/src/app/services/editBranchService.spec.ts
@@ -58,6 +58,13 @@ function editBranch() {
         id: 'node2',
         title: 'Branch point',
         type: 'node',
+        icon: {
+          color: '#00B0FF',
+          type: 'font',
+          fontSet: 'material-icons',
+          fontName: 'school',
+          imgSrc: ''
+        },
         transitionLogic: {
           transitions: [
             {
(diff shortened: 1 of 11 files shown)


breity (changes requested):
Looks good. 👍 

One suggestion: If a step being imported doesn't have an icon, we should set an icon (hat and color of the lesson being imported to) for the new step. Right now, the imported step will use the hat icon and default gray color.

hirokiterashima:
Fixed! PTAL.

breity (approved):

PR #2292, opened 2026-03-26
feat(Teacher Tools): Important activities

hirokiterashima:
This new feature allows teachers to quickly see important activities in the Teacher Tools. To implement this, we add an "!important" tag to an activity. These activities appear with a star icon in the teacher tools, in the step listing page and in the activities tabs in the step grading page.

## Notes
Please style as you see fit.

## Changes
- In the AT, author can mark and unmark an activity as important using a star/unstar button next to the activity type in the step editing page.
- In the Teacher Tools
   - In the step listing page, if a step contains a starred activity, a star will appear next to the step title. Clicking on this will take you to the step grading page, and will open to the first starred activity in that step.
   - In the step grading page, if an activity in the step is starred, it will display the star icon. You will also see the icon next to the starred activity in the "Class Responses" filter listing.

## Test
- Feature works as described above.


breity (approved):
Looks great. 👍 

I switched to the bookmark icon instead of the star and updated terminology.

PR #2296, opened 2026-03-31
feat(Discussion): Teacher can show/hide responses

hirokiterashima:
Teacher can now show/hide students' responses in the discussion summary view in the Teacher Tools.

## Notes
Please style and edit the phrasing as you see fit. 

## Changes
- For both top-level posts and replies, show a "Hide Post" button.
   - If top-level post is hidden, it and its replies will be hidden.
   - if a reply is hidden, it will be hidden.
   - both changes take immediate effect on the students' view (the posts and replies become hidden/visible right away) 
- You can show (unhide) a post or reply. Hidden posts will have a "Show Post" button.

## Test
- Hiding and showing posts and replies work as expected.
- Student discussion works as before. I cleaned up some code that was no longer needed due to this change.


breity (approved):
Looks good. 👍 

I updated to use icon buttons to show/hide responses and moved the show/hide buttons to the top of the post/comment. I also simplified the tooltip text.

hirokiterashima:
Changes look great.

PR #2297, opened 2026-04-03
style(Authoring Tool): Use tabs to organize activity advanced authoring

hirokiterashima:
The advanced activity authoring popup was a mess, with everything appearing on one page. This PR brings organization to the page using tabs. The default tabs are "general", "visibility", "connections", "tags", "rubric", and "JSON". Open Response has an additional "AI" tab.

## Notes
Please style as you see fit. But don't worry about the styling of the contents inside the tabs for this PR. We can do that in a future task.

## Changes
- Use tabs to organize activity advanced authoring for all activities
- Fix Summary authoring. This wasn't working at all.

## Test
- Above works as described. 
- Changes in the advanced activity authoring are saved as expected.


breity (approved):
Looks good. 👍 

I updated some styles and made the tab header sticky so that the tabs are visible when scrolling long setting content.

One suggestion: I think we should move the AI tab further up in the order for activity types that have it. AI should probably at least come before the Rubric and Tags tabs, as those are used rather infrequently. Once we make the AI settings more intuitive, we might even want to put AI before the Visibility and Connections tabs. What do you think?

hirokiterashima:
Good suggestion. I made the change. Moving the AI tab up more in the future makes sense too.

Right now, we have code that assumes that the visibility tab is the second tab in the dialog, for purpose of opening the tab when the author clicks on the visibility status icon on the component. We should remember this when we order the tabs around in the future, because we'll need to refactor the code then to make sure this functionality still works.

PR #2299, opened 2026-04-06
feat(TeacherTools): Redesign Match summary to organize by choice

breity:
## Changes
- Instead of showing each bucket, Match summaries now show each choice with the number of times the choice was moved into different buckets.

Screenshot 2026-04-06 at 11 15 13 AM

## Test
- In the Teacher Tools, ensure that Match summaries now list all the choices instead of buckets, with the number of times the choice was moved to different buckets.
- Choices should be ordered by total number of times they were moved to any bucket, then alphabetically.
- Choices that were not moved by any students to any buckets should say so and have lighter gray text.


hirokiterashima (approved):
LGTM

PR #2300, opened 2026-04-06
feat(AI Summary): Allow author to edit system prompt

hirokiterashima:
This PR allows authors to edit the default summarization system prompts for OR and Discussion components.

## Changes
- Show a textarea in the Advanced component authoring > "AI" tab for OR and Discussion components.
   - If this is your first time, the default summary system prompt should appear.
   - You can edit the system prompt. It should save automatically
- Use variables ```$QUESTION$``` and ```$RESPONSE_FORMAT$``` in the system prompt. They will be replaced by the component's prompt and the expected response format for each component type before the prompt is sent to the LLM.

## Test Prep
- Enable AI for the project by adding project.ai.enabled = true
```
{
...
"ai": {"enabled": true },
...
```

## Test
- Units with AI disabled should not show the textarea
- Units with AI enabled
   - Should show the text area and let you overwrite the system prompt, as described above
   - If author edited the system prompt, it the Teacher Tools should use the edited system prompt when summarizing student. Otherwise, it should use the default system prompt.


breity (approved):
Looks good. 👍

PR #2302, opened 2026-04-13
style(Component Selection): Group by function

hirokiterashima:
Screenshot 2026-04-12 at 7 25 39 PM

## Notes
- Please style as you see fit!
- There is a lot of code overlap in the two places where we show this component groupings, but I also saw that they had some differences in styles, so I didn't extract a new component. Feel free to extract to a new component if you can and reduce duplication.

## Changes
- In AddYourOwnNode (creating a new blank step) and ChooseNewComponent (adding a component to an existing step) views, organize the components in groups based on function, as seen in the screenshot above.

## Test
- Test the two views and make sure they work as before
   - AddYourOwnNode (creating a new blank step)
   - ChooseNewComponent (adding a component to an existing step)
- The component select dropdown in the component info dialog works as before. Code that this feature uses was slightly updated in this PR.

### DIFF
--- a/src/app/authoring-tool/add-component/choose-new-component/choose-new-component.component.html
+++ b/src/app/authoring-tool/add-component/choose-new-component/choose-new-component.component.html
@@ -1,22 +1,18 @@
 <h2 mat-dialog-title i18n>Add New Activity</h2>
-<mat-dialog-content>
+<mat-dialog-content class="@container">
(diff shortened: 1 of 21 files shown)


breity (approved):
Looks good. 👍 

I extracted the component type groups to its own component and updated styles and layout.

hirokiterashima:
New changes look great. 👍  Thanks!

PR #2307, opened 2026-04-24
feat(TeacherTools): Restore Match summary organize by bucket

breity:
## Changes
- Put back the organize by bucket view for Match (Sort) activity summaries in the teacher tools.
- Add a toggle to switch between the item and bucket views.

## Test
- Make sure item view works as before for Match teacher summaries.
- Make sure bucket view works properly and the buckets are sorted by total number of times any items were moved to each bucket.


hirokiterashima (approved):
LGTM. In the future, it'd be nice to see the student names when we hover over the person icon, like we do in the idea summary view.

PR #2309, opened 2026-05-06
style(MC Authoring): UI Improvements

hirokiterashima:
Screenshot 2026-05-05 at 8 30 51 PM

## Notes
Please style as you see fit. 
- I couldn't get the add choice button and the move choice up/down buttons to align vertically. 
- The delete choice button could also be closer to the top-right corner of the choice box. 
- Maybe distinguish the "Choices" section a little bit somehow?

## Changes
- Added button to allow adding new choices at any position. Before, it only let you add a new choice at the end
- Moved move up/down buttons to the right of the choices, similar to moving components up/down in node editing view
- Moved delete choice button to the top right corner of the choice box

## Test
- Above works as described


breity (approved):
Looks good. 👍 

I cleaned up styles, layout, and responsiveness.

PR #2310, opened 2026-05-06
feat(Authoring): Use slider to set activity width

hirokiterashima:
Screenshot 2026-05-06 at 4 43 24 PM

## Notes
Please style as you see fit. It might be better to hide the explanation text in an info tooltip instead of always displaying it.

## Changes
- In the activity advanced authoring > general tab, use a ```mat-slider``` to set the activity's width, in percentages. Possible values are 25, 33, 50, 66, 75, and 100. The default is 100.
- Add a blurb about what setting the activity width means in terms of display layout.

Before, we just displayed a input without any instructions, which was confusing for most people.

## Test
- Existing activities show 100 (which is the default value).
- You can change the activity width to any of the possible values.
- Changes are saved.

### DIFF
--- a/src/app/authoring-tool/edit-component-width/edit-component-width.component.ts
+++ b/src/app/authoring-tool/edit-component-width/edit-component-width.component.ts
@@ -6,7 +6,7 @@ import { MatSliderModule } from '@angular/material/slider';
 @Component({
   imports: [FormsModule, MatSliderModule],
   selector: 'edit-component-width',
-  template: `<span i18n>Activity width:</span>
+  template: `<span i18n>Activity width:</span>&nbsp;
     <mat-slider
       min="0"
       [max]="possibleValues.length - 1"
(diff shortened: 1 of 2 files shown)


breity (approved):
Functionality works well. I updated the explanation text and reduced the width of the slider slightly.

Because the slider goes from 25%-100%, the position of the slider thumb doesn't match visually with the selected value for anything under 100%. This is a bit confusing to me, as you would assume that a 50% width should be right in the middle of the slider, for example. Ideally, the slider could include 0%, but it would not be selectable. I experimented a bit and this is not easy to do with a mat-slider. Perhaps it's not too big of an issue for now. But another option could be to use a mat-button-toggle instead of a slider to reduce the visual disconnect.

I was also thinking that the width authoring setting could be moved out of the "General" advanced authoring tab, as it gets buried at the bottom for component types with lots of advanced settings. I think we could group it with the visibility constraint authoring, which is currently the only setting in the "Visibility" tab. We could rename the tab to "Appearance" or something like that and show width as the first setting and visibility constraints below that. This could be done is a separate PR.

hirokiterashima:
As we discussed, I changed this component to use a mat-toggle-button instead of a mat-slider and also hid the explanation text in a tooltip. Good improvements! PTAL and style as needed.

breity:
Looks great. I updated styles and tooltip text slightly.

PR #2319, opened 2026-06-05
feat(TeacherTools): Show workgroups in tooltip for Sort summary

breity:
## Changes
- Show a list of workgroup names in a tooltip when hovering over each item or bucket in Sort (Match) summary view in the teacher tools.

## Test
- Go to Grade by Step in the Teacher Tools for a run with a step that includes a Sort activity. Select the step and then the question summary for the Sort activity.
- In Bucket view, make sure that when you mouse over (or Tab to using keyboard) the person icon + count for each item that a tooltip appears showing the names for the workgroups that moved that item into the bucket.
- In Item view, make sure that when you mouse over (or Tab to using keyboard) the person icon + count for each bucket that a tooltip appears showing the names for the workgroups that moved the item into that bucket.

### DIFF
--- a/src/assets/wise5/directives/teacher-summary-display/match-summary-display/match-summary-display.component.html
+++ b/src/assets/wise5/directives/teacher-summary-display/match-summary-display/match-summary-display.component.html
@@ -15,7 +15,7 @@
             @if (!isChoiceReuseMatch) {
               <span
                 class="shrink-0 cursor-pointer"
-                [matTooltip]="getWorkgroupNames(choice)"
+                [matTooltip]="workgroupNamesTooltips.get(choice)"
                 matTooltipClass="multiline-tooltip"
                 matTooltipPosition="above"
                 tabindex="0"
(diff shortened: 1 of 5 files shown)


hirokiterashima (changes requested):
The new tooltip text works great for both owner and shared owner, but I'm a bit concerned about the hit to the performance this will cause.

The new ```getWorkgroupNames()``` function gets called many times. You can see this by adding console.log() in the function and just moving your mouse over the screen. It does the same calculations and returns the same result. 

This could cause an issue with a class with many students and many buckets, especially on slower machines. Can we cache the result somewhere- maybe in MatchSummaryDisplayComponent or MatchSummaryDataPoint?

breity:
@hirokiterashima Ah yes, I agree that could hinder performance. I refactored to cache the workgroup names in the MatchSummaryDisplayComponent.

hirokiterashima (approved):
New changes look good. 👍

PR #2324, opened 2026-06-25
feat(Home Page): Admin news on home

Aaron-Detre:
## Changes
- When Discourse is not set in application properties, the home page should show news items from the admin tools.
- Moved content of DiscourseLatestNewsComponent to LatestNewsComponent which is shared between Discourse and Admin news.
- Admin news items link to pre-existing /news page. 
- NewsComponent was using [innerHTML]="sanitizer.bypassSecurityTrustHtml(newsItem.news)" instead of just displaying the text. I didn't see any reason it shouldn't just display the text, so I changed it, but I'm not sure if that was something important.

## Test
- When Discourse url and news category are set, the home page should still show Discourse news, otherwise it should show admin news.
- Admin news should show the 3 latest public news items.
- The More News button should link to the separate admin news page.
- Each of the three news items should link to that same page and should automatically scroll to that news item.
- If signed out, the news page should only show public news items. If signed in, it should show all news items.


src/app/news/news.component.ts
```diff
@@ -1,7 +1,6 @@
-import { Component, OnInit } from '@angular/core';
+import { Component, ElementRef, OnInit, ViewChild } from '@angular/core';
```
hirokiterashima: Remove unused imports

src/app/home/admin-latest-news/admin-latest-news.component.ts
```diff
@@ -0,0 +1,26 @@
...
+
+@Component({
+  selector: 'admin-latest-news',
+  imports: [LatestNewsComponent],
+  templateUrl: './admin-latest-news.component.html',
+  styleUrl: './admin-latest-news.component.scss'
+})
+export class AdminLatestNewsComponent {
+  protected topics: News[] = [];
+  protected isLoaded: boolean = false;
```
hirokiterashima: Alphabetize names. Change name isLoaded => loaded
```suggestion
  protected loaded: boolean = false;
  protected topics: News[] = [];
```

src/app/home/admin-latest-news/admin-latest-news.component.ts
```diff
@@ -0,0 +1,26 @@
+import { Component } from '@angular/core';
+import { LatestNewsComponent } from '../latest-news/latest-news.component';
+import { News } from '../../domain/news';
+import { NewsService } from '../../services/news.service';
+
+@Component({
+  selector: 'admin-latest-news',
+  imports: [LatestNewsComponent],
+  templateUrl: './admin-latest-news.component.html',
+  styleUrl: './admin-latest-news.component.scss'
```
hirokiterashima: Delete line and also remove empty scss file.
```suggestion
```

src/app/home/discourse-latest-news/discourse-latest-news.component.ts
```diff
@@ -1,32 +1,11 @@
...
 import { DiscourseFeedComponent } from '../../discourse-feed/discourse-feed.component';
-import { CommonModule } from '@angular/common';
-import { MatIconModule } from '@angular/material/icon';
+import { LatestNewsComponent } from '../latest-news/latest-news.component';
 
 @Component({
-  imports: [CommonModule, MatIconModule],
+  imports: [LatestNewsComponent],
   selector: 'discourse-latest-news',
   styleUrl: 'discourse-latest-news.component.scss',
```
hirokiterashima: Delete line and remove empty scss file.
```suggestion
```

hirokiterashima (changes requested):
Looking great! 👍  Features work as described. 

Let's add back the bypassSecurityTrustHtml() bit. This allows the admins to use HTML syntax in their news posts, like adding a link or making text bold. Without it, it will show those html blocks, which is not ideal.

Also, see code improvement suggestions inline. I may not have caught all the places to clean up unused import/code.

hirokiterashima (approved):
LGTM

PR #2326, opened 2026-07-07
feat(Home Page): News caching

Aaron-Detre:
## Changes
- Switched to using the new news endpoints for the home page and dedicated news page which means you no longer have to filter or slice on the client side.

## Test
- test with https://github.com/WISE-Community/WISE-API/pull/329
- Everything should look and function the same as it did here: https://github.com/WISE-Community/WISE-Client/pull/2324


hirokiterashima (approved):
LGTM

